PR #43, opened 2019-11-20
Use tag of kubernetes-conjur-deploy for deploying OSS by default

orenbm:
Until we fix kubernetes-conjur-deploy to have both OSS & DAP properly we will use a tag of that repo that has this functionality.

### DIFF
--- a/test/test_in_docker.sh
+++ b/test/test_in_docker.sh
@@ -20,7 +20,7 @@ function buildTestRunnerImage() {
 function deployConjur() {
   pushd ..
     git clone --single-branch \
-      --branch master \
+      --branch deploy-oss-by-default \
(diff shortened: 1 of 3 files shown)


test/test_in_docker.sh
```diff
@@ -20,7 +20,7 @@ function buildTestRunnerImage() {
 function deployConjur() {
   pushd ..
     git clone --single-branch \
-      --branch master \
+      --branch deploy-oss-by-default \
```
sigalsax: is this a branch that is already pushed to the k8s-conjur-deploy master?

orenbm: it is a tagged commit in `master`, not a branch.

PR #51, opened 2019-12-08
Add additional tests

orenbm:
This PR adds tests for the following cases:
- Conjur authentication failure fails the pod and logs accordingly
- Conjur secrets retrieval failure fails the pod and logs accordingly

### DIFF
--- a/test/config/k8s/test-env.sh.yml
+++ b/test/config/k8s/test-env.sh.yml
@@ -9,6 +9,7 @@ set -euo pipefail
 SECRETS_DESTINATION_KEY_VALUE=${SECRETS_DESTINATION_KEY_VALUE:-"SECRETS_DESTINATION k8s_secrets"}
 CONTAINER_MODE_KEY_VALUE=${CONTAINER_MODE_KEY_VALUE:-"CONTAINER_MODE init"}
 K8S_SECRETS_KEY_VALUE=${K8S_SECRETS_KEY_VALUE:-"K8S_SECRETS test-k8s-secret"}
+CONJUR_AUTHN_LOGIN_KEY_VALUE=${CONJUR_AUTHN_LOGIN_KEY_VALUE:-"CONJUR_AUTHN_LOGIN host/conjur/authn-k8s/${AUTHENTICATOR_ID}/apps/${TEST_APP_NAMESPACE_NAME}/*/*"}
 
 cat << EOL
 ---
@@ -74,10 +75,6 @@ spec:
           - name: CONJUR_ACCOUNT
             value: ${CONJUR_ACCOUNT}
 
-          - name: CONJUR_AUTHN_LOGIN
-            value: >-
-              host/conjur/authn-k8s/${AUTHENTICATOR_ID}/apps/${TEST_APP_NAMESPACE_NAME}/*/*
-
           - name: CONJUR_SSL_CERTIFICATE
             valueFrom:
               configMapKeyRef:
@@ -90,6 +87,7 @@ spec:
 `yaml_print_key_name_value "          " ${K8S_SECRETS_KEY_VALUE}`
 `yaml_print_key_name_value "          " ${CONTAINER_MODE_KEY_VALUE}`
 `yaml_print_key_name_value "          " ${SECRETS_DESTINATION_KEY_VALUE}`
+`yaml_print_key_name_value "          " ${CONJUR_AUTHN_LOGIN_KEY_VALUE}`
 
       imagePullSecrets:
         - name: dockerpullsecret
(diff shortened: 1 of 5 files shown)


test/test_cases/TEST_ID_11_conjur_authn_failure.sh
```diff
@@ -0,0 +1,14 @@
...
+
+create_secret_access_role_binding
+
+export CONJUR_AUTHN_LOGIN_KEY_VALUE="CONJUR_AUTHN_LOGIN host/some-policy/non-existing-namespace/*/*"
+
+deploy_test_env
+
+echo "Expecting secrets provider to fail with error CAKC016E Login failed"
+pod_name=$(cli_get_pods_test_env | awk '{print $1}')
+wait_for_it 600 "$cli logs $pod_name -c cyberark-secrets-provider | grep 'CAKC015E'"
```
sigalsax: when checking error codes, it looks like `CAKC015E` is Login failure, not `CAKC016E`
from `log_messages`
`const CAKC015E string = "CAKC015E Login failed`

test/config/k8s/test-env.sh.yml
```diff
@@ -74,10 +75,6 @@ spec:
           - name: CONJUR_ACCOUNT
             value: ${CONJUR_ACCOUNT}
 
-          - name: CONJUR_AUTHN_LOGIN
-            value: >-
-              host/conjur/authn-k8s/${AUTHENTICATOR_ID}/apps/${TEST_APP_NAMESPACE_NAME}/*/*
-
```
sigalsax: In our docs, we ask that the customer provide this `CONJUR_AUTHN_LOGIN` variable in this manifest. For the sake of uniformity, can we find a way to keep `CONJUR_AUTHN_LOGIN` here and dynamically provide it's value before the tests run. We run the deployment in `deploy_test_env` and since you are already exported, before running the actual test so supply it. Like so:
```
- name: CONJUR_AUTHN_LOGIN
   value: ${CONJUR_AUTHN_LOGIN_KEY_VALUE}
```

test/config/openshift/test-env.sh.yml
```diff
@@ -73,10 +74,6 @@ spec:
           - name: CONJUR_ACCOUNT
             value: ${CONJUR_ACCOUNT}
 
-          - name: CONJUR_AUTHN_LOGIN
-            value: >-
-              host/conjur/authn-k8s/${AUTHENTICATOR_ID}/apps/${TEST_APP_NAMESPACE_NAME}/*/*
-
```
sigalsax: see above comment regarding keeping things uniform

test/config/k8s/test-env.sh.yml
orenbm: that's not how `CONJUR_AUTHN_LOGIN_KEY_VALUE` works. please note that we do the same with `SECRETS_DESTINATION`, `CONTAINER_MODE` & `K8S_SECRETS`. All these variables are in the docs and will also be in the manifest that will be created once the bash script finishes to add the variables. 

please go over this method and tell me if you better understand it: https://github.com/cyberark/secrets-provider-for-k8s/blob/master/test/utils.sh#L202

if you still think we should change the way we add variables dynamically tell me.

test/policy/templates/project-authn-def.template.sh.yml
```diff
@@ -20,6 +20,13 @@ cat << EOL
         kubernetes/authentication-container-name: cyberark-secrets-provider
         openshift: "true"
 
+    # this host will not have permissions on Conjur secrets to test this use-case
+    - !host
+      id: ${TEST_APP_NAMESPACE_NAME}/service_account/${TEST_APP_NAMESPACE_NAME}-sa
```
sigalsax: why do we have this long path here? and why service_account instead of just a regular host?

test/test_cases/TEST_ID_11_conjur_authn_failure.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euxo pipefail
+
+create_secret_access_role
+
+create_secret_access_role_binding
+
+export CONJUR_AUTHN_LOGIN_KEY_VALUE="CONJUR_AUTHN_LOGIN host/some-policy/non-existing-namespace/*/*"
```
sigalsax: what does this extra space in `CONJUR_AUTHN_LOGIN host/some-policy/non-existing-namespace/*/*"` do? Does it set CONJUR_AUTHN_LOGIN: host/some-policy/non-existing-namespace/*/*, if so, how?

test/test_cases/TEST_ID_12_no_conjur_secrets_permissions.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euxo pipefail
+
+create_secret_access_role
+
+create_secret_access_role_binding
+
+export CONJUR_AUTHN_LOGIN="host/conjur/authn-k8s/${AUTHENTICATOR_ID}/apps/${TEST_APP_NAMESPACE_NAME}/service_account/${TEST_APP_NAMESPACE_NAME}-sa1"
```
sigalsax: oh no! 'sa1' :(

orenbm: fixed

test/test_cases/TEST_ID_11_conjur_authn_failure.sh
orenbm: explain face to face

PR #55, opened 2019-12-26
Use a new authn-client version

orenbm:
connected to cyberark/conjur#1239

The new version sends the full host-id in the CSR request

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,19 +6,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [0.3.0] - 2019-12-26
+
+### Changed
+  - Using a new authn-client version that sends the full host-id in the CSR 
+    request so we have this capability in this project. This enables users to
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,19 +6,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-## [0.2.0] - 2019-09-19
+## [0.3.0] - 2019-12-26
```
orenbm: Should we change the version from `0.2.0` to `0.3.0` or to `0.2.1`?
@InbalZilberman @Tovli  wdyt?

orenbm: @izgeri can you give any insight on this?

izgeri: We usually follow semver for this (https://semver.org/). I'd bump the minor version if you feel like this adds functionality and patch otherwise.

orenbm: thanks, that's very educating.In that case i am fine with the minor version bump.

PR #56, opened 2019-12-31
WIP - Remove redundant RoleBinding from test env

orenbm:
EDIT: this is actually not redundant. This PR will change to only add a comment on why we need this.

We were loading a RoleBinding that is not needed for running the tests. This PR removes it. 

Connected to cyberark/conjur-docs#651

Blocked by cyberark/kubernetes-conjur-deploy#92

DoD

- [ ] Old tags are removed for `deploy-oss`
- [ ] Docs link to RoleBinding that is needed to be loaded


Tovli:
Won't be fix as the effort isn't worth the value

PR #60, opened 2020-01-23
Bump version to 0.4.0

orenbm:
Using a new conjur-authn-k8s-client version that enables authentication of
hosts that have their application identity defined in annotations


sigalsax:
We have to bump a version along with the conjur-authn-k8s-client bump? Out of curiosity, why and when would we not do this?

orenbm:
> We have to bump a version along with the conjur-authn-k8s-client bump? Out of curiosity, why and when would we not do this?

Anytime we want the secrets-provider to include the changes that were introduced in the new authn-client then we need to release a new version

PR #74, opened 2020-03-15
Make secrets-provider-for-k8s stable

eladkug:
secrets-provider-for-k8s is not stable and i did the following things to improve this job:
1. main issue was that sometimes we get route issue from os (known issue) 
`[2020-03-02T19:39:07.653Z] Unable to connect to the server: dial tcp 54.197.37.237:8443: connect: no route to host`
Since we already had retry function in the code, i changed the mechanism to make sure all our cli commands will use this function. 
I ran more than 150 runs and find 6 different commands that failed on the same issue and fixed by this solution. 
2. i saw sometimes gcloud crashed : database is locked -SQLite is meant to be a lightweight database, and thus can't support a high level of concurrency (today all the environments run in parallel ) i separate our work to run concurrency dap/oss

*Please be aware that this PR will not solve retry issue that may occur in kubernetes-conjur-deploy.
During my tests this problem occur 8/150 runs*

this code will . not pushed unless i will see 10 success job

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,8 @@ pipeline {
 
   options {
     timestamps()
+    /* we want to avoid running in parallel otherwise we can get gcloud crashed : database is locked
+       working with same environment in parallel */
     disableConcurrentBuilds()
     buildDiscarder(logRotator(numToKeepStr: '30'))
   }
@@ -32,21 +34,21 @@ pipeline {
       }
     }
 
+    /* we want to avoid running in parallel otherwise we can get gcloud crashed : database is locked
+       working with same environment in parallel */
    stage ("Run Integration Tests on oss") {
       steps {
         script {
           def tasks = [:]
-          ["oss"].each { deployment ->
-            tasks["Kubernetes GKE, ${deployment}"] = {
-                sh "./bin/test_integration --docker --${deployment} --gke"
+            tasks["Kubernetes GKE, oss"] = {
+                sh "./bin/test_integration --docker --oss --gke"
             }
-            tasks["Openshift v3.11, ${deployment}"] = {
-                sh "./bin/test_integration --docker --${deployment} --oc311"
+            tasks["Openshift v3.11, oss"] = {
+                sh "./bin/test_integration --docker --oss --oc311"
             }
-            tasks["Openshift v3.10, ${deployment}"] = {
-                sh "./bin/test_integration --docker --${deployment} --oc310"
+            tasks["Openshift v3.10, oss"] = {
+                sh "./bin/test_integration --docker --oss --oc310"
             }
-          }
           parallel tasks
         }
       }
@@ -56,17 +58,15 @@ pipeline {
           steps {
             script {
               def tasks = [:]
-              ["dap"].each { deployment ->
-                tasks["Kubernetes GKE, ${deployment}"] = {
-                    sh "./bin/test_integration --docker --${deployment} --gke"
+                tasks["Kubernetes GKE, dap"] = {
+                    sh "./bin/test_integration --docker --dap --gke"
                 }
-                tasks["Openshift v3.11, ${deployment}"] = {
-                    sh "./bin/test_integration --docker --${deployment} --oc311"
+                tasks["Openshift v3.11, dap"] = {
+                    sh "./bin/test_integration --docker --dap --oc311"
                 }
-                tasks["Openshift v3.10, ${deployment}"] = {
-                    sh "./bin/test_integration --docker --${deployment} --oc310"
+                tasks["Openshift v3.10, dap"] = {
+                    sh "./bin/test_integration --docker --dap --oc310"
                 }
-              }
               parallel tasks
             }
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ pipeline {
 
   options {
     timestamps()
+    disableConcurrentBuilds()
     buildDiscarder(logRotator(numToKeepStr: '30'))
   }
 
@@ -31,11 +32,11 @@ pipeline {
       }
     }
 
-    stage ("Run Integration Tests") {
+   stage ("Run Integration Tests on oss") {
       steps {
         script {
           def tasks = [:]
-          ["oss", "dap"].each { deployment ->
+          ["oss"].each { deployment ->
             tasks["Kubernetes GKE, ${deployment}"] = {
                 sh "./bin/test_integration --docker --${deployment} --gke"
             }
@@ -50,6 +51,26 @@ pipeline {
         }
       }
     }
+
+    stage ("Run Integration Tests on dap") {
+          steps {
+            script {
+              def tasks = [:]
+              ["dap"].each { deployment ->
+                tasks["Kubernetes GKE, ${deployment}"] = {
+                    sh "./bin/test_integration --docker --${deployment} --gke"
+                }
+                tasks["Openshift v3.11, ${deployment}"] = {
+                    sh "./bin/test_integration --docker --${deployment} --oc311"
+                }
+                tasks["Openshift v3.10, ${deployment}"] = {
+                    sh "./bin/test_integration --docker --${deployment} --oc310"
+                }
+              }
+              parallel tasks
+            }
+          }
+        }
 
     stage('Publish client Docker image') {
         steps {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,8 +5,9 @@ pipeline {
 
   options {
     timestamps()
-    /* we want to avoid running in parallel otherwise we can get gcloud crashed : database is locked
-       working with same environment in parallel */
+    // We want to avoid running in parallel.
+    // When we have 2 build running on the same environment (gke env only) in parallel,
+    // we get the error "gcloud crashed : database is locked"
     disableConcurrentBuilds()
     buildDiscarder(logRotator(numToKeepStr: '30'))
   }
@@ -34,8 +35,9 @@ pipeline {
       }
     }
 
-    /* we want to avoid running in parallel otherwise we can get gcloud crashed : database is locked
-       working with same environment in parallel */
+    // We want to avoid running in parallel.
+    // When we have 2 build running on the same environment (gke env only) in parallel,
+    // we get the error "gcloud crashed : database is locked"
    stage ("Run Integration Tests on oss") {
       steps {
         script {
@@ -54,17 +56,17 @@ pipeline {
       }
     }
 
-    stage ("Run Integration Tests on dap") {
+    stage ("Run Integration Tests on DAP") {
           steps {
             script {
               def tasks = [:]
-                tasks["Kubernetes GKE, dap"] = {
+                tasks["Kubernetes GKE, DAP"] = {
                     sh "./bin/test_integration --docker --dap --gke"
                 }
-                tasks["Openshift v3.11, dap"] = {
+                tasks["Openshift v3.11, DAP"] = {
                     sh "./bin/test_integration --docker --dap --oc311"
                 }
-                tasks["Openshift v3.10, dap"] = {
+                tasks["Openshift v3.10, DAP"] = {
                     sh "./bin/test_integration --docker --dap --oc310"
                 }
               parallel tasks
--- a/test/2_create_test_app_namespace.sh
+++ b/test/2_create_test_app_namespace.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 announce "Creating Test App namespace."
 
 if [[ $PLATFORM == openshift ]]; then
-  oc login -u $OPENSHIFT_USERNAME
+  $cli_with_timeout "login -u $OPENSHIFT_USERNAME"
 fi
 
 set_namespace default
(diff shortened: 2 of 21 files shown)


orenbm:
@eladkug is it possible to split this PR into smaller commits? It can enhance it readability and make the review easier.

Jenkinsfile
```diff
@@ -5,6 +5,7 @@ pipeline {
 
   options {
     timestamps()
+    disableConcurrentBuilds()
```
orenbm: does this mean that all the steps will run asynchronously?  How much time does it add to each build?

Jenkinsfile
```diff
@@ -31,11 +32,11 @@ pipeline {
       }
     }
 
-    stage ("Run Integration Tests") {
+   stage ("Run Integration Tests on oss") {
       steps {
         script {
           def tasks = [:]
-          ["oss", "dap"].each { deployment ->
+          ["oss"].each { deployment ->
```
orenbm: if you have only one object in the list you can get rid of it and use it hard-coded.

test/2_create_test_app_namespace.sh
```diff
@@ -18,9 +18,9 @@ else
   echo "Creating '$TEST_APP_NAMESPACE_NAME' namespace."
 
   if [ $PLATFORM = "kubernetes" ]; then
-    $cli create namespace $TEST_APP_NAMESPACE_NAME
+    $cli "create namespace $TEST_APP_NAMESPACE_NAME"
```
orenbm: what's the difference here?

orenbm: If it's important to add quotes please say so in the commit and split all these refactors into one commit.

test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
```diff
@@ -12,4 +12,4 @@ deploy_test_env
 pod_name=$(cli_get_pods_test_env | awk '{print $1}')
 
 echo "Expecting secrets provider to fail with error 'CSPFK004E Environment variable 'SECRETS_DESTINATION' must be provided'"
-wait_for_it 600 "$cli logs $pod_name -c cyberark-secrets-provider | grep 'CSPFK004E'"
+$cli "logs $pod_name -c cyberark-secrets-provider | grep CSPFK004E"
```
orenbm: why was this removed?

Jenkinsfile
eladkug: It means i cant run on the same branch two jobs in parallel. Why? since we are using same oc environment. It cause to some of the failures.

Jenkinsfile
eladkug: i agree with you but  i wanted to make it fixable if we want to add another environment in the future.

test/test_cases/TEST_ID_12_no_conjur_secrets_permissions.sh
```diff
@@ -11,4 +11,4 @@ deploy_test_env
 
 echo "Expecting secrets provider to fail with error CSPFK034E Failed to retrieve Conjur secrets"
 pod_name=$(cli_get_pods_test_env | awk '{print $1}')
-wait_for_it 600 "$cli logs $pod_name -c cyberark-secrets-provider | grep 'CSPFK034E'"
+$cli "logs $pod_name -c cyberark-secrets-provider | grep CSPFK034E"
```
orenbm: missing new line

test/2_create_test_app_namespace.sh
eladkug: This is almost all the changes you will see in the files (this is why i didnt split the PR). since i added to cli the retry timeout i have to encapsulate the command i'm working with.

test/utils.sh
```diff
@@ -1,13 +1,41 @@
 #!/bin/bash
-set -euo pipefail
+set -xeuo pipefail
 
 # lookup test-env.sh.yml for explanation.
 export KEY_VALUE_NOT_EXIST=" "
 
+wait_for_it() {
```
orenbm: did this actually change or just moved?

test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
eladkug: i suggest you to go to utils function and you will see the change. all cli command will run wait_for_it as part of the command

test/utils.sh
```diff
@@ -1,13 +1,41 @@
...
+      sleep $spacer
+    done
+  echo 'Success!'
+  fi
+}
+
 if [ $PLATFORM = "kubernetes" ]; then
-    cli=kubectl
+    cli="wait_for_it 600 kubectl"
+    cli_no_timeout=kubectl
```
orenbm: 1. why do we need the separation? When will we need to wait and when won't we need to?
2. I think that we need to change `cli` to `cli_with_timeout`. It's not good to hide such logic.
3. Assuming we do need a separation, is there a reason not to write `wait_for_it 600 $cli` whenever we write it so it's clear to the reader that we wait?

test/utils.sh
```diff
@@ -28,48 +56,19 @@ announce() {
 }
 
 has_namespace() {
-  if $cli get namespace  "$1" > /dev/null; then
+  #We dont need retry because some of the scenarios waiting for false
```
orenbm: add a space after #

orenbm: I don't understand the comment. Can you elaborate?

test/utils.sh
eladkug: moved and changed. main issues:
 eval $*  && return 0 instead of  eval $@ > /dev/null && echo 'Success!' && return 0
1. Because i didnt want to return echo with success because it fail some of the tests that waiting for the right input. 
2. eval $* take all the command as string and not as an array

test/utils.sh
```diff
@@ -28,48 +56,19 @@ announce() {
...
-  fi
-}
-
-
 set_namespace() {
   if [[ $# != 1 ]]; then
     printf "Error in %s/%s - expecting 1 arg.\n" "$(pwd)" $0
     exit 1
   fi
-
```
orenbm: why was this line removed?

test/utils.sh
```diff
@@ -219,13 +218,12 @@ yaml_print_key_name_value () {
...
 }
 
 verify_secret_value_in_pod () {
   pod_name=$1
   secret_name=$2
   expected_value=$3
-  wait_for_it 600 $cli exec -n "$TEST_APP_NAMESPACE_NAME ${pod_name} printenv
-     | grep $secret_name | cut -d '=' -f 2 | grep $expected_value"
+  $cli "exec -n $TEST_APP_NAMESPACE_NAME ${pod_name} printenv| grep $secret_name | cut -d '=' -f 2 | grep $expected_value"
 }
```
orenbm: missing new line

Jenkinsfile
orenbm: on the same git branch? So if we run 2 builds on the same branch they will wait in a queue?
I'd like to get @hughsaunders 's take on this.

test/utils.sh
eladkug: 1. all the command will run with retry unless you handle / waiting for failure currently there is one command using the timeout (has_namespace and i added comment there)
2. sounds reasonable i will change it
3. User will have to know if use retry or not.. i want it will be the default act otherwise we will need to update each command and make sure new commands have the retry function as well. This is what was done in the past some of the commands have the retry and pass and the commands without the retry sometime failed.

Jenkinsfile
orenbm: which other environment will we add? We can go to git history and get the syntax from there. 
The motivation for the separation is so there are no parallel jobs running on the same OCP/K8S environment? I believe we should add a comment on that in the Jenkinsfile itself as it's quite a big decision.

test/utils.sh
eladkug: didnt see a reason for empty line after if(other function doesnt contain empty line after if)

test/2_create_test_app_namespace.sh
orenbm: Still I believe it should be separated, and the commit message can explain why this is needed. Also - it will make the review much easier as the reviewer will understand the commit, and just skip it instead of going over all of these lines looking for a change other than adding quotes.

test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
orenbm: saw it :) Asked some questions there.

test/utils.sh
orenbm: why was it moved? makes it a bit harder to review it.

orenbm: > Because i didnt want to return echo with success because it fail some of the tests that waiting for the right input.

can you elaborate? maybe give an example?  I don't follow

Jenkinsfile
eladkug: if not.. you will run on OC environment more than one job and we can see gcloud crashed again.

test/utils.sh
eladkug: i had too, because i insert this function under cli and it didn't recognise it

test/utils.sh
eladkug: When we run retry it will do the same function until it pass or at the end will fail. This specific command handle failure by itself. (If i dont have namespace it will return false and its o.k.) if i will add retry it will not be relevant. clear ?

test/utils.sh
orenbm: 1. Sorry, i don't understand. If I understand correctly, `wait_for_it 600` waits *up to* 600 seconds. Thus, if a command doesn't need 600 seconds to run, it won't want 600 seconds. I still don't understand the case where waiting will cause an issue.
2. thanks.
3. I agree with your approach. If it's clear that there is `cli_with_timeout` & `cli_without_timeout` then it's ok to have the actual timeout logic inside. Please rename `cli_no_timeout` to `cli_without_timeout`

test/utils.sh
orenbm: I think the extra line looks better but if this pattern fits the rest of the file then it;s ok.

Jenkinsfile
eladkug: o.k i will change it to hard-coded i dont see a big issue here..

Jenkinsfile
orenbm: I see. Please add a comment explaining why this line is important.

test/2_create_test_app_namespace.sh
eladkug: i will do this for the next time because each change i'm adding i'm verifying 5 times to make sure no issues. and since almost all the files contain 1 change i guess it will be o.k. 
Thanks for this comment.

test/utils.sh
orenbm: I see. So in this case `false` is not a failure and `wait_for_it` will run for no reason.

orenbm: Please change comment to `We don't need a timeout here as `false` is a valid output. Running with a timeout will run this command repeatedly for no reason, ending with the same result.`

Jenkinsfile
orenbm: Thanks.

test/utils.sh
eladkug: right.

test/2_create_test_app_namespace.sh
orenbm: ok. However, if you're already making changes you'll need to push the commits so changing the original commits will not start any additional build.

test/utils.sh
eladkug: 1. The major issue in this job that you run command which suppose to work and from some reason it return no route problem. When i run the retry it will run the command again and will pass because the "no route" issue is not real one.
example 
`[2020-03-12T12:03:51.811Z] error: error when creating "STDIN": Post https://openshift-310.itci.conjur.net:8443/apis/apps.openshift.io/v1/namespaces/test-app-5-77cc69bd-c/deploymentconfigs: dial tcp 18.215.106.98:8443: connect: no route to host`
with retry the command will run again and pass

orenbm: which is good? do we want it to pass in the second time?

test/test_cases/TEST_ID_12_no_conjur_secrets_permissions.sh
eladkug: 👍

Jenkinsfile
eladkug: 👍

Jenkinsfile
eladkug: 👍

test/utils.sh
eladkug: 👍

test/utils.sh
eladkug: 👍

test/utils.sh
eladkug: 👍

eladkug:
Added all changes.
3 failures out of 15 - failed on `Error response from daemon: Get https://registry.tld/v2/conjur/manifests/1-stable: unauthorized: authentication required`
not related to those changes came from kubernetes-conjur-deploy.

Jenkinsfile
```diff
@@ -5,6 +5,9 @@ pipeline {
 
   options {
     timestamps()
+    /* we want to avoid running in parallel otherwise we can get gcloud crashed : database is locked
```
orenbm: is `gcloud crashed : database is locked` the error? If so, please surround by quotes. 

Also, use two lines of `//` instead of `/*`

orenbm: `We want to avoid running in parallel. When we have 2 build running on the same environment (.e.g oc 3.11) in parallel, we get the error "gcloud crashed : database is locked"`

@eladkug does this happen only in GKE or also in OCP? If so, state that in the comment above.

Jenkinsfile
```diff
@@ -31,26 +34,44 @@ pipeline {
...
+            tasks["Openshift v3.10, oss"] = {
+                sh "./bin/test_integration --docker --oss --oc310"
             }
-          }
           parallel tasks
         }
       }
     }
 
+    stage ("Run Integration Tests on dap") {
```
orenbm: can you change `dap` to `DAP` throughout the file? except for `--dap` of course.

Jenkinsfile
eladkug: 👍

Jenkinsfile
eladkug: 👍

test/utils.sh
hughsaunders: I know we already discussed this, but just noting that bash-lib has a good function for this.

test/2_create_test_app_namespace.sh
```diff
@@ -6,7 +6,7 @@ set -euo pipefail
 announce "Creating Test App namespace."
 
 if [[ $PLATFORM == openshift ]]; then
-  oc login -u $OPENSHIFT_USERNAME
+  $cli_with_timeout "login -u $OPENSHIFT_USERNAME"
```
hughsaunders:
```suggestion
  $cli_with_retry "login -u $OPENSHIFT_USERNAME"
```

test/utils.sh
eladkug: i saw it but since it was already in use and implement i used this one. kubernetes-conjur-deploy use this function as well do i didnt want to do dramatic changes...

test/2_create_test_app_namespace.sh
eladkug: i'm no sure retry is the right name here since we give timeout as parameter and not the retry number it may confuse the user.

PR #76, opened 2020-03-26
update pipe commands with retry and update kubectl version

eladkug:
- update commands with pipe which didnt worked well with retry and - those command didnt worked well before with the retry, since the retry command need to be for all the command.

- update kubectl to latest stable - our version was out of date changed from 9.7 to 10.8

The commit is updated with those 2 changes.

### DIFF
--- a/test/2_create_test_app_namespace.sh
+++ b/test/2_create_test_app_namespace.sh
@@ -33,7 +33,7 @@ if [[ "$PLATFORM" = "openshift" ]]; then
     TEST_DIR="config/openshift"
 fi
 
-./$TEST_DIR/test-app-conjur-authenticator-role-binding.sh.yml | $cli_with_timeout "create -f -"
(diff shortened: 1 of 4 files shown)


test/2_create_test_app_namespace.sh
```diff
@@ -33,7 +33,7 @@ if [[ "$PLATFORM" = "openshift" ]]; then
     TEST_DIR="config/openshift"
 fi
 
-./$TEST_DIR/test-app-conjur-authenticator-role-binding.sh.yml | $cli_with_timeout "create -f -"
+wait_for_it 600  "./$TEST_DIR/test-app-conjur-authenticator-role-binding.sh.yml | $cli_without_timeout create -f -"
```
hughsaunders: Did this fail on retries because the input could only be read once?

eladkug: it read the yml file and than run the retry only on  create command, after it failed once it tried to create with no yml file

PR #78, opened 2020-03-30
Add support for complex secrets

saprette:
The issue #77 is in the K8 go client, the merge logic does not work with complex values.
Using a secret update instead of a patch solve the issue.

I don't have much time to update your unit tests / integration tests so you can close this PR and maybe it'll at least help you to fix the issue faster.

I did my tests with a complex json document and an ssh key

Issue #77, opened 2020-03-27
Support for complex secrets


saprette:
Hello,

cyberark/secrets-provider-for-k8s seems to have issues understanding complex secrets such as ssh keys or json documents.

Some details on the environment I'm using :

- OpenShift 4.3.8
- Conjur OSS 1.5.0
- cyberark/secrets-provider-for-k8s latest (sha256:e207dfe8b80425d1f1acfe84001f3418e4c69b82c401d5ec601dee03df511e5b)

Setup the required policies to be able to use the authenticator kubernetes-authenticator-client in the namespace argocd.

A variable someenv/hf333ocp/artifactory-pull-secret/dockerconfigjson is set ->
```
root@conjur-configure-pwz4hziqh9-jbwnp:/# conjur variable values add 'someenv/hf333ocp/artifactory-pull-secret/dockerconfigjson' 'some_simple_secret_value'
Value added
root@conjur-configure-pwz4hziqh9-jbwnp:/# conjur variable value someenv/hf333ocp/artifactory-pull-secret/dockerconfigjson
some_simple_secret_value
```
Using this test secret as starting state of the secret, and this simple job.
```
---
kind: Secret
apiVersion: v1
metadata:
  name: test-credentials
type: Opaque
stringData:
  conjur-map: |-
    .dockerconfigjson: |-
      someenv/hf333ocp/artifactory-pull-secret/dockerconfigjson
---
apiVersion: batch/v1
kind: Job
metadata:
  name: conjur-test-provider-1
spec:
  activeDeadlineSeconds: 6000
  template:
    spec:
      serviceAccountName: kubernetes-authenticator-client
      containers:
        - image: 'cyberark/secrets-provider-for-k8s'
          imagePullPolicy: IfNotPresent
          name: kubernetes-authenticator-client
          env:
            - name: DEBUG
              value: 'true'
            - name: CONTAINER_MODE
              value: init
            - name: MY_POD_NAME
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.name
            - name: MY_POD_NAMESPACE
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
            - name: MY_POD_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
            - name: CONJUR_VERSION
              value: '5'
            - name: CONJUR_APPLIANCE_URL
              value: "someOCPRoute"
[...]

PR #79, opened 2020-04-06
Provide complex secrets

orenbm:
This PR was originally created by @saprette [here](https://github.com/cyberark/secrets-provider-for-k8s/pull/78). Please go over it to get more details.

In order to merge it to master we need to first merge it into a side-branch so the Jenkins build will run on it.


test/test_cases/TEST_ID_16_providing_complex_secret_successfully.sh
```diff
@@ -0,0 +1,24 @@
...
+set_namespace "$TEST_APP_NAMESPACE_NAME"
+
+deploy_test_env
+
+echo "Verifying pod test_env has environment variable 'TEST_SECRET' with value '$secret_value'"
+pod_name=$(cli_get_pods_test_env | awk '{print $1}')
+verify_secret_value_in_pod "$pod_name" TEST_SECRET "$secret_value"
+
+# set the secret back to original one for the rest of the tests
+set_namespace "$CONJUR_NAMESPACE_NAME"
```
sigalsax, line 21: I am not sure we need this last part because for every test we run 
```
  for filename in ./$TEST_NAME_PREFIX*.sh; do
      announce "Running '$filename'."
      ./test_case_setup.sh
      $filename
      ./test_case_teardown.sh
      announce "Test '$filename' ended successfully"
  done
```
so the teardown is happening after each test execution and looking at `TEST_ID_1`, it looks similar to this test and this isn't performed so it might not be needed. What do you think?

test/test_cases/TEST_ID_16_providing_complex_secret_successfully.sh
```diff
@@ -0,0 +1,24 @@
...
+$cli_with_timeout "exec $conjur_cli_pod -- conjur variable values add secrets/test_secret supersecret"
+set_namespace "$TEST_APP_NAMESPACE_NAME"
+
```
sigalsax, line 24: The[ original issue](https://github.com/cyberark/secrets-provider-for-k8s/issues/77) also mentioned a similar problem with an ssh key.

> Note that the same kind of parsing error occurs for an ssh key (but this time complaining about a \r character)

What do you say we also add a test for an SSH key?

sigalsax, line 24: Please also change TEST_5 and TEST_6 to say "update" instead of patch"

For TEST_6, the following string is present: 
"echo "Expecting secrets provider to fail with error 'CSPFK005D Failed to _patch_ k8s secret. Reason:...'"

It should be updated to "..._update_ k8s secret" 

For test 5, "export SECRET_CLUSTER_ROLE_VERBS_VALUE="[ \"patch\" ]"", I imagine it should be "update" but the tests pass because we are only testing if "get" permissions are not present

test/test_cases/TEST_ID_16_providing_complex_secret_successfully.sh
orenbm: You are correct, thanks!

test/test_cases/TEST_ID_16_providing_complex_secret_successfully.sh
orenbm: thanks

orenbm: Thanks, adding one.

orenbm:
> Add integration test to validate that we do not changing any other data in the K8S Secret resource in the update process, for example db-name: oracle, which is not part of conjur-map

thanks @nessiLahav ! i added a test for that, can you review [the commit](https://github.com/cyberark/secrets-provider-for-k8s/pull/79/commits/5e635751672e54f8a5ed0625ccf832b38ce0085d)?

orenbm:
@shaharglazner this PR changes the permissions of the service-account running the pod from `[get, patch]` to `[get, update]`.

do you see any security concern here? I don't.

CHANGELOG.md
```diff
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   (cyberark/conjur-authn-k8s-client#70)
 
 ### Fixed
-- Fixed issue with providing complex Conjur secrets (#77)
+- Fixed issue with providing complex Conjur secrets (#77). The secrets-provider
```
sigalsax, line 14: I don't think complex conjur secrets is enough to explain what was done?

maybe add "Fixed issue with providing complex Conjur secrets such as JSON and ssh secrets..."

orenbm:
```
- Fixed issue with providing complex Conjur secrets (e.g JSON, SSH keys)
  [cyberark/secrets-provider-for-k8s#77](https://github.com/cyberark/secrets-provider-for-k8s/issues/77). 
  The secrets-provider now updates k8s secrets using `update` instead of `patch` so the service-account
  needs to have that permission.
```

WDYT?

PR #80, opened 2020-04-12
Remove CONJUR_VERSION env var as it is not needed anymore

orenbm:
We are now consuming a newer version of conjur-authn-k8s-client that doesn't require the CONJUR_VERSION env var, and it will be defaulted to `5` if not present.

As we test only in version 5 we should remove this env var from our test env to verify
it is not required.

### DIFF
--- a/test/config/k8s/test-env.sh.yml
+++ b/test/config/k8s/test-env.sh.yml
@@ -63,9 +63,6 @@ spec:
               fieldRef:
                 fieldPath: status.podIP
 
-          - name: CONJUR_VERSION
-            value: '5'
-
           - name: CONJUR_APPLIANCE_URL
             value: ${CONJUR_APPLIANCE_URL}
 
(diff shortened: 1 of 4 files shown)


orenbm:
@sigalsax we need that for `kubernetes-conjur-deploy` that still requires us to set `CONJUR_VERSION`.

The main issue in this PR is that we do not use the `CONJUR_VERSION` variable in the secrets-provider pod, ensuring that our product doesn't require it.

sigalsax:
> @sigalsax we need that for `kubernetes-conjur-deploy` that still requires us to set `CONJUR_VERSION`.
> 
> The main issue in this PR is that we do not use the `CONJUR_VERSION` variable in the secrets-provider pod, ensuring that our product doesn't require it.

So we still need it in the repo for testing (see `test/1_check_dependencies.sh` and `test/3_load_conjur_policies.sh`) but not in the deployed pods?

orenbm:
> So we still need it in the repo for testing (see test/1_check_dependencies.sh and test/3_load_conjur_policies.sh) but not in the deployed pods?

correct.

PR #81, opened 2020-04-19
Update demo scripts to work to completion

sigalsax:
Adding GKE integration tests from previous project lead to breaking changes in the demo script for the repo. This restores the demo script, allowing users to get E2E with a running demo application.


sigalsax:
Closing because we are removing support for maintaining demo environment

PR #83, opened 2020-04-27
Remove redundant CONJUR_VERSION parameter

orenbm:
The parameter `CONJUR_VERSION` is no longer required by
`kubernetes-conjur-deploy` and can be removed from this
project


orenbm:
> CONJUR_VERSION

Are you sure? I can't see it. in which file?

PR #94, opened 2020-05-24
Add ability to build the secrets-provider with a multi-stage Dockerfile

sigalsax:
This PR adds the ability to build the image in a multi-stage fashion in order to enhance readability instead of separating build process into separate files.


bin/build
```diff
@@ -6,23 +6,14 @@ set -e
...
+echo "---"
 
 echo "Building secrets-provider-for-k8s:$FULL_VERSION_TAG Docker image"
-docker build -f Dockerfile.scratch \
-             --tag "secrets-provider-for-k8s:dev" \
-             --tag "secrets-provider-for-k8s:${FULL_VERSION_TAG}" \
-             --tag "secrets-provider-for-k8s:latest" \
-             .
+
+docker build --tag "secrets-provider-for-k8s:dev" \
```
orenbm, line 13: i would split the command after `build` so all the tags are in the same line

Dockerfile
```diff
@@ -1,13 +1,45 @@
...
 RUN apt-get update && apt-get install -y jq
 
 RUN go get -u github.com/jstemmer/go-junit-report && \
     go get github.com/smartystreets/goconvey
 
-ENV GOOS=linux
-ENV GOARCH=amd64
+RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 
-WORKDIR /opt/secrets-provider-for-k8s
```
orenbm, line 12: don't we need the `WORKDIR`?

Dockerfile
```diff
@@ -1,13 +1,45 @@
-FROM golang:1.13
+FROM golang:1.13 as secrets-provider-builder
 MAINTAINER Conjur Inc
```
orenbm, line 2: please change the maintainer to `CyberArk Software, Ltd.`

Dockerfile
```diff
@@ -1,13 +1,45 @@
...
 
-WORKDIR /opt/secrets-provider-for-k8s
-EXPOSE 8080
+# =================== BUSYBOX LAYER ===================
+# this layer is used to get binaries into the main container
+FROM busybox
+
+# =================== MAIN CONTAINER ===================
+FROM scratch as secrets-provider
+MAINTAINER CyberArk Software, Inc.
```
orenbm, line 25: please change the maintainer to `CyberArk Software, Ltd.`

Dockerfile
```diff
@@ -1,13 +1,45 @@
...
+COPY --from=busybox /bin/mkdir /bin/mkdir
+COPY --from=busybox /bin/chmod /bin/chmod
+COPY --from=busybox /bin/cat /bin/cat
+
+# allow anyone to write to this dir, container may not run as root
+RUN mkdir -p /etc/conjur/ssl && chmod 777 /etc/conjur/ssl
+
+ADD secrets-provider /bin
+
+COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /bin
```
orenbm, line 43: let's use COPY or ADD throughout the file: https://nickjanetakis.com/blog/docker-tip-2-the-difference-between-copy-and-add-in-a-dockerile

bin/build
```diff
@@ -6,23 +6,14 @@ set -e
...
-
-docker build --tag secrets-provider-for-k8s-go:builder .
-
-docker run --rm \
-           -v $PWD:/opt/secrets-provider-for-k8s \
-           secrets-provider-for-k8s-go:builder env CGO_ENABLED=0 GOOS=linux \
-                                              go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
+echo "---"
 
 echo "Building secrets-provider-for-k8s:$FULL_VERSION_TAG Docker image"
```
orenbm, line 11: let's remove `:$FULL_VERSION_TAG` as we build also other tags in this command

Dockerfile
sigalsax: good catch!

Dockerfile
```diff
@@ -1,13 +1,45 @@
...
 
 RUN apt-get update && apt-get install -y jq
 
 RUN go get -u github.com/jstemmer/go-junit-report && \
     go get github.com/smartystreets/goconvey
 
-ENV GOOS=linux
-ENV GOARCH=amd64
-
 WORKDIR /opt/secrets-provider-for-k8s
```
orenbm, line 17: let's move to be under the `COPY` line for better context

Dockerfile
```diff
@@ -1,13 +1,45 @@
-FROM golang:1.13
-MAINTAINER Conjur Inc
+FROM golang:1.13 as secrets-provider-builder
+MAINTAINER CyberArk Software, Inc.
```
orenbm, line 2: It's `CyberArk Software Ltd.`. not `Inc.`

Dockerfile
```diff
@@ -1,13 +1,45 @@
...
+    GOARCH=amd64 \
+    CGO_ENABLED=0
+
+COPY . /opt/secrets-provider-for-k8s
+
+WORKDIR /opt/secrets-provider-for-k8s
+
+EXPOSE 8080
+
+#RUN apt-get update && apt-get install -y jq
```
orenbm, line 14: what's this commented-out line?

sigalsax: missed that in between commits. Should be updated now

PR #95, opened 2020-05-24
Run the secrets-provider with a limited user instead of with root

sigalsax:
- A container running as root is not a good security practice so this should be avoided
- Add group permissions in Conjur directories to avoid UID mismatch upon deploying to Openshift

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 FROM busybox
 
 # =================== MAIN CONTAINER ===================
-FROM scratch as secrets-provider
+FROM alpine:3.11 as secrets-provider
 MAINTAINER CyberArk Software, Ltd.
 
 # copy a few commands from busybox
@@ -37,9 +37,30 @@ COPY --from=busybox /bin/mkdir /bin/mkdir
 COPY --from=busybox /bin/chmod /bin/chmod
 COPY --from=busybox /bin/cat /bin/cat
 
-# allow anyone to write to this dir, container may not run as root
-RUN mkdir -p /etc/conjur/ssl && chmod 777 /etc/conjur/ssl
+RUN apk add -u shadow libc6-compat && \
+    # Add limited user
+    groupadd -r secrets-provider \
+             -g 777 && \
+    useradd -c "secrets-provider runner account" \
+            -g secrets-provider \
+            -u 777 \
+            -m \
+            -r \
+            secrets-provider && \
+    # Ensure plugin dir is owned by secrets-provider user
+    mkdir -p /usr/local/lib/secrets-provider /etc/conjur/ssl /run/conjur && \
+    # Use GID of 0 since that is what OpenShift will want to be able to read things
+    chown secrets-provider:0 /usr/local/lib/secrets-provider \
+                           /etc/conjur/ssl \
+                           /run/conjur && \
+    # We need open group permissions in these directories since OpenShift won't
+    # match our UID when we try to write files to them
+    chmod 770 /etc/conjur/ssl \
+              /run/conjur
 
-COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /bin
+USER secrets-provider
 
-CMD ["secrets-provider"]
+COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /usr/local/bin/
+
+ENTRYPOINT [ "/usr/local/bin/secrets-provider"]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM golang:1.13 as secrets-provider-builder
-MAINTAINER CyberArk Software, Ltd.
+MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux \
     GOARCH=amd64 \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -37,9 +37,29 @@ COPY --from=busybox /bin/mkdir /bin/mkdir
...
+    chmod 770 /etc/conjur/ssl \
+              /run/conjur
 
-COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /bin
+USER secrets-provider
 
-CMD ["secrets-provider"]
+COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /usr/local/bin/
+
+ENTRYPOINT [ "/usr/local/bin/secrets-provider"]
```
orenbm, line 65: missing new line

Dockerfile
```diff
@@ -3,7 +3,7 @@ MAINTAINER CyberArk Software, Ltd.
 
 ENV GOOS=linux \
     GOARCH=amd64 \
-    CGO_ENABLED=0
+    CGO_ENABLED=1
```
orenbm, line 6: why do we need `cgo_enabled`? are you sure it is required?

Dockerfile
orenbm, line 65: why did you change from `CMD` to `ENTRYPOINT`? can you challenge this?

Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 FROM busybox
 
 # =================== MAIN CONTAINER ===================
-FROM scratch as secrets-provider
+FROM alpine:3.8 as secrets-provider
```
orenbm, line 26: why do we use 3.8? why not `latest`? why not a newer version?

i'm not saying it's incorrect, just want to verify you put thought into it and chose the best one.

orenbm:
i actually prefer the title `Run the secrets-provider with a limited user instead with root` as it better says what the change was

Dockerfile
sigalsax: CMD sets default command and can be overwritten from command line when docker container runs (with `docker run -it <image> bash`). ENTRYPOINT command and parameters will not be overwritten from command line. So I guess in our case it doesn't really matter but thought I would go with a command that cannot be overwritten

Dockerfile
sigalsax: good point. i did it for uniformity across projects but 3.11 is available and the local build passes so I will use that one

Dockerfile
sigalsax: I am not sure about this one. Other projects used it compile source code in Go and C

orenbm:
@sigalsax please add a changelog entry for this change.

Dockerfile
sigalsax: do you think we need this offering here?

orenbm: please don't enable cgo if we don't have to. if the build passes without it then leave it like that.

Dockerfile
```diff
@@ -1,9 +1,8 @@
 FROM golang:1.13 as secrets-provider-builder
-MAINTAINER CyberArk Software, Ltd.
+MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux \
-    GOARCH=amd64 \
-    CGO_ENABLED=0
```
orenbm, line 6: why was this removed? it should still be here (although it's the default value, it's best to have it for explicitness)

Dockerfile
```diff
@@ -23,8 +22,8 @@ RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 FROM busybox
 
 # =================== MAIN CONTAINER ===================
-FROM scratch as secrets-provider
-MAINTAINER CyberArk Software, Ltd.
+FROM alpine:3.11 as secrets-provider
```
orenbm, line 25: why not use `latest`?

Dockerfile
```diff
@@ -37,9 +36,30 @@ COPY --from=busybox /bin/mkdir /bin/mkdir
...
+    # match our UID when we try to write files to them
+    chmod 770 /etc/conjur/ssl \
+              /run/conjur
 
-COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /bin
+USER secrets-provider
+
+COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /usr/local/bin/
+
+CMD [ "/usr/local/bin/secrets-provider"]
```
orenbm, line 64: so `CMD` is better than `ENTRYPOINT`? it seemed that the latter is safer from your previous explanation.

Dockerfile
sigalsax: because I don't like the idea of automatically pulling latest for the same reason you thought; risky. But open to `latest` if you think this is the way to go

orenbm: i think that our nightly automation is good enough but i can live with 3.11.

Dockerfile
sigalsax: I didn't want secrets-provider to be overridden but CMD is safer...

Dockerfile
sigalsax: safer is CMD I would say but I was thinking that we didn't want `secrets-provider` to be overriden because this is the purpose of the build, to run the application. Changing it back because CMD _is_ safer.

sigalsax: ENTRYPOINT cannot be overriden and CMD can. What do you means in terms of "Safer"?

Issue #96, opened 2020-05-24
Send slack notification on master failure only


eladkug:
We  get notification for any failure on secrets-provider-for-k8s-owners to our group.
We want to get failure notification from master only.

Dod:

- [x] Verify failure sent to group only from master

- [x] Run test from master and from PR (since its jenkinsfile we can do the test from master as well)

PR #97, opened 2020-05-24
Send slack notification on master failure only

eladkug:
We want to notify failure to all group only when it run on master.

I run the following test:
PR - success
PR - failure (slack message wasent send to group channel)
Master - success
Master - failure (slack message sent to group channel)

Related to #96

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -101,7 +101,13 @@ pipeline {
       cleanupAndNotify(currentBuild.currentResult)
     }
     unsuccessful {
-      cleanupAndNotify(currentBuild.currentResult, "#development", "@secrets-provider-for-k8s-owners")
+        script {
+            if (env.BRANCH_NAME == 'master') {
+              cleanupAndNotify(currentBuild.currentResult, "#development", "@secrets-provider-for-k8s-owners")
+            } else {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -101,7 +101,13 @@ pipeline {
       cleanupAndNotify(currentBuild.currentResult)
     }
     unsuccessful {
-      cleanupAndNotify(currentBuild.currentResult, "#development", "@secrets-provider-for-k8s-owners")
+        script {
+            if (env.BRANCH_NAME == 'master') {
+              cleanupAndNotify(currentBuild.currentResult, "#development", "@secrets-provider-for-k8s-owners")
+            } else {
+              cleanupAndNotify(currentBuild.currentResult, "#development", "")
```
orenbm, line 108: this can be `cleanupAndNotify(currentBuild.currentResult, "#development")` as `""` is the dedault value for the additional message, as can be seen in the implementation for [cleanupAndNotify](https://github.com/conjurinc/jenkins-pipeline-library/blob/master/vars/cleanupAndNotify.groovy).

You can technically also omit the `#development` but i'd leave it for explicitness.

Issue #99, opened 2020-05-27
Add code coverage results in Jenkins


eladkug:
We want to verify our code coverage in the project and get the output results for each build.

- [x] Connect code climate to this project

- [x] Add support via Jenkins to our automation

PR #101, opened 2020-05-30
Add code coverage 

eladkug:
We run unit tests as part of our repo. Its important to see our code coverage statistics to improve our self. 
For each build we will see the status on Jenkins.
This PR refer to #99

### DIFF
--- a/Dockerfile.junit
+++ b/Dockerfile.junit
@@ -11,4 +11,8 @@ RUN apk add -u curl \
                musl-dev \
                bash
 
-RUN go get -u github.com/jstemmer/go-junit-report
+# gocov converts native coverage output to gocov's JSON interchange format
+# gocov-xml converts gocov format to XML for use with Jenkins/Cobertura
+RUN go get -u github.com/jstemmer/go-junit-report && \
+    go get github.com/axw/gocov/gocov && \
+    go get github.com/AlekSi/gocov-xml
--- a/bin/test_unit
+++ b/bin/test_unit
@@ -2,20 +2,17 @@
 
 set -eox pipefail
 
-rm -rf unit-test
-mkdir unit-test
-
-junit_output_file="unit-test/junit.output"
-
-rm -f "$junit_output_file"
-touch "$junit_output_file"
+junit_output_file="./junit.output"
 
 echo "Building unit test image..."
 docker build -f Dockerfile.test -t secrets-provider-for-k8s-test-runner:latest .
 
 echo "Running unit tests..."
 set +e
-  docker run --rm -t secrets-provider-for-k8s-test-runner:latest \
+  docker run --rm -t \
+             --volume "$PWD"/:/secrets-provider-for-k8s/test/ \
+             secrets-provider-for-k8s-test-runner:latest \
+             -coverprofile="./test/c.out" \
              ./cmd/... \
              ./pkg/... \
              | tee -a "$junit_output_file"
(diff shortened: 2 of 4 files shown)


Dockerfile.junit
```diff
@@ -11,4 +11,8 @@ RUN apk add -u curl \
                musl-dev \
                bash
 
-RUN go get -u github.com/jstemmer/go-junit-report
+# gocov => converts native coverage output to gocov's JSON interchange format
+# gocov-xml => converts gocov format to XML for use with Jenkins/Cobertura
```
orenbm, line 15: can we remove the rockets and just write sentences. 
```
gocov converts native coverage output to gocov's JSON interchange format.
gocov-xml converts gocov format to XML for use with Jenkins/Cobertura
```

bin/test_unit
```diff
@@ -36,3 +39,11 @@ docker run --rm \
...
     cat ./junit.output | go-junit-report > ./junit.xml
   "
+
+# format coverage output XML
+docker run --rm \
+  -v $PWD/:/test \
+  secrets-provider-for-k8s-junit:latest \
+  bash -exc "
+    gocov convert ./c.out | gocov-xml > ./coverage.xml
+  "
```
orenbm, line 49: can't we combine these two? So the already existing container will run both commands instead of creating 2 of them?

bin/test_unit
```diff
@@ -15,7 +15,10 @@ docker build -f Dockerfile.test -t secrets-provider-for-k8s-test-runner:latest .
 
 echo "Running unit tests..."
 set +e
-  docker run --rm -t secrets-provider-for-k8s-test-runner:latest \
+  docker run -t \
```
orenbm, line 18: did you remove the `--rm` on purpose?

eladkug: good call , my mistake..

bin/test_unit
```diff
@@ -2,20 +2,17 @@
...
 echo "Running unit tests..."
 set +e
-  docker run --rm -t secrets-provider-for-k8s-test-runner:latest \
+  docker run --rm -t \
+             --volume "$PWD"/:/secrets-provider-for-k8s/test/ \
+             secrets-provider-for-k8s-test-runner:latest \
+             -coverprofile="./test/c.out" \
              ./cmd/... \
              ./pkg/... \
              | tee -a "$junit_output_file"
```
orenbm, line 18: what happens if we run this script twice? will it delete the output of the previous run?

eladkug: Yes. I verified it.(relevant for local test only)

Dockerfile.junit
```diff
@@ -11,4 +11,8 @@ RUN apk add -u curl \
                musl-dev \
                bash
 
-RUN go get -u github.com/jstemmer/go-junit-report
+# gocov converts native coverage output to gocov's JSON interchange format
+# gocov-xml converts gocov format to XML for use with Jenkins/Cobertura
+RUN go get -u github.com/jstemmer/go-junit-report && \
+    go get github.com/axw/gocov/gocov && \
```
sigalsax, line 17: do we need to reference this to our NOTICES?

bin/test_unit
```diff
@@ -2,20 +2,17 @@
...
 echo "Building unit test image..."
 docker build -f Dockerfile.test -t secrets-provider-for-k8s-test-runner:latest .
 
 echo "Running unit tests..."
 set +e
-  docker run --rm -t secrets-provider-for-k8s-test-runner:latest \
+  docker run --rm -t \
+             --volume "$PWD"/:/secrets-provider-for-k8s/test/ \
+             secrets-provider-for-k8s-test-runner:latest \
+             -coverprofile="./test/c.out" \
```
sigalsax, line 15: is there a reason why `c.out` is the name of this file? Does this hold the coverage report? If so, I would give it a more descriptive name

bin/test_unit
```diff
@@ -28,11 +25,12 @@ echo "Building junit image..."
 
 docker build -f Dockerfile.junit -t secrets-provider-for-k8s-junit:latest .
 
-echo "Creating junit report..."
+echo "Creating junit report and coverage output XML"
```
sigalsax, line 28: I would add "..." at the end of this to conform with other echos

bin/test_unit
eladkug: Thanks.. i agree about the c file but its came from [codeclimate](https://github.com/codeclimate/test-reporter/blob/0ad7d9c70d6b9ebca14a4394d80856391aa81a17/formatters/gocov/gocov.go)

PR #111, opened 2020-06-07
Test variables with spaces

orenbm:
Verify that the secrets-provider can provide Conjur secrets that have spaces in their variable id.

### DIFF
--- a/test/config/k8s/k8s-secret.yml
+++ b/test/config/k8s/k8s-secret.yml
@@ -7,4 +7,5 @@ stringData:
   conjur-map: |-
     secret: secrets/test_secret
     var_with_spaces: secrets/var with spaces
+    var_with_pluses: secrets/var+with+pluses
   non-conjur-key: some-value
--- a/test/config/k8s/k8s-secret.yml
+++ b/test/config/k8s/k8s-secret.yml
@@ -6,4 +6,5 @@ type: Opaque
 stringData:
   conjur-map: |-
     secret: secrets/test_secret
+    var_with_spaces: secrets/var with spaces
   non-conjur-key: some-value
--- a/test/test_cases/TEST_ID_1.3_providing_variables_with_spaces_successfully.sh
+++ b/test/test_cases/TEST_ID_1.3_providing_variables_with_spaces_successfully.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -euxo pipefail
+
+create_secret_access_role
+
+create_secret_access_role_binding
+
+test_secret_is_provided "var_with_spaces_secret" "secrets/var with spaces" "VARIABLE_WITH_SPACES_SECRET"
--- a/test/utils.sh
+++ b/test/utils.sh
@@ -222,24 +222,26 @@ cli_get_pods_test_env () {
 
 test_secret_is_provided () {
   secret_value=$1
+  variable_name="${2:-secrets/test_secret}"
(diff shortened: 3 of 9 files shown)


test/test_cases/TEST_ID_1.3_providing_variables_with_spaces_successfully.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euxo pipefail
+
+create_secret_access_role
+
+create_secret_access_role_binding
+
+test_secret_is_provided "var_with_spaces_secret" "secrets/var with spaces" "VARIABLE_WITH_SPACES_SECRET"
```
eladkug, line 8: new line is missing

test/config/k8s/k8s-secret.yml
```diff
@@ -6,4 +6,5 @@ type: Opaque
 stringData:
   conjur-map: |-
     secret: secrets/test_secret
+    var_with_spaces: secrets/var with spaces
   non-conjur-key: some-value
```
sigalsax, line 10: GENERAL note: you have Geri's commit in your PR. Run `git fetch push ignore` (AKA `git rebase origin/master` git push --force) and it will go away

test/utils.sh
```diff
@@ -222,24 +222,26 @@ cli_get_pods_test_env () {
 
 test_secret_is_provided () {
   secret_value=$1
+  variable_name="${2:-secrets/test_secret}"
```
sigalsax, line 225: What is `#:-`?

`variable_name=secrets/test_secret` is the second parm wasn't supplied?

test/utils.sh
```diff
@@ -222,24 +222,26 @@ cli_get_pods_test_env () {
...
+  environment_variable_name="${3:-TEST_SECRET}"
 
   set_namespace "$CONJUR_NAMESPACE_NAME"
   conjur_cli_pod=$(get_conjur_cli_pod_name)
-  $cli_with_timeout "exec $conjur_cli_pod -- conjur variable values add secrets/test_secret $secret_value"
+  $cli_with_timeout "exec $conjur_cli_pod -- conjur variable values add \"$variable_name\" $secret_value"
```
sigalsax, line 230: isn't `$variable_name` enough here since the value of variable_name is already wrapped in `""`

sigalsax: and also matches with what you have below
`'$environment_variable_name' `

test/config/k8s/k8s-secret.yml
sigalsax: @eladkug ^

test/utils.sh
orenbm: yes. it checks if `$2` exists and if not sets to a default value

test/utils.sh
orenbm: > isn't `$variable_name` enough here since the value of variable_name is already wrapped in `""`

No because this text is then sent to a command that needs the quotes in case of spaces in the variable name.

test/test_cases/TEST_ID_1.3_providing_variables_with_spaces_successfully.sh
eladkug: still missing new line..

orenbm: no it's not

PR #128, opened 2020-06-23
Add/Update CODEOWNERS

JakeQuilty:
Add line to CODEOWNES, or add file if it doesn't exist. Issuse https://github.com/cyberark/community/issues/70

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,8 @@
+# Changes to .trivyignore require Security Architect approval
+.trivyignore @cyberark/security-architects @conjurinc/security-architects @conjurdemos/security-architects
+
+# Changes to .codeclimate.yml require Quality Architect approval
+.codeclimate.yml @cyberark/quality-architects @conjurinc/quality-architects @conjurdemos/quality-architects
+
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,10 @@
+* @cyberark/community-and-integrations-team @conjurinc/community-and-integrations-team @conjurdemos/community-and-integrations-team
```
andytinkham, line 1: @Tovli - does your team have an alias that should go here instead of the C&I team? Ig not, we should probably remove this line.

andytinkham:
> Why do we need all these approvals from a specific person?
> Can't the team approve it? they are the owners of the code and its quality

The team can approve almost all changes themselves. The 3 files that have roles assigned to them have that approval for helping ensure consistency across all repos.  For the .trivyignore, it's also to make sure that Shahar or I understand why a CVE is being ignored and that there's an appropriate comment because one or the other of us is likely going to have to explain why we still have the 3rd party vuln we're ignoring at some point in the future long after anyone who works on it will remember why we're not vulnerable.

JakeQuilty:
Failing test doesn't look related

Issue #126, opened 2020-06-22
Some tests failed on OC3.10


eladkug:
For example: Test TEST_ID_5_no_get_permission_to_secret failed on OC3.10 environment only. (DAP/OSS)
In oc 3.10 we saw:
`INFO: 2020/06/22 09:29:00 requests.go:47: CAKC012I Authn request to: https://conjur-cluster.conjur-2b2ad8c6-c-test.svc.cluster.local/authn-k8s/conjur-2b2ad8c6-c-test/my-account/host%2Fconjur%2Fauthn-k8s%2Fconjur-2b2ad8c6-c-test%2Fapps%2Ftest-app-2b2ad8c6-c%2F%2A%2F%2A/authenticate
INFO: 2020/06/22 09:29:00 authenticator.go:250: CAKC001I Successfully authenticated
INFO: 2020/06/22 09:29:00 k8s_secrets_client.go:54: CSPFK004I Creating Kubernetes client...
INFO: 2020/06/22 09:29:00 k8s_secrets_client.go:19: CSPFK005I Retrieving Kubernetes secret 'test-k8s-secret' from namespace 'test-app-2b2ad8c6-c'...
DEBUG: 2020/06/22 09:29:00 provide_conjur_secrets.go:127: CSPFK009D Processing 'conjur-map' data entry value of k8s secret 'test-k8s-secret'
INFO: 2020/06/22 09:29:00 conjur_secrets_retriever.go:11: CSPFK003I Retrieving following secrets from Conjur: [secrets/var with spaces secrets/var+with+pluses secrets/test_secret]`

In oc3.11 we saw:
`INFO: 2020/06/22 10:18:20 authenticator.go:250: CAKC001I Successfully authenticated
INFO: 2020/06/22 10:18:20 k8s_secrets_client.go:54: CSPFK004I Creating Kubernetes client...
INFO: 2020/06/22 10:18:20 k8s_secrets_client.go:19: CSPFK005I Retrieving Kubernetes secret 'test-k8s-secret' from namespace 'test-app-76a181a0-1'...
DEBUG: 2020/06/22 10:18:20 k8s_secrets_client.go:23: CSPFK004D Failed to retrieve k8s secret. Reason: secrets "test-k8s-secret" is forbidden: User "system:serviceaccount:test-app-76a181a0-1:test-app-76a181a0-1-sa" cannot get secrets in the namespace "test-app-76a181a0-1": no RBAC policy matched
[...]

PR #129, opened 2020-06-23
Skip OC3.10 tests. OC3.10 environment issues.

eladkug:
Related to #126 .
It looks like we have issues with oc3.10 environment, need to rebuild the environment and it may take 1-2 days.
Meanwhile we skipped oc3.10 tests.


garymoon:
RedHat have confirmed that we have to apply for new licenses and the process will take 48 hours. I'll re-enable these tests when that's complete.

PR #137, opened 2020-07-02
Add design for k8s secret label

sigalsax:
As part of Milestone 2 we would like to make an effort to advance customer experience for this product. As part of this, we introduce k8s secrets labels where customers can tell us which secrets to bring from Conjur without having to list them explicitly

### DIFF
--- a/design/k8s_secrets_labels_design.md
+++ b/design/k8s_secrets_labels_design.md
@@ -1,0 +1,218 @@
+# K8s Secrets Labels Design
+
+## Table of Contents
+
+[//]: # "You can use this tool to generate a TOC - https://ecotrust-canada.github.io/markdown-toc/"
+
+## Glossary
+
+| **Term**                                                     | **Description**                                              |
+| ------------------------------------------------------------ | ------------------------------------------------------------ |
+| [Label](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) | Key/value pairs that are attached to K8s objects. They are used to identify and group objects for effective querying and selection. |
+
+## Background
+
+As part of Phase 2 for Secrets Provider for K8s effort, we want to improve the customer experience by offering the ability to use labels for easy identification of K8s secrets whose secret values should come from Conjur. This will allow customers to list in terms of groupings instead of having to list the K8s secrets individually.
+
+## Issue description
+
+At current, the Secrets Provider code looks at the `K8S_SECRETS` environment variable in each of the pod manifests to know which K8s Secret to update with Conjur values. 
+
+Listing each K8s Secret in this way is not a scalable solution, especially for customers with hundreds of K8s Secrets. 
+
+## Solution
+
+We will offer the customer an *additional* option to define which K8s Secrets values are hosted in Conjur by defining them in terms of a label instead of individually. They will define the label that is attached to the K8s Secret and add it under `K8S_SECRETS_LABEL` environment variable in the Secrets Provider manifest. 
+
+The Secrets Provider will then make a separate API call to [K8s](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#secret-v1-core)/[OC](https://docs.openshift.com/container-platform/4.4/rest_api/index.html#list-14), fetching all K8s Secrets with that label. We will use that information to fetch their values in Conjur and update the K8s Secrets.
+
+This enhancement can be used in the current *init solution* and will help build the foundation for later Milestones.
+
+### Design
+
+|      | Fetch/Update K8s Secret                                      | Pros                                                         | Cons                                                         | UX                                                           | Effort estimation |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------- |
+| 1    | Get K8s secrets with the names specified in `K8S_SECRETS`    | - No code changes  - Backwards compatible                    | - Repetitive work <br />- Not scalable                       | List K8S Secrets one-by-one                                  | Free              |
+| 2    | Read all K8s Secrets that are marked with labels specified in `K8S_SECRETS_LABEL` env var | - Scalable grouping  - Less intervention on our part as a customer can define which types of secrets we should parse  - K8s native, as-is solution | - Not general solution so we are still K8s bound - Demands SA has "list" privileges on secrets | Add "list" to SA `verbs: ["get", "update", "list"]` <br />- List label grouping in Secrets Provider manifest<br />- Label key `conjur-secrets-label` will need to be added to each K8s Secret | 5 days            |
+
+*Decision*: Solution #2, Read all K8s Secrets that are marked with labels. It would be the most K8s native and avoids the repetitive work of having to list all the K8s Secrets each app needs. 
+
+### Customer Experience 
+
+A customer would add a label on a K8s Secret like so:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: db-secret
+  labels:
+    conjur-secrets-label: prod # label name can be configurable
+  type: Opaque
+stringData:
+  conjur-map: |-
+    username: secrets-accessors/db_username
+    password: secrets-accessors/db_password
+```
+
+Secrets Provider Manifest with the `K8S_SECRETS_LABEL` environment variable:
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: secrets-provider-job
+  namespace: app-namespace
+spec:
+  template:
+    spec:
+      serviceAccountName: secret-provider-sa
+      containers:
+      - image: secrets-provider:latest
+        name: cyberark-secrets-provider-1
+        env:
+        - name: SECRETS_DESTINATION
+          value: k8s_secrets
+
+        - name: K8S_SECRETS_LABEL
+          value: conjur-secrets-label:prod
+        
+        # K8s authn-client configurations (CONJUR_APPLIANCE_URL, CONJUR_AUTHN_URL, CONJUR_ACCOUNT, CONJUR_SSL_CERTIFICATE) 
+```
+
+These labels will have an OR inclusive relationship, not an AND relationship. The label key will not be configurable and will have to be `conjur-secrets-label`. In other words, to use this enhancement, they will need to add `conjur-secrets-label` as an entry under labels.
+
+Utilizing labels is a more K8s native solution and avoids the repetitive work of having to list all the K8s Secrets each app needs.
+
+To be able to read the labels set on K8s Secrets, the Service Account used for the Secrets Provider will need "list" privileges on K8s Secrets resources.
+
+The Service Account / Role / RoleBinding will be defined as the following:
+
+```yaml
+# Service Account definition 
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: secret-provider-sa
+
+# Role definition
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: app-namespace
+  name: secrets-access
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "update", "list"] # "list" is a new privilege here
+
+# RoleBinding definition to associate the SA with the Role
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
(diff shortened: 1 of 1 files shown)


design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+| ------------------------------------------------------------ | ------------------------------------------------------------ |
+| [Label](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) | Key/value pairs that are attached to K8s objects. They are used to identify and group objects for effective querying and selection. |
+
+## Background
+
+As part of Phase 2 for Secrets Provider for K8s effort, we want to improve the customer experience by offering the ability to use labels for easy identification of K8s secrets whose secret values should come from Conjur. This will allow customers to list in terms of groupings instead of having to list the K8s secrets individually.
+
+## Issue description
+
+At current, the Secrets Provider code looks at the `K8S_SECRETS` environment variable in each of the pod manifests to know which K8s Secret to update with Conjur values.
```
orenbm, line 19: what do you mean by `in each of the pod manifests`? the secrets-provider has one manifest that defines its container and it looks at that manifest.

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Background
+
+As part of Phase 2 for Secrets Provider for K8s effort, we want to improve the customer experience by offering the ability to use labels for easy identification of K8s secrets whose secret values should come from Conjur. This will allow customers to list in terms of groupings instead of having to list the K8s secrets individually.
+
+## Issue description
+
+At current, the Secrets Provider code looks at the `K8S_SECRETS` environment variable in each of the pod manifests to know which K8s Secret to update with Conjur values. 
+
+Listing each K8s Secret in this way is not a scalable solution, especially for customers with hundreds of K8s Secrets.
```
orenbm, line 21: why?

design/k8s_secrets_labels_design.md
orenbm: i think that an example of a manifest snippet can help those who are not familiar with it.

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+
+This enhancement can be used in the current *init solution* and will help build the foundation for later Milestones.
+
+### Design
+
+|      | Fetch/Update K8s Secret                                      | Pros                                                         | Cons                                                         | UX                                                           | Effort estimation |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------- |
+| 1    | Get K8s secrets with the names specified in `K8S_SECRETS`    | - No code changes  - Backwards compatible                    | - Repetitive work <br />- Not scalable                       | List K8S Secrets one-by-one                                  | Free              |
```
orenbm, line 35: i would rephrase to `List K8S Secrets in a comma-separated list`

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+| 2    | Read all K8s Secrets that are marked with labels specified in `K8S_SECRETS_LABEL` env var | - Scalable grouping  - Less intervention on our part as a customer can define which types of secrets we should parse  - K8s native, as-is solution | - Not general solution so we are still K8s bound - Demands SA has "list" privileges on secrets | Add "list" to SA `verbs: ["get", "update", "list"]` <br />- List label grouping in Secrets Provider manifest<br />- Label key `conjur-secrets-label` will need to be added to each K8s Secret | 5 days            |
```
orenbm, line 36: > Label key `conjur-secrets-label` will need to be added to each K8s Secret

only if they don't already have a label they want to use

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+
+*Decision*: Solution #2, Read all K8s Secrets that are marked with labels. It would be the most K8s native and avoids the repetitive work of having to list all the K8s Secrets each app needs.
```
orenbm, line 38: I would really want us to talk with customers or SEs to understand if this experience is better. Both for cases where customers already have labels in their secrets and in cases they don't.

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+
+A customer would add a label on a K8s Secret like so:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: db-secret
+  labels:
+    conjur-secrets-label: prod # label name can be configurable
```
orenbm, line 50: do we need the word `label`? can't it be implicit `conjur-secrets`

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+  type: Opaque
+stringData:
+  conjur-map: |-
+    username: secrets-accessors/db_username
```
orenbm, line 54: nit - you have variables defined in a policy named `secrets-accessors`? that policy seems to hold hosts by its name.

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+    spec:
+      serviceAccountName: secret-provider-sa
+      containers:
+      - image: secrets-provider:latest
+        name: cyberark-secrets-provider-1
+        env:
+        - name: SECRETS_DESTINATION
+          value: k8s_secrets
+
+        - name: K8S_SECRETS_LABEL
```
orenbm, line 77: why not plural?

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+          value: conjur-secrets-label:prod
+        
+        # K8s authn-client configurations (CONJUR_APPLIANCE_URL, CONJUR_AUTHN_URL, CONJUR_ACCOUNT, CONJUR_SSL_CERTIFICATE) 
+```
+
+These labels will have an OR inclusive relationship, not an AND relationship. The label key will not be configurable and will have to be `conjur-secrets-label`. In other words, to use this enhancement, they will need to add `conjur-secrets-label` as an entry under labels.
```
orenbm, line 83: i am confused. the label is not configurable? then what is `conjur-secrets-label:prod`? what is `prod` in this case? is there another valid value to this label?

If not, let's change it to `true`.

Also - if it's not configurable, then why does the customer need to add an env var? why can't we look for it by ourselves without the customer needing to add this value?

orenbm: or do you mean that the value of this label _is_ configurable? where a customer will have one secrets-provider with `conjur-secrets-label:prod` and one with `conjur-secrets-label:staging`?

design/k8s_secrets_labels_design.md
```diff
@@ -0,0 +1,218 @@
...
+| ------------------------------------------------------------ | ------------------------------------------------------------ | ----- |
+| Admin supplies `K8S_SECRETS_LABELS` value                    | Fetching all k8s secrets with label %s in namespace…         | Info  |
+| Admin does not provide `K8S_SECRETS_LABELS` or `K8S_SECRETS` environment variables | Environment variable `K8S_SECRETS_LABELS` or `K8S_SECRETS` must be provided... | Error |
+| Admin provides label that does not return K8s Secrets (empty list) | Failed to retrieve k8s secrets with label %s                 | Error |
+| `K8S_SECRETS_LABELS` key-value pairs has invalid character, K8s API has problem with value (400 error) | Invalid characters in `K8S_SECRETS_LABELS`                   | Error |
+| Secrets Provider was unable to provide K8s secret with Conjur value | Failed to retrieve Conjur secrets. Reason: %s (Already exists) | Error |
+
+## Open questions
+
+1. Should we write documentation for this feature?
```
orenbm, line 196: why of course!

design/k8s_secrets_labels_design.md
sigalsax: As in why is it not scalable? It is not a scalable solution because customers will need to detail each secret that their apps need. So if a customer has 100 applications and each pod needs 2 secrets that is a lot of manual work

design/k8s_secrets_labels_design.md
sigalsax: So we are evaluating a few different directions
1. Supply a key/value pair that the customer cannot configure. Ex: `conjur-secrets-label: true`
2. Supply a `conjur-secrets-label` key that the customer cannot configure but a configurable value. For example: `conjur-secrets-label: prod`
3. Allow the customer to configure both multiple key/value pairs. This is will be problematic for upcoming milestones when we will have to monitor changes in the K8s environment by performing `watch` on all k8s secrets resources. When querying the K8s Watch endpoint, the customer would supply a [comma-separated list which acts as &&](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors). So the following:
```
key: K8S_SECRETS_LABELS
value: config-label1:conjur-key1, config-label2:conjur-key2`
```
would mean that the K8S Secret that would be returned would need to have `config-label1:conjur-key1` **&&** `config-label2:conjur-key2`

_A way to curb this_ would be limit the user to a single key and multiple values. For example:  
 `my-config-conjur-label in (prod, dev)`. K8s gives us the option to use [set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) which gives us OR relationships on the values

```
key: K8S_SECRETS_LABELS
value: my-config-conjur-label in (production, dev)'
```

design/k8s_secrets_labels_design.md
sigalsax: I agree. @InbalZilberman is working on that

design/k8s_secrets_labels_design.md
sigalsax: Yes! It can be anything. I wanted `-label` to make it clear we are looking for labels and not the actual name of the K8s Secret. Which do you feel is most representative? I think there might be a problem with `conjur` in the name because of the DAP offering.
@InbalZilberman what say you?

design/k8s_secrets_labels_design.md
sigalsax: because we wanted to start by supporting only 1 label

design/k8s_secrets_labels_design.md
sigalsax: knew it!

design/k8s_secrets_labels_design.md
orenbm: but what's the alternative? In the scenario you describe, the customer will need to create 50 labels and define them across the k8s secrets manifests. Is that what you mean? Or do you mean that they will have only one label and each secerts-provider will retrieve all of them? 

Can you please describe what you expect the customer to do in the new feature in case of the use-case that you described above?

design/k8s_secrets_labels_design.md
orenbm: In case you go with option 2 then i would change the value of `K8S_SECRETS_LABELS` to have only the values.
if `conjur-secrets-label` is hard-coded then they don't need to add it all the time.

Instead of 
```
key: K8S_SECRETS_LABELS
value: conjur-secrets-label:prod,conjur-secrets-label:staging
```

do
```
key: K8S_SECRETS_LABELS
value: prod,staging
```

design/k8s_secrets_labels_design.md
sigalsax: if a customer has multiple secrets providers, then there would be multiple manifests. I will add an example

design/k8s_secrets_labels_design.md
orenbm: but i really think that this is a complicated solution. I would go with option 1 and let the customer decide if they want segregation of duty between secrets-providers (and then define the secrets by name) or have all the secrets-providers get all the secrets (and then get them by the static label `conjur-secrets-label:true` without the need to define the env var `K8S_SECRETS_LABELS`).

design/k8s_secrets_labels_design.md
orenbm: well, we already have `conjur-map` so that ship has sailed.
i would go with `conjur-secrets`

design/k8s_secrets_labels_design.md
orenbm: but will they change to multiple in the near future? if so, let's not create a future breaking change.

sigalsax:
closing this for now due to deprioritization

PR #146, opened 2020-07-08
Remove redundant RoleBinding for conjur-cluster SA

orenbm:
Connected to cyberark/kubernetes-conjur-deploy#129

We were creating a RoleBinding for the service-account running in
the pod of the DAP follower for our tests, while it makes no sense to
do so.

This was done because the ClusterRole wasn't created properly in `kubernetes-conjur-deploy`
but as it is fixed there we should remove it from here.


orenbm:
This is actually not redundant. Created [this doc issue](https://github.com/cyberark/conjur-docs/issues/1004) to emphasize it in the doc.

PR #148, opened 2020-07-12
Add debugging abilities to secrets provider

abrahamko:
### What does this PR do?
Add ability to build a debug flavor of secrets provider, and a debugging configuration for `GoLand` IDE

Connected to #[143]

### DIFF
--- a/.run/Debug Secrets Provider.run.xml
+++ b/.run/Debug Secrets Provider.run.xml
@@ -1,0 +1,5 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Debug Secrets Provider" type="GoRemoteDebugConfigurationType" factoryName="Go Remote" port="43210">
+    <method v="2" />
+  </configuration>
+</component>
--- a/.run/build debug.run.xml
+++ b/.run/build debug.run.xml
@@ -2,7 +2,7 @@
   <configuration default="false" name="build debug" type="ShConfigurationType">
     <option name="INDEPENDENT_SCRIPT_PATH" value="true" />
     <option name="SCRIPT_PATH" value="$PROJECT_DIR$/bin/build" />
-    <option name="SCRIPT_OPTIONS" value="debug" />
+    <option name="SCRIPT_OPTIONS" value="--debug" />
     <option name="INDEPENDENT_SCRIPT_WORKING_DIRECTORY" value="true" />
     <option name="SCRIPT_WORKING_DIRECTORY" value="$PROJECT_DIR$" />
     <option name="INDEPENDENT_INTERPRETER_PATH" value="true" />
--- a/.run/build debug.run.xml
+++ b/.run/build debug.run.xml
@@ -1,0 +1,13 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="build debug" type="ShConfigurationType">
+    <option name="INDEPENDENT_SCRIPT_PATH" value="true" />
+    <option name="SCRIPT_PATH" value="$PROJECT_DIR$/bin/build" />
+    <option name="SCRIPT_OPTIONS" value="debug" />
+    <option name="INDEPENDENT_SCRIPT_WORKING_DIRECTORY" value="true" />
+    <option name="SCRIPT_WORKING_DIRECTORY" value="$PROJECT_DIR$" />
+    <option name="INDEPENDENT_INTERPRETER_PATH" value="true" />
+    <option name="INTERPRETER_PATH" value="/bin/sh" />
+    <option name="INTERPRETER_OPTIONS" value="" />
+    <method v="2" />
+  </configuration>
+</component>
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# =================== BASE BUILD LAYER ===================
+# this layer is used to prepare a common layer for both debug and release builds
 FROM golang:1.13 as secrets-provider-builder-base
 MAINTAINER CyberArk Software Ltd.
 
@@ -17,6 +19,7 @@ EXPOSE 8080
 COPY go.mod .
 COPY go.sum .
 
+# Add a layer of prefetched modules so in case of rebuild the modules are already cached
 RUN go mod download
 
 # =================== RELEASE BUILD LAYER ===================
@@ -45,6 +48,7 @@ RUN go build -a -installsuffix cgo -gcflags="all=-N -l" -o secrets-provider ./cm
 FROM busybox
 
 # =================== BASE MAIN CONTAINER ===================
+# this layer is used to prepare a common layer for both debug and release containers
 FROM alpine:3.11 as secrets-provider-base
 MAINTAINER CyberArk Software Ltd.
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,13 +32,15 @@ RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 # this layer is used to build the debug binaries
 FROM secrets-provider-builder-base as secrets-provider-builder-debug
 
-# Build Delve
+# Build Delve - debugging tool for Go
 RUN go get github.com/go-delve/delve/cmd/dlv
 
+# Expose port 40000 for debugging
 EXPOSE 40000
 
 COPY . .
 
+# Build debug flavor without compilation optimizations using "all=-N -l"
 RUN go build -a -installsuffix cgo -gcflags="all=-N -l" -o secrets-provider ./cmd/secrets-provider
 
 # =================== BUSYBOX LAYER ===================
@@ -98,6 +100,8 @@ COPY --from=secrets-provider-builder-debug /go/bin/dlv /usr/local/bin/
 
 COPY --from=secrets-provider-builder-debug /opt/secrets-provider/secrets-provider /usr/local/bin/
 
+# Execute secrets provider wrapped with dlv debugger listening on port 40000 for remote debugger connection.
+# Will wait indefinitely until a debugger is connected.
 CMD ["/usr/local/bin/dlv", "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/usr/local/bin/secrets-provider"]
 
 # =================== MAIN CONTAINER (REDHAT) ===================
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update
 RUN go get -u github.com/jstemmer/go-junit-report && \
     go get github.com/smartystreets/goconvey
 
-WORKDIR /opt/secrets-provider-for-k8s
+WORKDIR /opt/secrets-provider
 
 EXPOSE 8080
 
@@ -18,6 +18,27 @@ COPY go.mod .
 COPY go.sum .
 
 RUN go mod download
+
+# =================== RELEASE BUILD LAYER ===================
+# this layer is used to build the release binaries
+FROM secrets-provider-builder-base as secrets-provider-builder
+
+COPY . .
+
+RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
+
+# =================== DEBUG BUILD LAYER ===================
+# this layer is used to build the debug binaries
+FROM secrets-provider-builder-base as secrets-provider-builder-debug
+
+# Build Delve
+RUN go get github.com/go-delve/delve/cmd/dlv
+
+EXPOSE 40000
+
+COPY . .
+
+RUN go build -a -installsuffix cgo -gcflags="all=-N -l" -o secrets-provider ./cmd/secrets-provider
 
 # =================== BUSYBOX LAYER ===================
 # this layer is used to get binaries into the main container
@@ -61,50 +82,29 @@ RUN apk add -u shadow libc6-compat && \
 
 USER secrets-provider
 
-# =================== DEBUG BUILD LAYER ===================
-# this layer is used to build the debug binaries
-FROM secrets-provider-builder-base as secrets-provider-builder-debug
+# =================== RELEASE MAIN CONTAINER ===================
+FROM secrets-provider-base as secrets-provider
 
-# Build Delve
-RUN go get github.com/go-delve/delve/cmd/dlv
+COPY --from=secrets-provider-builder /opt/secrets-provider/secrets-provider /usr/local/bin/
 
-EXPOSE 40000
-
-COPY . .
-
-RUN go build -a -installsuffix cgo -gcflags="all=-N -l" -o secrets-provider ./cmd/secrets-provider
+CMD [ "/usr/local/bin/secrets-provider"]
 
 # =================== DEBUG MAIN CONTAINER ===================
 FROM secrets-provider-base as secrets-provider-debug
 
 COPY --from=secrets-provider-builder-debug /go/bin/dlv /usr/local/bin/
 
-COPY --from=secrets-provider-builder-debug /opt/secrets-provider-for-k8s/secrets-provider /usr/local/bin/
+COPY --from=secrets-provider-builder-debug /opt/secrets-provider/secrets-provider /usr/local/bin/
 
 CMD ["/usr/local/bin/dlv", "--listen=:40000", "--headless=true", "--api-version=2", "--accept-multiclient", "exec", "/usr/local/bin/secrets-provider"]
 
-# =================== RELEASE BUILD LAYER ===================
-# this layer is used to build the release binaries
-FROM secrets-provider-builder-base as secrets-provider-builder
-
-COPY . .
-
-RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
-
-# =================== RELEASE MAIN CONTAINER ===================
-FROM secrets-provider-base as secrets-provider
-
-COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /usr/local/bin/
-
-CMD [ "/usr/local/bin/secrets-provider"]
-
 # =================== MAIN CONTAINER (REDHAT) ===================
-FROM registry.access.redhat.com/rhel as secrets-provider-for-k8s-redhat
+FROM registry.access.redhat.com/rhel as secrets-provider-redhat
 MAINTAINER CyberArk Software Ltd.
 
 ARG VERSION
 
-LABEL name="secrets-provider-for-k8s"
+LABEL name="secrets-provider"
 LABEL vendor="CyberArk"
 LABEL version="$VERSION"
 LABEL release="$VERSION"
@@ -132,7 +132,7 @@ RUN groupadd -r secrets-provider \
     chmod 770 /etc/conjur/ssl \
               /run/conjur
 
-COPY --from=secrets-provider-builder /opt/secrets-provider-for-k8s/secrets-provider /usr/local/bin/
+COPY --from=secrets-provider-builder /opt/secrets-provider/secrets-provider /usr/local/bin/
 
 COPY LICENSE.md /licenses
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,30 @@
-FROM golang:1.13 as secrets-provider-builder
+FROM golang:1.13 as secrets-provider-builder-base
 MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux \
     GOARCH=amd64 \
     CGO_ENABLED=0
 
-COPY . /opt/secrets-provider-for-k8s
+RUN apt-get update
+
+RUN go get -u github.com/jstemmer/go-junit-report && \
+    go get github.com/smartystreets/goconvey
 
 WORKDIR /opt/secrets-provider-for-k8s
 
 EXPOSE 8080
 
-RUN apt-get update && apt-get install -y jq
+COPY go.mod .
+COPY go.sum .
 
-RUN go get -u github.com/jstemmer/go-junit-report && \
-    go get github.com/smartystreets/goconvey
-
-RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
+RUN go mod download
 
 # =================== BUSYBOX LAYER ===================
 # this layer is used to get binaries into the main container
 FROM busybox
 
-# =================== MAIN CONTAINER ===================
-FROM alpine:3.11 as secrets-provider
+# =================== BASE MAIN CONTAINER ===================
+FROM alpine:3.11 as secrets-provider-base
 MAINTAINER CyberArk Software Ltd.
 
 # copy a few commands from busybox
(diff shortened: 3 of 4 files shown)


.run/Debug Secrets Provider.run.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Debug Secrets Provider" type="GoRemoteDebugConfigurationType" factoryName="Go Remote" port="43210">
+    <method v="2" />
+  </configuration>
+</component>
```
orenbm, line 5: can you please add a newline?

.run/build debug.run.xml
```diff
@@ -0,0 +1,13 @@
...
+    <option name="SCRIPT_PATH" value="$PROJECT_DIR$/bin/build" />
+    <option name="SCRIPT_OPTIONS" value="debug" />
+    <option name="INDEPENDENT_SCRIPT_WORKING_DIRECTORY" value="true" />
+    <option name="SCRIPT_WORKING_DIRECTORY" value="$PROJECT_DIR$" />
+    <option name="INDEPENDENT_INTERPRETER_PATH" value="true" />
+    <option name="INTERPRETER_PATH" value="/bin/sh" />
+    <option name="INTERPRETER_OPTIONS" value="" />
+    <method v="2" />
+  </configuration>
+</component>
```
orenbm, line 13: can you please add a newline?

Dockerfile
```diff
@@ -1,29 +1,30 @@
-FROM golang:1.13 as secrets-provider-builder
+FROM golang:1.13 as secrets-provider-builder-base
 MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux \
     GOARCH=amd64 \
     CGO_ENABLED=0
 
-COPY . /opt/secrets-provider-for-k8s
+RUN apt-get update
```
orenbm, line 8: what's the purpose of this update?

bin/build
```diff
@@ -1,13 +1,29 @@
...
 FULL_VERSION_TAG="$(full_version_tag)"
 
 echo "---"
 
+if [ $DEBUG ]; then
+
+  echo "Building secrets-provider-debug Docker image"
+
+  docker build \
+    --tag "secrets-provider-debug:latest" \
```
orenbm, line 20: let's maintain the name `secrets-provider-for-k8s` so in this case it can be `secrets-provider-for-k8s-debug`. same goes for the target name.

bin/build
```diff
@@ -1,13 +1,29 @@
...
+    --target "secrets-provider-debug" \
+    .
+
+  exit 0
```
orenbm, line 24: can we please not exit? how about putting this and the rest of the logic in an "if-else"?

what i'm afraid is that we will add more logic below that will be missed with this exit. if it's in an "if-else" then it's harder to miss in the future.

Dockerfile
```diff
@@ -60,9 +61,42 @@ RUN apk add -u shadow libc6-compat && \
...
+# =================== DEBUG BUILD LAYER ===================
+# this layer is used to build the debug binaries
+FROM secrets-provider-builder-base as secrets-provider-builder-debug
+
+# Build Delve
+RUN go get github.com/go-delve/delve/cmd/dlv
+
+EXPOSE 40000
+
+COPY . .
```
orenbm, line 73: @abrahamko this Dockerfile is quite hard to follow. can we have all the builders stacked in the top of the file? 

i think a better order will be:
- secrets-provider-builder-base
- secrets-provider-builder
- secrets-provider-builder-debug
- busybox
- secrets-provider-base
- secrets-provider
- secrets-provider-debug
- secrets-provider-for-k8s-redhat

orenbm: also - it's not from your PR but can we have an alignment in the naming and choose between `secrets-provider` and `secrets-provider-for-k8s`? either is fine by me.

.run/Debug Secrets Provider.run.xml
abrahamko: No can do, this file is generated by the GoLand IDE.

.run/build debug.run.xml
abrahamko: No can do, this file is generated by the GoLand IDE.

Dockerfile
```diff
@@ -1,29 +1,30 @@
...
 
-RUN apt-get update && apt-get install -y jq
+COPY go.mod .
+COPY go.sum .
 
-RUN go get -u github.com/jstemmer/go-junit-report && \
-    go get github.com/smartystreets/goconvey
-
-RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
+RUN go mod download
```
orenbm, line 20: why do we need this? can you please add a comment?

Dockerfile
```diff
@@ -1,29 +1,30 @@
...
 
 # =================== BUSYBOX LAYER ===================
 # this layer is used to get binaries into the main container
 FROM busybox
 
-# =================== MAIN CONTAINER ===================
-FROM alpine:3.11 as secrets-provider
+# =================== BASE MAIN CONTAINER ===================
```
orenbm, line 26: can you please add a comment on why we need a base main container and base builder?

Dockerfile
abrahamko: It was there, I just rearranged it and moved it to the top for improved performance.

Dockerfile
```diff
@@ -1,29 +1,30 @@
...
     GOARCH=amd64 \
     CGO_ENABLED=0
 
-COPY . /opt/secrets-provider-for-k8s
+RUN apt-get update
+
+RUN go get -u github.com/jstemmer/go-junit-report && \
+    go get github.com/smartystreets/goconvey
 
 WORKDIR /opt/secrets-provider-for-k8s
```
orenbm, line 13: can you set a workdir to a path that doesn't exist?

Dockerfile
```diff
@@ -1,29 +1,30 @@
...
 
 EXPOSE 8080
 
-RUN apt-get update && apt-get install -y jq
```
orenbm, line 14: is `jq` not required anymore?

Dockerfile
abrahamko: @orenbm I like the order you put here.
Just so you know, there's a reason for the order I did because using `--target` flag for docker build invokes ALL steps until the specified target.
In our case, it means building for debug builds for also for release EVERY TIME.
There is a way to handle it, though, as there's an addition of a BuildKit to docker that executes only the needed steps for the specified target, but it is not used by default.
Would you like me to add this to the script?

Dockerfile
orenbm: then you missed something. it was `RUN apt-get update && apt-get install -y jq`

Dockerfile
orenbm: i think that readability is more important in this case. if the `BuildKit` addition is not to big then go for it. If not, i still think that it's fine to build redundant parts (as long as they don't take too long) to improve the readability of such a large Dockerfile.

orenbm: wdyt?

Dockerfile
abrahamko: Since when it is required for building go?

abrahamko: Also, I don't know why we need the busybox tools in production...

Dockerfile
abrahamko: Yes https://docs.docker.com/engine/reference/builder/#workdir

Dockerfile
abrahamko: Also, we can use `FROM scratch` in the main containers instead of `FROM alpine`

Dockerfile
abrahamko: I added the buildkit. 😉

Dockerfile
abrahamko: I can't see any reason why `jq` is needed for building Go.

orenbm: yes, it seems that it is not used. Anyway, in that case let's remove also the `apt-get update`.

Dockerfile
orenbm: let's do it one step at a time. Once we finish this PR we can challenge the base image (`scratch/alpine`) and the `busybox` tools.

They may be a leftover but this PR is already big enough to chew on :)

Dockerfile
orenbm: thanks. can you please change it to `Add a layer of prefetched modules so the modules are already cached in case we rebuild`

Dockerfile
```diff
@@ -1,29 +1,55 @@
...
+RUN go build -a -installsuffix cgo -gcflags="all=-N -l" -o secrets-provider ./cmd/secrets-provider
+
 # =================== BUSYBOX LAYER ===================
 # this layer is used to get binaries into the main container
 FROM busybox
 
-# =================== MAIN CONTAINER ===================
-FROM alpine:3.11 as secrets-provider
+# =================== BASE MAIN CONTAINER ===================
+# this layer is used to prepare a common layer for both debug and release containers
```
orenbm, line 51: can you please add a blank line after this comment. Or maybe move this comment to be above the `BASE MAIN CONTAINER` line.

Dockerfile
```diff
@@ -1,29 +1,55 @@
...
-RUN go get -u github.com/jstemmer/go-junit-report && \
-    go get github.com/smartystreets/goconvey
+# Add a layer of prefetched modules so in case of rebuild the modules are already cached
+RUN go mod download
+
+# =================== RELEASE BUILD LAYER ===================
+# this layer is used to build the release binaries
+FROM secrets-provider-builder-base as secrets-provider-builder
+
+COPY . .
```
orenbm, line 29: looks like this is done for both build layers (release and debug). can it move to the base build layer?

bin/build
```diff
@@ -1,27 +1,45 @@
...
+
 . bin/build_utils
 
 FULL_VERSION_TAG="$(full_version_tag)"
 
 echo "---"
 
-echo "Building secrets-provider-for-k8s Docker image"
+if [ $DEBUG == "debug" ]; then
+
```
orenbm, line 17: please remove this blank line

bin/build
```diff
@@ -1,27 +1,45 @@
...
+  echo "Building secrets-provider-debug Docker image"
+
+  docker build \
+    --tag "secrets-provider-for-k8s-debug:latest" \
+    --target "secrets-provider-debug" \
+    .
+
```
orenbm, line 24: please remove this blank line

bin/build
```diff
@@ -1,27 +1,45 @@
...
+else
+
```
orenbm, line 26: please remove this blank line

bin/build
```diff
@@ -1,27 +1,45 @@
 #!/usr/bin/env bash
 
+DEBUG="$1"
+export DOCKER_BUILDKIT=1
```
orenbm, line 4: i think these should go after `set -e`

bin/build
```diff
@@ -1,27 +1,45 @@
...
+cd $(dirname $0)/..
+
 . bin/build_utils
 
 FULL_VERSION_TAG="$(full_version_tag)"
 
 echo "---"
 
-echo "Building secrets-provider-for-k8s Docker image"
+if [ $DEBUG == "debug" ]; then
```
orenbm, line 16: if i understand correctly, you are planning to run this with `./bin/build debug`. Please see how we handle flags in [`test_integration`](https://github.com/cyberark/secrets-provider-for-k8s/blob/master/bin/test_integration) and maintain that behaviour here as well.

Dockerfile
abrahamko: I prefer not to.
This step copies the entire folder, and I want it to be the last one so previous steps could be cached (e.g. in debug build)

orenbm: ok thanks for clarifying.

Dockerfile
abrahamko: We don't do it for any other layer comment. Why should I do it here?

orenbm: you are correct, i didn't notice that was the pattern.

orenbm:
@abrahamko do you think that a section in the README about this change can help? What are the steps that a user should do in order to debug the project?

abrahamko:
@orenbm I agree this should be documented in README, but to be fully useful, I need the manifests that @sigalsax is adding in her PR for local deployment.
So it should be added to README once the local dev deployment also works.

.run/build debug.run.xml
```diff
@@ -0,0 +1,13 @@
...
+    <option name="SCRIPT_PATH" value="$PROJECT_DIR$/bin/build" />
+    <option name="SCRIPT_OPTIONS" value="--debug" />
+    <option name="INDEPENDENT_SCRIPT_WORKING_DIRECTORY" value="true" />
+    <option name="SCRIPT_WORKING_DIRECTORY" value="$PROJECT_DIR$" />
+    <option name="INDEPENDENT_INTERPRETER_PATH" value="true" />
+    <option name="INTERPRETER_PATH" value="/bin/sh" />
+    <option name="INTERPRETER_OPTIONS" value="" />
+    <method v="2" />
+  </configuration>
+</component>
```
sigalsax, line 13: New line

bin/build
```diff
@@ -2,26 +2,44 @@
...
-   .
+if [ $DEBUG = true ]; then
+  echo "Building secrets-provider-debug Docker image"
+
+  docker build \
+    --tag "secrets-provider-for-k8s-debug:latest" \
+    --target "secrets-provider-debug" \
+    .
+else
+  echo "Building secrets-provider-for-k8s Docker image"
```
sigalsax, line 28: Let's add back in the version
echo "Building secrets-provider-for-k8s-redhat:$FULL_VERSION_TAG Docker image"

bin/build
```diff
@@ -2,26 +2,44 @@
...
 
-echo "Building secrets-provider-for-k8s-redhat:$FULL_VERSION_TAG Docker image"
+echo "---"
 
-docker build --tag "secrets-provider-for-k8s-redhat:${FULL_VERSION_TAG}" \
-   --build-arg VERSION="$FULL_VERSION_TAG" \
-   --target "secrets-provider-for-k8s-redhat" \
-   .
+if [ $DEBUG = true ]; then
+  echo "Building secrets-provider-debug Docker image"
```
sigalsax, line 21: I understand for backwards compatibility but not sure we need to debug the red hat image because there are no real changes to the code to make it "rh-certified"

sigalsax:
There are lots of moving pieces in this PR and not a lot of explanation. I would rework your commits, add comments in actual files, etc to explain what you did and why you did them. For example, why are we introducing Delve, why are we exposing 4000. 

Without some background knowledge it is more challenging to give a thorough review

These details can be packed in commits or as comments. Take a look at some of @orenbm commits

Dockerfile
```diff
@@ -1,29 +1,53 @@
...
+FROM secrets-provider-builder-base as secrets-provider-builder-debug
+
+# Build Delve
+RUN go get github.com/go-delve/delve/cmd/dlv
+
+EXPOSE 40000
+
+COPY . .
+
+RUN go build -a -installsuffix cgo -gcflags="all=-N -l" -o secrets-provider ./cmd/secrets-provider
```
sigalsax, line 42: what is `"all=-N -l"`?

Dockerfile
```diff
@@ -60,17 +84,29 @@ RUN apk add -u shadow libc6-compat && \
...
-ENTRYPOINT [ "/usr/local/bin/secrets-provider"]
+COPY --from=secrets-provider-builder /opt/secrets-provider/secrets-provider /usr/local/bin/
+
+CMD [ "/usr/local/bin/secrets-provider"]
+
+# =================== DEBUG MAIN CONTAINER ===================
+FROM secrets-provider-base as secrets-provider-debug
+
+COPY --from=secrets-provider-builder-debug /go/bin/dlv /usr/local/bin/
+
```
sigalsax, line 98: you can group the COPYs together IMO

bin/build
```diff
@@ -2,26 +2,44 @@
 
 set -e
 
-. bin/build_utils
+export DOCKER_BUILDKIT=1
```
sigalsax, line 5: do we need this? what is it for?

abrahamko: I added a comment to explain it

Dockerfile
abrahamko: We don't do it for any other consequent COPY in the file.

orenbm:
> @orenbm I agree this should be documented in README, but to be fully useful, I need the manifests that @sigalsax is adding in her PR for local deployment.
> So it should be added to README once the local dev deployment also works.

No problem, please create an issue for you to update the README once all the effort is done.

Dockerfile
abrahamko: I added a comment to explain it

Dockerfile
orenbm: @abrahamko can you create an issue for challenging the base image (scratch/alpine) and the busybox tools?

Dockerfile
```diff
@@ -60,17 +86,31 @@ RUN apk add -u shadow libc6-compat && \
...
 
 # =================== MAIN CONTAINER (REDHAT) ===================
-FROM registry.access.redhat.com/rhel as secrets-provider-for-k8s-redhat
+FROM registry.access.redhat.com/rhel as secrets-provider-redhat
 MAINTAINER CyberArk Software Ltd.
 
 ARG VERSION
 
-LABEL name="secrets-provider-for-k8s"
+LABEL name="secrets-provider"
```
orenbm, line 113: i wouldn't change the label as it is a breaking change. I am not sure but the label may be present on the RH site and customers may be looking for it.

i see that you chose to go with `secrets-provider` instead of `secrets-provider-for-k8s` throughout the file. maybe the latter is better.

bin/build
```diff
@@ -2,26 +2,47 @@
...
+DEBUG=false
 
-FULL_VERSION_TAG="$(full_version_tag)"
+if [ "$1" == "--debug" ]; then
+    DEBUG=true
+fi
 
-echo "---"
+# go to repo root folder for execution
+cd $(dirname $0)/..
```
orenbm, line 15: how come we need this now whereas before we didn't?

abrahamko: Who said we didn't?
The script cannot be run from any other location besides from the root folder using `./bin/build`.
I wanted to allow it to run from any location; why it should care where is it called from?

bin/build
```diff
@@ -2,26 +2,47 @@
...
 set -e
 
-. bin/build_utils
+# Use Docker BuildKit to efficiently build the targets.
+# With BuildKit, only relevant layers for the target are built.
+export DOCKER_BUILDKIT=1
+DEBUG=false
 
-FULL_VERSION_TAG="$(full_version_tag)"
+if [ "$1" == "--debug" ]; then
```
orenbm, line 10: I really prefer that we handle this in a way that it is easy to add more params in the future. Please see the way that we handled this in `./bin/test_integration`. Do you think it's not the correct way to do it?

Dockerfile
abrahamko: I'm against this change, this was not my decision to call it like that, and all the file is full of `secrets-provider`.
I'll revert it.

orenbm: i think it is also better to use the full name anyway. sorry for not saying so earlier.

bin/build
orenbm: ah i see. I don't see the need to run it from anywhere else than the root folder but if you think there is such a need then i'm ok with it if this is the only change. 

From which location did you need to run it other than the root folder?

Dockerfile
orenbm: again, sorry i didn't decide earlier on my preffered name but I do think that the full name is better. So will you change all the occurrences of `secrets-provider` to `secrets-provider-for-k8s` throughout the code? I think it is better to have uniformity.

bin/build
abrahamko: I tried from `bin` folder and failed.
Short investigation brought this solution, which allows running from any location.

Dockerfile
abrahamko: I added it to the dockerfile improvements issue I opened: #150

Dockerfile
```diff
@@ -32,13 +32,15 @@ RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 # this layer is used to build the debug binaries
 FROM secrets-provider-builder-base as secrets-provider-builder-debug
 
-# Build Delve
+# Build Delve - debugging tool for Go
```
sigalsax, line 35: <3

Dockerfile
```diff
@@ -32,13 +32,15 @@ RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
...
 RUN go get github.com/go-delve/delve/cmd/dlv
 
+# Expose port 40000 for debugging
```
sigalsax, line 38: bingo

bin/build
```diff
@@ -2,26 +2,50 @@
...
+    --tag "secrets-provider-for-k8s:latest" \
+    --target "secrets-provider" \
+    .
+
+  echo "Building secrets-provider-for-k8s-redhat:$FULL_VERSION_TAG Docker image"
+
+  docker build --tag "secrets-provider-for-k8s-redhat:${FULL_VERSION_TAG}" \
+     --build-arg VERSION="$FULL_VERSION_TAG" \
+     --target "secrets-provider-for-k8s-redhat" \
+     .
```
sgnn7, line 48: While in general use of Dockerfile with the `target` is very obscure use, I can see why this approach is done here. For me though, this seems to be overall too much bundling of too many things into a single file, making it hard to read and making it even harder to maintain in the future. I think a better design might be
- A split-out of individual pieces:
```
- Dockerfile
- Dockerfile.debug (used the `FROM` linking to product tag
- Dockerfile.redhat
```
- Clear split of responsibilities within the `Dockerfile`:
```
<prerequisites>
<main container build out - single block>
<redhat container build out - single block>
<debug container build out - single block>
```

Dockerfile
```diff
@@ -1,29 +1,55 @@
...
+# this layer is used to prepare a common layer for both debug and release builds
+FROM golang:1.13 as secrets-provider-builder-base
 MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux \
     GOARCH=amd64 \
     CGO_ENABLED=0
 
-COPY . /opt/secrets-provider-for-k8s
+RUN go get -u github.com/jstemmer/go-junit-report && \
```
sgnn7, line 10: This seems like it should be installed during test runs and not included in the container we ship out

Dockerfile
```diff
@@ -1,29 +1,55 @@
...
+# this layer is used to build the release binaries
+FROM secrets-provider-builder-base as secrets-provider-builder
+
+COPY . .
 
 RUN go build -a -installsuffix cgo -o secrets-provider ./cmd/secrets-provider
 
+# =================== DEBUG BUILD LAYER ===================
+# this layer is used to build the debug binaries
+FROM secrets-provider-builder-base as secrets-provider-builder-debug
```
sgnn7, line 33: If the only thing you're doing with this base layer is to get some files, you might as well collapse this layer and `DEBUG MAIN CONTAINER` together

bin/build
abrahamko: Added it as a consideration for a following issue #150

Dockerfile
abrahamko: Added it as a consideration for a following issue #150

Dockerfile
abrahamko: No can do, the base image for building and running is not the same.

Issue #102, opened 2020-05-31
Secrets Provider support Rotations and deployed separately - Research


InbalZilberman:
### Overview
We would like to improve the secret provider scope and support
Milestone 1
- Secrets Provider runs and deployed without affecting the app deployment or app life cycle
- Secrets Provider can serve many apps
Milestone 2
- Secrets Provider support rotations 

#### Milestone 1
**As a** operation team, Liz
**I would like to** deploy Secret provider in a way that do not interfere with my application deployment
**So That** my app can be upgraded and managed in separation from secret provider

**AS an** k8s app developer, David
**I would like to** fetch my secret in an easy and naive way without changing my code
**So That** my app will get the most updated secret as a K8s secret (usually provided in the deployment as an env var or file).
The secrets are expected to be provided in advance before the app is initialised.

If the secrets change value please look at :How to handle rotation? section below

#### Milestone 2
**AS an** k8s app developer, David
**I would like to** fetch my secret in an easy and naive way without changing my code
**So That** my app will get the most updated secret as a K8s secret (usually provided in the deployment as an env var or file).
The secrets are expected to be provided in advance before the app is initialized.

**If the secrets change value please look at :How to handle rotation? section below**

**As an** info sec can be Vault admin or Conjur admin or DAP admin
**I would like to** rotate the secret value in Vault or in DAP manually or using CPM policies
**So That** I'll stay secure and follow security best practices

**As an** info sec can be Vault admin or Conjur admin or DAP admin
**I would like to** change the secret value 
**So That** I'll stay secure and follow security best practices

### Requirements for  Milestone 1
In this phase Secrets provider needs to be able to run as a separate entity and serve multiple application containers that runs on multiple pods.

### Deployment requirements 

- Secrets provider needs to be able to run as a separate entity and serve multiple application containers that runs on multiple pods.
- The lifecycle of this deployment option should support also upgrade and removal of deployment
- Provide a way for our customer to understand the state of the secret provider - when it finished initializing 
- The deployment method of the secret provider should be native to K8s. Deployment and uninstall should be done using **helm**

**Deployment Flow with helm chart**
Prerequisite
 1. Configure K8s Secrets with conjur-map metadata defined 
     Optional: add label to K8s Secret (Milestone 2)

2. Create custom-values.yaml file that will optionally contain the following params:

 2.1 Service Account Name (default: secret-provider-account)
 2.2 Role Name(default: secret-provider-role)
[...]

PR #151, opened 2020-07-13
Add solution doc for Milestone 1

sigalsax:
This PR adds the solution doc for the Secrets Provider for K8s (Phase 2). This documentation covers Milestone 1 and references future efforts for Milestone 2

Readable version can be found [here](https://github.com/cyberark/secrets-provider-for-k8s/blob/job-deployment-design/design/milestone_1_2_design_doc.md)

### DIFF
--- a/design/batch_retrieval_design.md
+++ b/design/batch_retrieval_design.md
@@ -6,17 +6,15 @@ This document will cover the options for enhancing Batch Retrieval requests.
 
 *Previously*, for the init container solution, each Secrets Provider was sitting within the same Deployment as the app. Because of this, when a batch retrieval request was made, the requests were dispersed across many call to the Conjur server. If one K8s Secret was not able to be updated with a Conjur value (for example due to a permission error), then the batch request would fail and only that pod would not spin up successfully. 
 
- *Now* that the Secrets Provider is outside, a single batch retrieval request will be made to Conjur for *all* application. So if there is a failure in receiving one of the values, then the whole request would fail. In other words, **no application** will spin up in the namespace.
+*Now* that the Secrets Provider is outside, a single batch retrieval request will be made to Conjur for *all* application. So if there is a failure in receiving one of the values, then the whole request would fail. In other words, **no application** will spin up in the namespace.
 
-|      | Solution                                                     | Pros                                                         | Cons                                | Effort |
-| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------------------------- | ------ |
-| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*) |        |
-| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) | - Will help us during rotation for Milestone 2<br />- Better / straightforward design for how batch endpoints | May introduce breaking changes (**) |        |
-| 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                              | 0      |
+|      | Solution                                                     | Pros                                                         | Cons                                                         | Effort (T-shirt size) |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | --------------------- |
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                          | S, 3 days             |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Better / straightforward design for how batch endpoints | - Requires both client/server implementation<br />- Need to handle backwards compatibility | M, 7 days             |
+| 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                                                       | -                     |
 
 (*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
 
-(**) Fallback solution: slowly deprecate old batch retrieval API *or* create a new API endpoint
-
-
+(**) This solution can be broken up in two: 1. Create a new API endpoint or 2. use existing one (return 409 response)
 
--- a/design/batch_retrieval_design.md
+++ b/design/batch_retrieval_design.md
@@ -1,0 +1,22 @@
+# Batch retrieval design
+
+This document will cover the options for enhancing Batch Retrieval requests.
+
+### Motivation
+
+*Previously*, for the init container solution, each Secrets Provider was sitting within the same Deployment as the app. Because of this, when a batch retrieval request was made, the requests were dispersed across many call to the Conjur server. If one K8s Secret was not able to be updated with a Conjur value (for example due to a permission error), then the batch request would fail and only that pod would not spin up successfully. 
+
+ *Now* that the Secrets Provider is outside, a single batch retrieval request will be made to Conjur for *all* application. So if there is a failure in receiving one of the values, then the whole request would fail. In other words, **no application** will spin up in the namespace.
+
+|      | Solution                                                     | Pros                                                         | Cons                                | Effort |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------------------------- | ------ |
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*) |        |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) | - Will help us during rotation for Milestone 2<br />- Better / straightforward design for how batch endpoints | May introduce breaking changes (**) |        |
+| 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                              | 0      |
+
+(*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
+
+(**) Fallback solution: slowly deprecate old batch retrieval API *or* create a new API endpoint
+
+
+
--- a/design/batch_retrieval_design.md
+++ b/design/batch_retrieval_design.md
@@ -10,8 +10,8 @@ This document will cover the options for enhancing Batch Retrieval requests.
 
 |      | Solution                                                     | Pros                                                         | Cons                                                         | Effort (T-shirt size) |
 | ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | --------------------- |
-| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                          | S, 3 days             |
-| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Better / straightforward design for how batch endpoints | - Requires both client/server implementation<br />- Need to handle backwards compatibility | M, 7 days             |
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                          | S, 5 days             |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Better / straightforward design for how batch endpoints | - Requires both client/server implementation<br />- Need to handle backwards compatibility | M, 10 days            |
 | 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                                                       | -                     |
 
 (*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
--- a/design/milestone_1_2_design_doc.md
+++ b/design/milestone_1_2_design_doc.md
@@ -1,4 +1,4 @@
-# Secrets Provider Phase 2 Milestone 1 - Solution Design
+# Secrets Provider Phase 2 Milestone Job - Solution Design
 
 ## Table of Contents
 
@@ -8,29 +8,29 @@
   * [Background](#background)
     + [Motivation](#motivation)
   * [Requirements](#requirements)
-    + [Milestone 1 *(current)*](#milestone-1---current--)
-    + [Milestone 2](#milestone-2)
+    + [Milestone Job *(current)*](#milestone-1---current--)
+    + [Milestone Rotation](#milestone-2)
   * [Solution](#solution)
-    + [Milestone 1: Serve K8s secrets to multiple applications once (no rotation)](#milestone-1--serve-k8s-secrets-to-multiple-applications-once--no-rotation-)
+    + [Milestone Job: Serve K8s secrets to multiple applications once (no rotation)](#milestone-1--serve-k8s-secrets-to-multiple-applications-once--no-rotation-)
     + [Design](#design)
       - [How does a Job configured via Helm answer the requirements?](#how-does-a-job-configured-via-helm-answer-the-requirements-)
       - [What drawbacks does this solution have?](#what-drawbacks-does-this-solution-have-)
       - [Customer experience](#customer-experience)
       - [Packaging Helm Charts](#packaging-helm-charts)
-    + [Milestone 2: Rotation](#milestone-2--rotation)
+    + [Milestone Rotation](#milestone-2--rotation)
     + [Lifecycle](#lifecycle)
-      - [Milestone 1](#milestone-1)
+      - [Milestone Job](#milestone-1)
         * [Installation](#installation)
         * [Lifecycle](#lifecycle-1)
         * [Update](#update)
         * [Delete](#delete)
     + [Network Disruptions](#network-disruptions)
     + [Order of Deployment](#order-of-deployment)
-      - [Milestone 1](#milestone-1-1)
+      - [Milestone Job](#milestone-1-1)
     + [Backwards compatibility](#backwards-compatibility)
-      - [Milestone 1](#milestone-1-2)
+      - [Milestone Job](#milestone-1-2)
     + [Performance](#performance)
-      - [Milestone 1](#milestone-1-3)
+      - [Milestone Job](#milestone-1-3)
     + [Affected Components](#affected-components)
   * [Security](#security)
   * [Test Plan](#test-plan)
@@ -70,14 +70,14 @@ The current implementation of the Secrets Provider only supports one application
 
 The solution should support the following capabilities:
 
-###  Milestone 1 *(current)*
+###  Milestone Job *(current)*
 
 - Secrets Provider runs as a separate entity, serving multiple application containers that run on multiple pods
 - Solution needs to be native to Kubernetes
 - Lifecycle should support update/removal of deployment
 - Provide a way for customer to understand the state of the Secret Provider - when it finished initializing
 
-###  Milestone 2
+###  Milestone Rotation
 
 - Secrets Provider support secret rotation
 
@@ -91,9 +91,9 @@ The solution is to allow different deployments and behave differently based on t
 
 These variations will be supported using the same Secrets Provider image, but the behavior will vary dynamically depending on the chosen deployment method that is configured via Helm Chart.
 
-*In this document we will focus on **Milestone 1***, deploying the Secrets Provider as a Job using Helm
+*In this document we will focus on **Milestone Job***, deploying the Secrets Provider as a Job using Helm
 
-### Milestone 1: Serve K8s secrets to multiple applications once (no rotation)
+### Milestone Job: Serve K8s secrets to multiple applications once (no rotation)
 
 ### Design
 
@@ -121,7 +121,7 @@ To handle the missing **Job** granularity there are 2 options:
 | 1    | Keep using only **namespace** and/or **service account** granularities. | \- Since Secrets Provider runs as a separate entity, we suggest creating a dedicated service account for Secrets Provider.<br />That way, the **service account** granularity serves as an exact match and **Job** granularity is redundant<br />- No code changes | \- Service account might be reused, which allows other apps to authenticate using Secrets Provider host<br />- Break application identity convention | Free              |
 | 2    | Add support for **Job** granularity in Conjur                | Allows specific identification of the Secrets Provider **Job**, which enhances security | \- Costly; requires code changes, verify compatibility, tests, docs<br />- Redundant if **service account** serves only the Secrets Provider<br />- ***Not*** a ***seamless*** experience between Milestones | 10 days           |
 
-Implementing the Job granularity would mean that the experience in transitioning between the Milestones is not a smooth one. If a customer uses the Job granularity, they will need to update this in Milestone 2 because the deployment type will no longer be a Job. 
+Implementing the Job granularity would mean that the experience in transitioning between the Milestones is not a smooth one. If a customer uses the Job granularity, they will need to update this in Milestone Rotation because the deployment type will no longer be a Job. 
 
 *Decision*: Solution #1, use **namespace** and **service account** to define the Secrets Provider host in Conjur policy. We would recommend to the customer to create a dedicated service account for Secrets Provider.
 
@@ -363,9 +363,9 @@ In the case where the customer has an older Conjur version and the new Conjur AP
 
 For a detailed breakdown of the decision process see [here](batch_retrieval_design.md).
 
-### Milestone 2: Rotation
+### Milestone Rotation
 
-The following section will be dedicated to a high-level design discussion on rotation for Milestone 2.
+The following section will be dedicated to a high-level design discussion on rotation for Milestone Rotation.
 
 The following high-level solutions for rotation were evaluated:
 
@@ -383,7 +383,7 @@ Low level design of the solution will be introduced in a separate document.
 This solution introduces an additional requirement for the customer. They must upgrade their Conjur server to access the new API.
 This is not mandatory and if not done so, we will provide a fallback solution **(1.a)** but at a price of efficiency.
 
-For this Milestone, the Secrets Provider Chart will be using `Deployment` in-place of `Job`. New default variables may be introduced but no additional steps will needed to migrate from Milestone 1 to 2.
+For this Milestone, the Secrets Provider Chart will be using `Deployment` in-place of `Job`. New default variables may be introduced but no additional steps will needed to migrate from Milestone Job to Rotation.
 
 #### Concerns
 
@@ -393,7 +393,7 @@ To assuage this concern we will test performance and supply our customers with c
 
 ### Lifecycle
 
-#### Milestone 1
+#### Milestone Job
 
 ##### Installation
 
@@ -425,7 +425,7 @@ Our customers may experience disruptions in their network connection so it is im
 
 ### Order of Deployment
 
-#### Milestone 1
+#### Milestone Job
 
 Steps to follow for successful deployment
 
@@ -437,13 +437,13 @@ Steps to follow for successful deployment
 
 ### Backwards compatibility
 
-#### Milestone 1
+#### Milestone Job
 
 Introducing Helm will enhance the user experience and not impact the previous init container solution.
 
 ### Performance
 
-#### Milestone 1
+#### Milestone Job
 
 We will test and document how many K8s Secrets can be updated in 5 minutes on average. A secret should be either extremely long password. See [Performance tests](#performance-tests) for further explanation.
 
@@ -465,7 +465,7 @@ The security boundary of the Secrets Provider is the Host identity it uses to au
 
 #### Controls
 
-We accept user input via the *`stringData`* in the K8s Secret resource. Kubernetes mitigates vulnerabilities in the Secret resource by encoding the string when the Secret is created or updated.
+We accept user input via the *`stringData`* in the K8s Secret resource. We will need to escape/encode this information before using them in the backend. 
 
 The values placed in `custom-values.yaml` are determined by the user. To guarantee that the parameters we accept have not been manipulated for malicious purposes, we will supply a JSON schema to impose an expected structure on input.
 
@@ -476,7 +476,7 @@ The values placed in `custom-values.yaml` are determined by the user. To guarant
 | **Title** | Description                                                  | Given                                                        | When                                          | Then                                                         |
 | --------- | ------------------------------------------------------------ | ------------------------------------------------------------ | --------------------------------------------- | ------------------------------------------------------------ |
 | 1         | *Vanilla flow*, Secret Provider Job successfully updates K8s Secrets | - Conjur is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and K8s Secrets are configured<br />- All mandatory values are defined in `custom-values.yaml`<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job                | - Secrets Provider pod authenticates and fetches Conjur secrets successfully<br />- All K8s Secrets are updated with  Conjur value <br />- App pods receive K8s Secret with Conjur secret values as environment variables<br />- Verify in logs that deployment kind is listed <br />- Secrets Provider Job runs to completion<br />- Verify logs |
-| 2         | Validate new batch retrieval functionality *(for Milestone 1.5)* | Secrets Provider host does not have permissions on Conjur secrets / secret is missing / secret value doesn't exist | Secrets Provider runs as a Job                | - Verify in logs that a list of  failed Conjur secrets and those K8s Secrets that were skipped are noted |
+| 2         | Validate new batch retrieval functionality *(for Milestone Batch)* | Secrets Provider host does not have permissions on Conjur secrets / secret is missing / secret value doesn't exist | Secrets Provider runs as a Job                | - Verify in logs that a list of  failed Conjur secrets and those K8s Secrets that were skipped are noted |
 | 3         | *Vanilla flow*, non-conflicting Secrets Provider             | Two Secrets Providers have access to different Conjur secret | 2 Secrets Provider Jobs run in same namespace | - All relevant K8s secrets are updated<br />- Verify logs    |
 | 4         | Multiple Secrets Providers in same namespace have access to same K8s Secret | Two Secrets Providers have access to same secret             | 2 Secrets Provider Jobs run in same namespace | - No error will be returned and both Secrets Providers will be able to access and update K8s Secret <br />- Verify logs |
 | 5         | Check the integrity of all user input from `values.yaml` / `custom-values.yaml` |                                                              |                                               | - Verify all input values are escaped/encoded                |
@@ -514,15 +514,15 @@ From these performance tests we will be able to detail our limitations in our do
 
 ## Logs
 
-| **Scenario**                                         | **Log message**                                              | Log level |
-| ---------------------------------------------------- | ------------------------------------------------------------ | --------- |
-| Secrets Provider spins up                            | Kubernetes Secrets Provider v\*%s\* starting up as a %s... (Job/Deployment) | Info      |
-| Secrets Provider batch retrieval has partial success | Failed to retrieve Conjur secrets '%s'                       | Error     |
-| Secrets Provider batch retrieval has partial success | Skipped on K8s Secrets '%s'. Reason: failed to retrieve their Conjur secrets | Debug     |
-| Secrets Provider success                             | Successfully updated '%d' out of '%d' K8s Secrets            | Info      |
-| Old batch retrieval endpoint *(for Milestone 1.5)*   | Warning: Secrets Provider cannot efficiently run because Conjur server is not up-to-date. Please consider upgrading to '%d' | Warn      |
-| Job has completed and is terminating                 | Kubernetes Secrets Provider has finished task and is terminating… | Info      |
-| Acknowledge when a retry is taking place             | Retrying '%d' out of '%d' ...                                | Info      |
+| **Scenario**                                                 | **Log message**                                              | Log level |
+| ------------------------------------------------------------ | ------------------------------------------------------------ | --------- |
+| Secrets Provider spins up                                    | Kubernetes Secrets Provider v\*%s\* starting up as a %s... (Job/Deployment) | Info      |
+| Secrets Provider batch retrieval has partial success         | Failed to retrieve Conjur secrets '%s'                       | Error     |
+| Secrets Provider batch retrieval has partial success *(for Milestone Batch)* | Skipped on K8s Secrets '%s'. Reason: failed to retrieve their Conjur secrets | Debug     |
+| Secrets Provider success *(for Milestone Batch)*             | Successfully updated '%d' out of '%d' K8s Secrets            | Info      |
+| Old batch retrieval endpoint *(for Milestone Batch)*         | Warning: Secrets Provider cannot efficiently run because Conjur server is not up-to-date. Please consider upgrading to '%d' | Warn      |
+| Job has completed and is terminating                         | Kubernetes Secrets Provider Job has completed…               | Info      |
+| Acknowledge when a retry is taking place                     | Retrying '%d' out of '%d' ...                                | Info      |
 
 ### Audit 
 
@@ -544,36 +544,41 @@ All fetches on a Conjur Resource are individually audited, creating its own audi
 
 ### Delivery Plan 
 
-#### Milestone 1
+#### Milestone Job
 
-- [ ] Solution design approval + Security review approval
-- [ ] Implement Phase 2 Milestone 1 functionality 
+- [x] Solution design approval + Security review approval
+- [ ] Implement Phase 2 Milestone Job functionality 
   - [ ] Build Helm charts ***(3 days)***
 - [ ] Implement test plan
   - [ ] Integration ***(3 days)***
-  - [ ] Performance tests align with SLA ***(3 days)***
-- [ ] Add support for OC 4.3 in our pipeline ***(not final, 2 days)***
-- [ ] Security items have been taken care of ***(2 days)***
+- [ ] Security items have been taken care of
+  - [ ] StringData encoding/escaping
+  - [ ] JSON schema for ` values`/`custom-values.yaml` ***(2 days)***
 - [ ] Logs review by TW + PO ***(1 day)***
 - [ ] Documentation has been given to TW + approved ***(2 days)***
 - [ ] Engineer(s) not involved in project use documentation to get end-to-end ***(1 day)***
-- [ ] Create demo for Milestone 1 functionality ***(1 day)***
+- [ ] Create demo for Milestone Job functionality ***(1 day)***
 - [ ] Versions are bumped in all relevant projects (if necessary) and automate the Helm package and release process ***(2 days)***
 
- **Total:** ~20 days **(~4 weeks)**
+ **Total:** ~15 days **(~3 weeks)**
 
-#### Milestone 1.5
+Note that the above estimation does not include OC 4.3 testing and pipeline additions
 
-- [ ] Batch Retrieval, create new Conjur API endpoint ***(10 days)***
+#### Milestone Batch
+
+- [ ] Batch Retrieval, create new Conjur API endpoint 
+  - [ ] Work in Conjur ***(10 days)***
+  - [ ] Work in Secrets Provider for K8s ***(3 days)***
 - [ ] Implement test plan
   - [ ] Integration tests in Secrets Provider ***(2 days)***
   - [ ] Integration tests in Conjur ***(3 days)***
-  - [ ] Performance tests align with SLA ***(2 days)***
-- [ ] Update documentation, detailing new Conjur Batch Retrieval endpoint ***(2 days)***
+  - [ ] Performance tests align with SLA ***(3 days)***
+- [ ] Update documentation, updating Batch retrieval limitation ***(1 days)***
 - [ ] Engineer(s) not involved in project use documentation to get end-to-end ***(1 day)***
+- [ ] Create demo for Milestone Job functionality ***(1 day)***
 - [ ] Versions are bumped in all relevant projects (if necessary) ***(1 day)***
 
-**Total:** ~21 days **(~4 weeks)**
+**Total:** ~25 days **(~5 weeks)**
 
 *Risks that could delay project completion*
 
--- a/design/milestone_1_2_design_doc.md
+++ b/design/milestone_1_2_design_doc.md
@@ -178,7 +178,7 @@ secrets-provider-environment:
   conjurAccount:
   conjurApplianceUrl:
   conjurAuthnUrl:
-  conjurAuthnLogin:
+  conjurAuthnLogin: 
   conjurSslCertificateValue:
   k8sSecrets:		# Format needs to be array
 ```
@@ -465,7 +465,7 @@ The security boundary of the Secrets Provider is the Host identity it uses to au
 
 #### Controls
 
-The value for *`stringData`* in the K8s Secret resource is a String of user input values. To guarantee that this field is not manipulated for malicious purposes, we are validating this input.
+We accept user input via the *`stringData`* in the K8s Secret resource. Kubernetes mitigates vulnerabilities in the Secret resource by encoding the string when the Secret is created or updated.
 
 The values placed in `custom-values.yaml` are determined by the user. To guarantee that the parameters we accept have not been manipulated for malicious purposes, we will supply a JSON schema to impose an expected structure on input.
 
@@ -476,16 +476,15 @@ The values placed in `custom-values.yaml` are determined by the user. To guarant
 | **Title** | Description                                                  | Given                                                        | When                                          | Then                                                         |
 | --------- | ------------------------------------------------------------ | ------------------------------------------------------------ | --------------------------------------------- | ------------------------------------------------------------ |
 | 1         | *Vanilla flow*, Secret Provider Job successfully updates K8s Secrets | - Conjur is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and K8s Secrets are configured<br />- All mandatory values are defined in `custom-values.yaml`<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job                | - Secrets Provider pod authenticates and fetches Conjur secrets successfully<br />- All K8s Secrets are updated with  Conjur value <br />- App pods receive K8s Secret with Conjur secret values as environment variables<br />- Verify in logs that deployment kind is listed <br />- Secrets Provider Job runs to completion<br />- Verify logs |
-| 2         | Validate new batch retrieval functionality                   | Secrets Provider host does not have permissions on Conjur secrets / secret is missing / secret value doesn't exist | Secrets Provider runs as a Job                | - Verify in logs that a list of  failed Conjur secrets and those K8s Secrets that were skipped are noted |
+| 2         | Validate new batch retrieval functionality *(for Milestone 1.5)* | Secrets Provider host does not have permissions on Conjur secrets / secret is missing / secret value doesn't exist | Secrets Provider runs as a Job                | - Verify in logs that a list of  failed Conjur secrets and those K8s Secrets that were skipped are noted |
 | 3         | *Vanilla flow*, non-conflicting Secrets Provider             | Two Secrets Providers have access to different Conjur secret | 2 Secrets Provider Jobs run in same namespace | - All relevant K8s secrets are updated<br />- Verify logs    |
 | 4         | Multiple Secrets Providers in same namespace have access to same K8s Secret | Two Secrets Providers have access to same secret             | 2 Secrets Provider Jobs run in same namespace | - No error will be returned and both Secrets Providers will be able to access and update K8s Secret <br />- Verify logs |
-| 5         | Check the integrity of all user input from `values.yaml` / `custom-values.yaml` |                                                              |                                               | - Verify all input values are escaped/encode d               |
-|           | Service Account does not exist                               | `provideExistingServiceAccount=true` but Service Account provided does not exist in namespace | Secrets Provider Chart is installed           | Job will fail because Service Account does not exist         |
+| 5         | Check the integrity of all user input from `values.yaml` / `custom-values.yaml` |                                                              |                                               | - Verify all input values are escaped/encoded                |
+| 6         | Service Account does not exist                               | `provideExistingServiceAccount=true` but Service Account provided does not exist in namespace | Secrets Provider Chart is installed           | Job will fail because Service Account does not exist         |
 | 7         | Add retry check in all regression tests                      |                                                              |                                               | - Verify retry is also written in logs                       |
 | 8         | *Regression tests*                                           | All regression tests should pass (both Conjur / DAP)         |                                               |                                                              |
-| 9         | *Add support for OC 4.3*                                     | All integration tests should pass on OC 4.3                  |                                               | *Not final*                                                  |
-
-Note we will also write tests in Conjur for the new batch retrieval endpoint.
+| 9         | *Performance tests* according to SLA                         |                                                              |                                               |                                                              |
+| 10        | *Add support for OC 4.3*                                     | All integration tests should pass on OC 4.3                  |                                               | *Not final*                                                  |
 
 ### Performance tests
 
@@ -511,19 +510,19 @@ Since this number may vary depending on many variables, we will do this test for
 | Follower distance                           | None (Same K8S cluster)                   | None, Medium, Far           |
 | Number of Followers behind L4 Load Balancer | 1                                         | 1, 2, 3, 5, 10              |
 
-From these performance tests we will be able to detail our limitations in our documentation and provide recommendated deployments.
+From these performance tests we will be able to detail our limitations in our documentation and provide recommended deployments.
 
 ## Logs
 
-| **Scenario**                                         | **Log message**                                              |
-| ---------------------------------------------------- | ------------------------------------------------------------ |
-| Secrets Provider spins up                            | Kubernetes Secrets Provider v\*%s\* starting up as a %s... (Job/Deployment) |
-| Secrets Provider batch retrieval has partial success | Failed to retrieve Conjur secrets '%s'                       |
-| Secrets Provider batch retrieval has partial success | Skipped on K8s Secrets '%s'. Reason: failed to retrieve their Conjur secrets |
-| Secrets Provider success                             | Successfully updated '%d' out of '%d' K8s Secrets            |
-| Old batch retrieval endpoint                         | Warning: Secrets Provider cannot efficiently run because Conjur server is not up-to-date. Please consider upgrading to '%d' |
-| Job has completed and is terminating                 | Kubernetes Secrets Provider has finished task and is terminating… |
-| Acknowledge when a retry is taking place             | Retrying '%d' out of '%d' ...                                |
+| **Scenario**                                         | **Log message**                                              | Log level |
+| ---------------------------------------------------- | ------------------------------------------------------------ | --------- |
+| Secrets Provider spins up                            | Kubernetes Secrets Provider v\*%s\* starting up as a %s... (Job/Deployment) | Info      |
+| Secrets Provider batch retrieval has partial success | Failed to retrieve Conjur secrets '%s'                       | Error     |
+| Secrets Provider batch retrieval has partial success | Skipped on K8s Secrets '%s'. Reason: failed to retrieve their Conjur secrets | Debug     |
+| Secrets Provider success                             | Successfully updated '%d' out of '%d' K8s Secrets            | Info      |
+| Old batch retrieval endpoint *(for Milestone 1.5)*   | Warning: Secrets Provider cannot efficiently run because Conjur server is not up-to-date. Please consider upgrading to '%d' | Warn      |
+| Job has completed and is terminating                 | Kubernetes Secrets Provider has finished task and is terminating… | Info      |
+| Acknowledge when a retry is taking place             | Retrying '%d' out of '%d' ...                                | Info      |
 
 ### Audit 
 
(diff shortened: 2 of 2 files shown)


design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+## Requirements
+
+The solution should support the following capabilities:
+
+###  Milestone 1 *(current)*
+
+- Secrets Provider runs as a separate entity, serving multiple application containers that run on multiple pods
+- Solution needs to be native to Kubernetes
+- Lifecycle should support update/removal of deployment
+- Provide a way for customer to understand the state of the Secret Provider - when it finished initializing
```
oburstein-hub, line 77: I think we should provide more details on how will the customer understand the state of secret provider and what are the features we should develop for it

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+| 1    | Keep using only **namespace** and/or **service account** granularities. | \- Since Secrets Provider runs as a separate entity, we suggest creating a dedicated service account for Secrets Provider.<br />That way, the **service account** granularity serves as an exact match and **Job** granularity is redundant<br />- No code changes | \- Service account might be reused, which allows other apps to authenticate using Secrets Provider host<br />- Break application identity convention | Free              |
+| 2    | Add support for **Job** granularity in Conjur                | Allows specific identification of the Secrets Provider **Job**, which enhances security | \- Costly; requires code changes, verify compatibility, tests, docs<br />- Redundant if **service account** serves only the Secrets Provider | 10 days           |
+
+*Decision*: Solution #1, use **namespace** and **service account** to define the Secrets Provider host in Conjur policy. We would recommend to the customer to create a dedicated service account for Secrets Provider.
```
oburstein-hub, line 123: Agree

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+
+2. Create `values.yaml` that will contain the following parameters and their defaults:
+
+   2.1. Service Account for Secrets Provider (default: `secrets-provider-account`)
+
+   2.2. Role (default: `secrets-provider-role`)
+
+   2.3. RoleBinding Name (default: `secrets-provider-role-binding`)
+
+   2.4. `K8S_SECRETS` (default: `[]`)
```
oburstein-hub, line 139: please verify if other parameters are required - like conjur url and authenticator url

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+
+We will supply the customer with a  `values.yaml` file where all default parameters are collected. Our `values.yaml` will resemble the following:
+
+```
+# K8s permissioning defaults
+permission:
+	serviceAccountName: secrets-provider-account
+	roleName: secrets-provider-role
+	roleBinding: secrets-provider-role-binding
+	rules:
```
oburstein-hub, line 153: Why do we need to provide rules section here ? I think it should be hidden inside helm

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+# Secrets Provider environment variable defaults
+secrets-provider-environment:
+	conjurApplianceUrl: https://conjur-follower.default.svc.cluster.local
+	conjurAuthnUrl: https://conjur-follower.default.svc.cluster.local/authn-k8s/authn-default
+	conjurAuthnLogin: host/conjur/authn-k8s/authn-default/default
+	conjurAccount: default
+	conjurSslCertificate: conjur-master-ca-env
+	debug=true
+	k8sSecrets=[]
+	containerMode=init
```
oburstein-hub, line 173: I think containerMode should be removed

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+	namespace: default-namespace
+	ttlSecondsAfterFinished: 0	# How long the Job should stay alive even after finishing task. Used for logging.
+	
+# Secrets Provider environment variable defaults
+secrets-provider-environment:
+	conjurApplianceUrl: https://conjur-follower.default.svc.cluster.local
+	conjurAuthnUrl: https://conjur-follower.default.svc.cluster.local/authn-k8s/authn-default
+	conjurAuthnLogin: host/conjur/authn-k8s/authn-default/default
+	conjurAccount: default
+	conjurSslCertificate: conjur-master-ca-env
```
oburstein-hub, line 170: We should consider to retrieve certificate as part if start script that issues secret provider.
This may improve customer experience

design/milestone_1_2_design_doc.md
oburstein-hub: Will K8S_SECRETS variable contain a full list of all secrets that will be synced ? If yes why do we need step 1 above ?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+
+The lifecycle of the Job is the length of the process. In other words, the time it takes to retrieve the Conjur secrets value and update them in the K8s Secrets. To examine the Secrets Provider Job logs, the customer can configure the Job to stay "alive" by adding **ttlSecondsAfterFinished**, with the desired about of seconds, to their Secrets Provider yaml. Because of the nature of a Job, a manual delete is not necessary.
+
+### Order of Deployment
+
+#### Milestone 1
+
+Steps to follow for successful deployment
+
+1. Add all necessary K8s Secrets to `custom-values.yaml`
```
oburstein-hub, line 341: Maybe we missed here creation of configmap with secrets to be synced

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+Customer will install our **Helm Chart** to deploy the Secrets Provider as a **Job**. Once installed, the Job spins up and authenticates to Conjur/DAP via authn-k8s.
+It will then fetch all the K8s secrets update them with the Conjur secrets they require and terminate upon completion.
+
+#### How does a Job installed via Helm answer the requirements?
+
+Running as a **Job** allows separation from the applications' deployment and serve multiple applications at once. 
+
+Because for this Milestone we are concerned with updating K8s Secrets once at intial spin up, a **Job** is the most native solution. It will terminate upon task completion and not waste customer's resources.
+
+Kubernetes Helm is a tool that streamlines and organizes the management of Kubernetes installation, deployment, and upgrade processes. When we use Helm for deployment and lifecycle management, we can provide our customer's a one-click solution even though the way at which we deploy changes.
```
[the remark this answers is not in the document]
sigalsax: What do you mean? Do you mean that the upgrade is still a manual process but HELM makes it easy to do so?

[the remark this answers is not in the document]
oburstein-hub: I meant that in helm you can do only install or uninstall.

design/milestone_1_2_design_doc.md
sigalsax: I added that we will supply a log detailing if it is a job or a deployment
> Kubernetes Secrets Provider v\*%s\* starting up as a %s... (Job/Deployment)

Is that enough in your opinion?

oburstein-hub: I think we should add some indication that secret provider is stuck - maybe using liveliness probe.
Also - we should add a printout saying when the secrets where lately copied from conjur, as well as error message with secrets whose retrieval / copy failed

design/milestone_1_2_design_doc.md
sigalsax: they are, I just didn't detail each one of them... but you make a good point. I removed the list and added
> For a full list of required parameters, see `values.yaml` below

> Will K8S_SECRETS variable contain a full list of all secrets that will be synced ? If yes why do we need step 1 above 

What do you mean here? `K8S_SECRETS` are all the K8s secrets the user wants Conjur values for. Step 1, just says that inside the K8s secret, add conjur-map. Like so:
```
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
type: Opaque
stringData:
  conjur-map: |-
    secret: secrets/test_secret
```
This is how we know the path to the Secret in Conjur

design/milestone_1_2_design_doc.md
sigalsax: moving, good point

design/milestone_1_2_design_doc.md
sigalsax: @orenbm can we do this in terms of the authn-client? Does it look for either 'side-car' or 'init'?

design/milestone_1_2_design_doc.md
sigalsax: @InbalZilberman what do you think? If we have a small script that runs before the secrets provider that can drop down into the follower container and fetch the ssl certificate and fill the configMap? That way the customer won't need to do it manually. 

Note this adds time to delivery

design/milestone_1_2_design_doc.md
sigalsax: What do you mean?

design/milestone_1_2_design_doc.md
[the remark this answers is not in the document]
sigalsax: I see this exists 
https://helm.sh/docs/helm/helm_upgrade/

design/milestone_1_2_design_doc.md
oburstein-hub: I suggest not event to fill the configmap, but to pass it as an environment var directly to secret provider.
It will be exactly the same like we provided PVWA certificate to PAS Replicator in DAPNG101

design/milestone_1_2_design_doc.md
sigalsax: Failures in secret retrieval already exist. For example:

```
const CSPFK020E string = "CSPFK020E Failed to retrieve k8s secret"
const CSPFK021E string = "CSPFK021E Failed to retrieve k8s secrets"
const CSPFK022E string = "CSPFK022E Failed to update k8s secret"
const CSPFK023E string = "CSPFK023E Failed to update K8s secrets"
```
These log entries are timestamped

in terms of liveliness probe under what types of circumstances would the secrets provider be stuck?

design/milestone_1_2_design_doc.md
oburstein-hub: OK, I understood that secrets list  will be provided to helm in config-values.yaml

design/milestone_1_2_design_doc.md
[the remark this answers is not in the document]
oburstein-hub: yes - but is not always works and thus not recommended

design/milestone_1_2_design_doc.md
oburstein-hub: it can be bug, out-of-memory on computer, network issue or any other problem

sigalsax: This seems like an addition in Milestone 2 when we are deployed as a deployment and not a job but this is a good point. 

From what I see [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/): 

A ping must be sent to /tmp/healthy or /healthz (using HTTP) of the Secrets Provider and expect a 0 or return 200 response which I can imagine we can wrap that in code. Regarding `periodSeconds` for how often the Secrets Provider should be probed, how often is conventional?

`initialDelaySeconds` for how long the initial probe should wait before performing the first probe, what is conventional? 5 or 10 seconds?

We can also via HTTP request but why after X seconds (in the link HTTP example, 10 seconds) an unhealthy response is returned? why is there a limit for how long the container is deemed healthy?

oburstein-hub: Sigal - what you wrote looks a right direction. 
regarding periodSeconds -the typical value is 3 seconds
initialDelaySeconds - you can try 5 seconds
There is a limit of time after which if container does not return a response it is considered to be stuck - exactly the same way as you consider the regular application stuck if it does not respond

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+  * [Logs](#logs)
+  * [Documentation](#documentation)
+  * [Open questions](#open-questions)
+  * [Implementation plan](#implementation-plan)
+
+## Glossary
+
+| **Term**                                                     | **Description**                                              |
+| ------------------------------------------------------------ | ------------------------------------------------------------ |
+| [Job](*https://kubernetes.io/docs/concepts/workloads/controllers/job/*) | K8s entity that runs a pod for one-time operation. Once the pod exists, the Job terminates |
```
orenbm, line 47: Once the pods exists? Once a pods exits?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+
+Because for this Milestone we are concerned with updating K8s Secrets once at intial spin up, a **Job** is the most native solution. It will terminate upon task completion and not waste customer's resources.
+
+Kubernetes Helm is a tool that streamlines and organizes the management of Kubernetes installation and deployment processes. When we use Helm for deployment and lifecycle management, we can provide our customer's a one-click solution even though the way at which we deploy changes.
+
+#### What drawbacks does this solution have?
+
+**Job** is not an existing application identity granularity.
+
+ As a result, one can define the host representing the Secrets Provider in Conjur policy using only **namespace** and/or **service account** granularities.
```
orenbm, line 114: namespace is required. service-account is optional

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+  namespace: {{ .Values.secrets-provider.namespace }}
+spec:
+{{- end }}
+  template:
+    serviceAccountName: {{ .Values.serviceAccount }}
+    containers:
+    - image: {{ .Values.secrets-provider.image }}
+        imagePullPolicy: Always
+        name: {{ .Values.secrets-provider.name }}
+        env:
```
orenbm, line 241: I think that the following are missing from the env:
- `SECRETS_DESTINATION` - is there a default value for this?
- `CONTAINER_MODE` (note that the default value is `sidecar`. what is the value that you want to have?)
- `K8S_SECRETS`
  - I know that `K8S_SECRETS_LABELS` replaces this. Do you think we should add this as an optional param?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+
+*Decision*: Solution #1, use **namespace** and **service account** to define the Secrets Provider host in Conjur policy. We would recommend to the customer to create a dedicated service account for Secrets Provider.
+
+#### Customer experience
+
+Deployment for the Secrets Provider will be done using Helm. That way, if the 
+
+1. Configure K8s Secrets with ***conjur-map\*** metadata defined
+
+2. Create `values.yaml` that will contain the following parameters and their defaults:
```
orenbm, line 131: is `values.yaml` a known best practice for helm? If not, how about `secrets-provider-values.yaml`?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+   *When parameters are not configured by customer, their defaults are used*
+
+3. Install Helm Chart for the Secrets Provider
+
+We will supply the customer with a  `values.yaml` file where all default parameters are collected. Our `values.yaml` will resemble the following:
+
+```
+# K8s permissioning defaults
+permission:
+	serviceAccountName: secrets-provider-account
```
orenbm, line 144: i would call it `secrets-provider-service-account` to be more precise

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+	name: cyberark-secrets-provider
+	namespace: default
+	ttlSecondsAfterFinished: 0	# Length at which Job should stay alive even after finishing task. Used for logging.
+	
+# Secrets Provider environment variable defaults
+secrets-provider-environment:
+	conjurApplianceUrl: https://conjur-follower.default.svc.cluster.local
+	conjurAuthnUrl: https://conjur-follower.default.svc.cluster.local/authn-k8s/authn-default
+	conjurAuthnLogin: host/conjur/authn-k8s/authn-default/default
+	conjurAccount: default
```
orenbm, line 160: does it make sense to have these as default values? do we see a use-case in which the account will be `default`, the service-id will be `authn-default` or the host-id will be `default`?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+	conjurSslCertificate: conjur-master-ca-env
+	debug=true
```
orenbm, line 162: is it a good idea to have debug logs enabled by default?

design/milestone_1_2_design_doc.md
orenbm: the default value is `sidecar` so i don't think it's the behaviour you are looking for. In case of `init` we exit the container after the first cycle. in case of `sidecar` we continue to perform the cycle after some interval.

orenbm: however, i think that providing `init` as the value where the container is not an init container is a weird behaviour. Ideally, if any value would have worked for your purposes, what would you call it?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+Variables from the `values.yaml` / `custom-values.yml `are applied for our project as follows:
+
+*templates/secrets-access-role.yaml*
+
+```
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.permission.serviceAccount }}
```
orenbm, line 182: it is called `serviceAccountName`

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+
+   *When parameters are not configured by customer, their defaults are used*
+
+3. Install Helm Chart for the Secrets Provider
+
+We will supply the customer with a  `values.yaml` file where all default parameters are collected. Our `values.yaml` will resemble the following:
+
+```
+# K8s permissioning defaults
+permission:
```
orenbm, line 143: what about `rbac`?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Values.permission.roleName }}
+rules:
+  - apiGroups: [""]
+    resources: {{ .Values.permission.rules.resources }}
+    verbs:
+    	rules:
```
orenbm, line 193: what's this?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+
+The following template takes into account the anticipated flows for Milestone 1 and 2.
+
+```yaml
+---
+{{- if eq .Values.secrets-provider-environment.rotationSupport "false" }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}
```
orenbm, line 223: where does this come from?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+  name: {{ .Values.permission.roleName }}
+```
+
+*templates/secrets-provider.yaml*
+
+The following template takes into account the anticipated flows for Milestone 1 and 2.
+
+```yaml
+---
+{{- if eq .Values.secrets-provider-environment.rotationSupport "false" }}
```
orenbm, line 219: supportRotation?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Values.secrets-provider.namespace }}
+spec:
+	ttlSecondsAfterFinished: {{ .Values.secrets-provider.ttlSecondsAfterFinished }}
```
orenbm, line 226: what happens if we don't provide this field?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+maintainers:
+- email: conj_maintainers@cyberark.com
+  name: Conjur Maintainers
+name: secrets-provider
+version: 1.0.0
+```
+
+Helm Charts are packaged in `tgz` format. 
+
+For each release, we will package our Chart (`helm package secrets-provider`) and the output will be `secrets-provider-<version>.tgz`. This `.tgz` is what we will push to Github.
```
orenbm, line 317: will this be automated?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+
+For more detailed information on how a customer will install our Chart, see the section on [Installation](#installation).
+
+### Milestone 2: Rotation
```
orenbm, line 321: should we review this?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+
+All fetches on a Conjur Resource are individually audited, creating its own audit entry. Therefore there will be no changes in audit behavior for this Milestone.
+
+## Documentation
+
+
+
+## Open questions
+
+1. Once the Job terminates, there is no way to access the Job's logs. **ttlSecondsAfterFinished** allows the Job to stay alive even after the task terminates. Do we want to supply a **ttlSecondsAfterFinished** default for customers so that they can evaluate Job logs even after the Job terminates?
```
orenbm, line 438: this is quite a big question. will this happen also if the secrets-provider failed to retrieve secrets or is it just after a successful cycle?

Also - i've seen that the default value you gave is `0`. Does that mean that by default the user will not be able to see logs? Is this the behaviour we want?

design/milestone_1_2_design_doc.md
orenbm: I have a related question - I didn't see in the `ConfigMap` creation in the helm chart. Do we expect the customer to create it?

How about letting the customer add the ssl_cert in `values.yaml` and adding this to the helm:
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: conjur-master-ca-env
  labels:
    app: test-env
data:
  ssl-certificate: |
    {{ .Values.secrets-provider-environment.conjurSslCertificate }}
```

this way the ssl_cert name (`conjur-master-ca-env`) is not something that the user even needs to bother with (or we can still have it in `values.yaml` under `conjurSslCertificateName`.

design/milestone_1_2_design_doc.md
sigalsax: Adding any value for `containerMode` would demand updating also in `authn-client` right?
In terms of naming, maybe `job` (Milestone 1) and `consistentDeployment` (Milestone 2). What did you think?

design/milestone_1_2_design_doc.md
sigalsax: Good call and I like your idea... Are we okay though with saving secrets (like ssl) in files?

sigalsax: one clarification, `values.yaml` is for default parms we ship to customers. `custom-values.yaml` is customer supplied and used for overriding the defaults. Did you mean 
> How about letting the customer add the ssl_cert in... `custom-values.yaml`?

design/milestone_1_2_design_doc.md
sigalsax: the pod that the Job brings up

design/milestone_1_2_design_doc.md
sigalsax: Good catch! Fixing this now.

That reminds me...
In order for a customer to add a new K8s Secret to the K8S_SECRETS, they will need to tear down the Chart and reinstall it with the updated list of `custom-values.yaml`. AKA tear down the deployment
Adding this to the limitations...

sigalsax: > I know that `K8S_SECRETS_LABELS` replaces this. Do you think we should add this as an optional param?

I want to clarify, adding `K8S_SECRETS_LABELS` doesn't replace the ability for the customer to supply use with `K8S_SECRETS` values. If a customer would supply us with both, we would take both values and query for them in the K8s namespace.
I forgot `K8S_SECRETS` though, adding now!

sigalsax: I am however wrap it in an if clause like the following

```
 {{- if eq .Values.secrets-provider-environment.k8sSecrets }}
   - name: K8S_SECRETS
     value: {{ .Values.secrets-provider-environment.k8sSecrets }}
{{- end }}			 
```

b/c in later milestones it is not required and it is enough that a customer will just have `K8S_SECRETS_LABELS`

design/milestone_1_2_design_doc.md
sigalsax: From what I understood, `values.yaml` is the standard defaults file
@oburstein-hub 

Now that I think of it, could a customer ideally supply us with multiple files that override the values.yaml and load them in like so:
```
`helm install conjur -f custom-values.yaml custom-values2.yaml custom-values3.yaml https://github.com/cyberark/secrets-provider-for-k8s/helm/secrets-provider-1.0.0.tgz`
```

design/milestone_1_2_design_doc.md
sigalsax: I do think that there should be defaults for these values. Customers take what is int he `values.yaml` file and copy them over to their customer values file. So if anything, it is a guide of what the customer needs to configure.

@orenbm  Is that your concern or do you agree that there should be defaults and just not agree with the values I gave to the defaults?

@oburstein-hub do you agree? do all configurations need to have defaults?

design/milestone_1_2_design_doc.md
sigalsax: This is a tricky one because any change to the customer's `custom-values.yaml` demands a restart of the HELM Chart which means tearing down the Secrets Provider...
So if they have a problem in their env and they want to see debug logs, they cannot unless they do so. I guess that this aligns with the current experience for the secrets provider no?

Seems like a wonky experience for this one which is why I wanted to default to debug.. What do you think?

design/milestone_1_2_design_doc.md
sigalsax: instead of `permissions`?

design/milestone_1_2_design_doc.md
sigalsax: whoops. updating

design/milestone_1_2_design_doc.md
sigalsax: `.Release.Name` comes from the `Chart.yaml` so for our purposes
`name: secrets-provider`

design/milestone_1_2_design_doc.md
sigalsax: like it

design/milestone_1_2_design_doc.md
sigalsax: we will take the default

design/milestone_1_2_design_doc.md
sigalsax: We definitely could but IMO there isn't much to automate here just that command and `git push`. Do you disagree?

design/milestone_1_2_design_doc.md
sigalsax: Nothing to review yet for this one. There are Milestone 2 crumbs sprinkled across the doc but will hope to finalize this section today

design/milestone_1_2_design_doc.md
sigalsax: If the secrets provider fails, it will behave like how a normal pod that fails behaves. `ttlSecondsAfterFinished` is for finished resources

> Does that mean that by default the user will not be able to see logs? 
Yes

> Is this the behaviour we want?
No not a good idea. How long do you think it should stay up? How long does our init container stay up before terminating? 60 secs?

@oburstein-hub what do you think the default `ttlSecondsAfterFinished` should be?

sigalsax: @InbalZilberman @oburstein-hub 
I notice [here](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs) that `ttlSecondsAfterFinished` is an alpha feature.
I see here that in order to use this feature the customer might need to enable alpha features in their GKE cluster with something like:
`kube-apiserver --feature-gates=TTLAfterFinished=true` AND in their kube-controller-manager because it isn't enabled by default

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,455 @@
...
+```
+# K8s permissioning defaults
+permission:
+	serviceAccountName: secrets-provider-account
+	roleName: secrets-provider-role
+	roleBinding: secrets-provider-role-binding
+
+# Secrets Provider pod defaults
+secrets-provider:
+	image: cyberark/secrets-provider-for-k8s
```
eladkug, line 150: Which version it suppose to take here?

design/milestone_1_2_design_doc.md
oburstein-hub: I think that these fields do need defaults as they appear here - because it is the most typical case

design/milestone_1_2_design_doc.md
oburstein-hub: debug default value should be false. Regarding thaging debug level on the fly - it should be possible with some changes i n secret provider if we provide it as a file and not as env var

eladkug: We need to challenge this. I think the default need to be false.

design/milestone_1_2_design_doc.md
oburstein-hub: I think we should distinguish here between mode of running of secret provider process and the parameters provided in values.yaml. We should not expose a parameter of running mode of secret provider in helm chart, because it should be calculated from the fact that it is ran as Job or Deployment, and that it is not init container in this case

design/milestone_1_2_design_doc.md
oburstein-hub: Basically this is what is done in these case now, but it would be great (only if you have a time) to add 'helm package' command to repo's jenkinsfile

design/milestone_1_2_design_doc.md
oburstein-hub: If it is alpha feature we cannot use it at all IMO
I will look to find common ways of retrieving logs for Job

oburstein-hub: Please see what I found: When a Job completes, no more Pods are created, but the Pods are not deleted either. Since they are terminated, they don’t show up with kubectl get pods, but they will show up with kubectl get pods -a. Keeping them around allows you to still view the logs of completed pods

oburstein-hub: Sigal - please try this, and if it works you don't need any special parameter to see logs of job that finished to run

design/milestone_1_2_design_doc.md
oburstein-hub: how will the customer provide this secrets-provider-environment.conjurSslCertificate to custom-values.yaml ? If you will just put it there it is not secure. You can provide it via helm command line (there is a way to provide Values in command line),.

design/milestone_1_2_design_doc.md
orenbm: @sigalsax we will not need to perform changes in the authn-client if we add a new value (want to look at the code and see why for yourself?). However, if we choose a different value than `init` and still want it to terminate after one cycle then we will need to add some code in our own repo.

i actually don't think there's a better value for `containerMode`. However, we can find another solution for finishing the secrets-provider after the first cycle. Do you have any suggestion to which of the current env vars we can use for that?

design/milestone_1_2_design_doc.md
oburstein-hub: right

design/milestone_1_2_design_doc.md
orenbm: yes, in that case then `custom-values.yaml` is better. Is this what we expect to have required values that do not have default values?

orenbm: > Are we okay though with saving secrets (like ssl) in files?

well, that's something to check with @shaharglazner

design/milestone_1_2_design_doc.md
orenbm: so the Job terminates when the pod is up? Isn't the pod part of the Job?

design/milestone_1_2_design_doc.md
orenbm: yes, i guess both need to be conditional.

design/milestone_1_2_design_doc.md
orenbm: is this a question or a statement?

design/milestone_1_2_design_doc.md
orenbm: i don't think we should have defaults for these values. What's the common way to state required fields that don't have defualt values?

design/milestone_1_2_design_doc.md
orenbm: i also agree that the default should be false. we can find a solution to reload the debug level on runtime if we want (do you have an idea?)

design/milestone_1_2_design_doc.md
oburstein-hub: > yes, in that case then `custom-values.yaml` is better. Is this what we expect to have required values that do not have default values?

it can be done for every values - not only for those which don't have defaults.

design/milestone_1_2_design_doc.md
orenbm: yes. seems more fitting for service-account, role and rolebinding

design/milestone_1_2_design_doc.md
orenbm: oh i see. thanks!

design/milestone_1_2_design_doc.md
orenbm: i think that any process that needs to happen should be automated if possible. we already have some automated process for releases so we can add it there.

design/milestone_1_2_design_doc.md
orenbm: Nice @oburstein-hub ! then in this case we just need to update our docs to mention that.

design/milestone_1_2_design_doc.md
oburstein-hub: they terminate together

orenbm: so `exits` seems more fitting

design/milestone_1_2_design_doc.md
sigalsax: > how will the customer provide this secrets-provider-environment.conjurSslCertificate to custom-values.yaml ? If you will just put it there it is not secure. You can provide it via helm command line (there is a way to provide Values in command line)

So you are saying not to have the customer copy + paste to custom-value, rather use a helm command for populating?

design/milestone_1_2_design_doc.md
sigalsax: They terminate together but are not deleted.

See here: https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup Specifically,
> When a Job completes, no more Pods are created, but the Pods are not deleted either

design/milestone_1_2_design_doc.md
sigalsax: deleting this conditional for now because we are focusing on milestone 1 charts. This conditional is relevant for milestone 2

design/milestone_1_2_design_doc.md
oburstein-hub: mostly yes. the  cases that values.yaml parameters are passed via command line are usually the cases when we don't want the used to see / store them. for example:
helm install abc -set certificate=calculateCert() secret-provider-helm.tgz

design/milestone_1_2_design_doc.md
sigalsax: There are values that are defaults and values that are mandatory. Those with defaults should be in `values.yaml` and those that are mandatory OR a customer wants to override the defaults should be in `custom-values.yaml`. I will add that distinction

design/milestone_1_2_design_doc.md
sigalsax: default can be false but if they supply us with `debug: true` when loading the HELM, we will update. Adding the following:
```
{{- if eq .Values.secrets-provider-environment.debug "true" }}
   - name: DEBUG
      value: true
{{- end }} 
```

design/milestone_1_2_design_doc.md
sigalsax: like it

design/milestone_1_2_design_doc.md
sigalsax: If I am understanding correctly, during the `publish` step in Jenkins where we are pushing to our Docker registry and RH registry, add the Helm package and git add/commit/push step?

design/milestone_1_2_design_doc.md
sigalsax: yeah, nice find! also came across this but I will double check this for sanity purposes

design/milestone_1_2_design_doc.md
oburstein-hub: helm code is pushed to github registry when you are doing git push. It issues run of jenkins job. One of steps of jenkins job is build, which will execute 'helm package secret-provider-helm' and create secret-provider-helm.tgz. 

You may also try to look how the conjur-oss-helm.tgz is created and to it the same

design/milestone_1_2_design_doc.md
orenbm: i think this is a good solution

design/milestone_1_2_design_doc.md
sigalsax: but from my understanding `secret-provider-helm.tgz` is the artifact that needs to be pushed to GH as well

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,471 @@
...
+
+These variations will be supported using the same Secrets Provider image, but the behavior will vary dynamically depending on the chosen deployment method that is configured via Helm Chart.
+
+*In this document we will focus on **Milestone 1***, deploying the Secrets Provider as a Job
+
+### Milestone 1: Serve K8s secrets to multiple applications once (no rotation)
+
+### Design
+
+Customer will install our **Helm Chart** to deploy the Secrets Provider as a **Job**. Once installed, the Job spins up and authenticates to Conjur/DAP via authn-k8s.
```
Tovli, line 99: What about the init container? as we add HELM we should enable deploying as init container as well to better support our current users

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,471 @@
...
+
+ As a result, one can define the host representing the Secrets Provider in Conjur policy using only **namespace** with/without **service account** granularities.
+
+To handle the missing **Job** granularity there are 2 options:
+
+|      | Solution                                                     | Pros                                                         | Cons                                                         | Effort Estimation |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------- |
+| 1    | Keep using only **namespace** and/or **service account** granularities. | \- Since Secrets Provider runs as a separate entity, we suggest creating a dedicated service account for Secrets Provider.<br />That way, the **service account** granularity serves as an exact match and **Job** granularity is redundant<br />- No code changes | \- Service account might be reused, which allows other apps to authenticate using Secrets Provider host<br />- Break application identity convention | Free              |
```
Tovli, line 120: Can you elaborate on why it breaks application identity (a.k.a machine identity) convention?

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,471 @@
...
+
+# Secrets Provider pod defaults
+secrets-provider:
+  image: cyberark/secrets-provider-for-k8s
+  name: cyberark-secrets-provider
+	
+# Secrets Provider environment variable defaults
+secrets-provider-environment:
+  conjurSslCertificateName: secrets-provider-ssl-config-map
+  containerMode: init
```
Tovli, line 163: Don't we want to have the use case as the parameter? so instead of `containerMode` we can have something like `serveMultipleApps: false` for init

design/milestone_1_2_design_doc.md
Tovli: @sigalsax can you please organise this thread and see all concerns have been addressed?
It got multiple subjects already

Tovli: >I want to clarify, adding K8S_SECRETS_LABELS doesn't replace the ability for the customer to supply use with K8S_SECRETS values

Shouldn't we deprecate K8S_Secrets? it seems we are adding complexity to our product. IMHO we should at least log a warning if the customer use it that it might be deprecated in the future

design/milestone_1_2_design_doc.md
Tovli: >helm code is pushed to github registry when you are doing git push

What is github registry? you mean [github packages](https://github.com/features/packages)?

design/milestone_1_2_design_doc.md
oburstein-hub: We should decide with @InbalZilberman if we are going to support K8S_SECRETS. Currently there is misalignment in this area

design/milestone_1_2_design_doc.md
oburstein-hub: I meant github repo (which will be dedicated for this purpose)

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,471 @@
...
+home: https://www.cyberark.com
+icon: https://xebialabs-clients-iglusjax.stackpathdns.com/assets/files/logos/CyberArkConjurLogoWhiteBlue.png
+keywords:
+- security
+- secrets management
+maintainers:
+- email: conj_maintainers@cyberark.com
+  name: Conjur Maintainers
+name: secrets-provider
+version: 1.0.0
```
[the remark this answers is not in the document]
sigalsax: Helm version as in 3.2.4 ? or the `version` here for the `Chart.yaml`?

design/milestone_1_2_design_doc.md
sigalsax: deploying an init container with Helm?

design/milestone_1_2_design_doc.md
sigalsax: Because we currently have a Deployment application identity to represent the type deployment
 `kind: deployment`. We are suggesting that for Job, we don't have a Job granularity even though the secrets provider will be deployed as a job `kind: job`

design/milestone_1_2_design_doc.md
sigalsax: containerMode is value that comes from the authn client. We didn't want to change the containerMode value for now because the job is still acting like an init container where, although not attached to a certain application, it runs before all the pods, spins up to perform it's task and terminates upon completion

design/milestone_1_2_design_doc.md
sigalsax: Our plan for the `K8S_SECRETS_LABELS` was not to deprecate the `K8S_SECRETS`, rather allow them to live side by side. We spoke about K8S_SECRETS_LABELS being an enhancement for a later milestone but will bring this up again

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,430 @@
...
+
+The current implementation of the Secrets Provider only supports one application and makes the deployment of the Secrets Provider deeply coupled to the customer's applications. 
+
+## Requirements
+
+The solution should support the following capabilities:
+
+###  Milestone 1 *(current)*
+
+- Secrets Provider runs as a separate entity, serving multiple application containers that run on multiple pods
```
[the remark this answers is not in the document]
orenbm: @InbalZilberman are you suggesting that the secrets-provider will serve more than one namespace and have only the service-account in its application identity for Conjur authentication?

@oburstein-hub is it possible to have a service-account that has access to secrets in more than one namespace?

design/milestone_1_2_design_doc.md
orenbm: not saying it's not possible but how can it be done when we don't know how the application container looks like?

design/milestone_1_2_design_doc.md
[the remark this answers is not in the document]
oburstein-hub: it is possible, but them you need to use ClusterRole / ClusterRoleBinding instead of Role / RoleBinding

design/milestone_1_2_design_doc.md
orenbm: i'll just clarify something - `containerMode` is set in the config of the authn-client but is used in the secrets-provider for terminating the container after one cycle if it is set to `init`. If we don't provide it, and instead query on some other variable then everything will be fine. 

I tried to hint in that direction in another comment i gave in this doc. I think that this is the correct approach. `container_mode = init` is pretty weird when the container is not an init container. 

However, i don't think that `serveMultipleApps` will give us what you are looking for as this is not the question we want to ask when we terminate the pod or wait for another cycle. i think that `performRoatation` (or something like that) will be more fitting and will give what you need. this way you can remove the `containerMode`.

design/milestone_1_2_design_doc.md
oburstein-hub: deploying init container in helm is not a common thing to do and also problematic technically.
I dont think we should do it

design/milestone_1_2_design_doc.md
orenbm: this is the other thread - https://github.com/cyberark/secrets-provider-for-k8s/pull/151#discussion_r454826482

design/milestone_1_2_design_doc.md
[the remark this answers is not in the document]
orenbm: thanks.

Then @InbalZilberman it seems that it is technically possible but it will require changes in Conjur as today `namespace` is required for authentication. we can change it so that the requirement will be `namespace` *OR* `service-account`.

Coming to think about it, it is possible today because the `namespace` that we check is the namespace in which the secrets-provider pod will be deployed in. So the application identity will still have the `namespace` attribute even if the service account serves multiple namespaces.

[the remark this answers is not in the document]
sigalsax: > are you suggesting that the secrets-provider will serve more than one namespace and have only the service-account in its application identity for Conjur authentication?

@InbalZilberman please clarify (:

design/milestone_1_2_design_doc.md
sigalsax: as in this secrets-provider-for-k8s repo correct?

design/milestone_1_2_design_doc.md
sigalsax: > If we don't provide it, and instead query on some other variable then everything will be fine.

Are you saying that if we another value and it will pass even if it was `containerMode=foobar` and we are only checking for init?

orenbm: correct. I encourage you to look in the code of the secrets-provider and the authn-client to see for yourself :)

sigalsax: or just add the environment variable `performRotation=false` in the manifest and check for that and add in code
```
if authnConfig.ContainerMode == "init" || authncConfig.performRotation=="false" {
     os.Exit(0)
}
```

orenbm: yes, that's exactly what I was thinking about

design/milestone_1_2_design_doc.md
oburstein-hub: yes

design/milestone_1_2_design_doc.md
sigalsax: @oburstein-hub the ssl is not private.. it is the public SSL certificate value so it sitting in custom-values of the customer is not a security concern. @shaharglazner 
See here for explanation: https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/Kubernetes_deployApplicationsConjur-k8s-Secrets.htm?Highlight=secrets%20provider

oburstein-hub: in this case it can be in custom-values.yaml

shaharglazner: @sigalsax @oburstein-hub afaiu the discussion, the certificate itself is a public piece of information (think of google.com certificate which can be retrieved without any permissions/restrictions). the sensitive information is the private key that signed that certificate, and should never leave the server.

design/milestone_1_2_design_doc.md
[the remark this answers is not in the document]
abrahamko: @oburstein-hub We want to use Helm 3, which fixes the support for upgrade. See [Helm 3 Upgrade](https://helm.sh/docs/faq/#improved-upgrade-strategy-3-way-strategic-merge-patches)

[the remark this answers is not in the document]
oburstein-hub: you can try t do it if you have enough time.

design/milestone_1_2_design_doc.md
abrahamko: @oburstein-hub @sigalsax I think we should use [Helm Package in GitHub](https://helm.sh/docs/topics/chart_repository/#github-pages-example)

design/milestone_1_2_design_doc.md
```diff
@@ -350,25 +350,27 @@ For a detailed breakdown of the decision process see [here](batch_retrieval_desi
...
+|      | Solution                               | Elaboration                                                  | Pros                                                         | Cons                                                         | Effort estimation |
+| ---- | -------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ----------------- |
+| 1    | Client-side polling for secrets        | Every configurable amount of time, contact Conjur API server.<br />Possible solutions can be:<br />1. Retrieve all secrets<br />2. Check if there's something new for the requesting host<br />3. Retrieve secrets that were changed since last request | - Straight-forward<br />- Simple<br />- No additional requirements needed | - Regularly sends API calls even if nothing has changed<br />- Introduces load on the server to hanlde many secrets providers | 15 days           |
```
sigalsax, line 357: Unclear what is the "something new"
> Check if there's a secret update for the requesting host

sigalsax: what does "No additional requirements needed" mean?

sigalsax: >  Introduces load on the server to hanlde many secrets providers

Load on server to handle additional requests, especially with many secrets providers

design/milestone_1_2_design_doc.md
```diff
@@ -350,25 +350,27 @@ For a detailed breakdown of the decision process see [here](batch_retrieval_desi
...
 
-(*) As a possible fallback, we can provide the user with a warning that the endpoint feature is not available in their Conjur version and fail back to Solution #1.
+We decided to go with option No. 1.
+Low level design of the solution will be introduced in a separate document.
 
-#### Concerns
+#### Customer experience
 
-*Load on Server*
+This solution does not introduce any additional requirements for the solution besides using `Deployment` instead of a `Job`.
```
sigalsax, line 366: > any additional requirements

what do you mean by additional requirements? do you mean any addition user intervention?  as in the customer doesn't have to perform any additional tasks?

design/milestone_1_2_design_doc.md
```diff
@@ -350,25 +350,27 @@ For a detailed breakdown of the decision process see [here](batch_retrieval_desi
...
+We might add some variables, all with reasonable defaults as to ensure no additinal steps are needed to migrate from Milestone 1 to 2.
 
-To mitigate this concern we will test performance and supply our customers with clear limitations, detailing how many Secrets Providers and Followers can be used for optimal performance.
+#### Concerns
 
-#### Customer experience
+*Load on Server*
 
-A decision has not been determined. Regardless of the final decision for rotation, we do not anticipate any change in terms of the customer experience when updating from Milestone 1 to 2. 
+To assuage this concern we will test performance and supply our customers with clear limitations, detailing how many Secrets Providers and Followers can be used for optimal performance.
```
sigalsax, line 373: > assuage

NICE!

design/milestone_1_2_design_doc.md
abrahamko: @orenbm Now you can...

design/batch_retrieval_design.md
```diff
@@ -0,0 +1,20 @@
+# Batch retrieval design
+
+This document will cover the options for enhancing Batch Retrieval requests.
+
+### Motivation
+
+*Previously*, for the init container solution, each Secrets Provider was sitting within the same Deployment as the app. Because of this, when a batch retrieval request was made, the requests were dispersed across many call to the Conjur server. If one K8s Secret was not able to be updated with a Conjur value (for example due to a permission error), then the batch request would fail and only that pod would not spin up successfully. 
+
+*Now* that the Secrets Provider is outside, a single batch retrieval request will be made to Conjur for *all* application. So if there is a failure in receiving one of the values, then the whole request would fail. In other words, **no application** will spin up in the namespace.
```
Tovli, line 9: I don't see any reference to the fact that the authn-k8s which secrets provider use will not work when using it from out side the cluster as @oburstein-hub wrote [here](https://ca-il-confluence.il.cyber-ark.com/display/rndp/K8S+Secret+provider+for+HomeDepot+Solution+Architecture+Issues#/) when using option number 3 we need access to the k8s api from the follower.
Please address the limitations from the doc in this design

oburstein-hub: @Tovli - lets wait with this to this subject closure with Inbal / Naama.

design/milestone_1_2_design_doc.md
sigalsax: Updated the values.yaml to include <version>

design/batch_retrieval_design.md
```diff
@@ -0,0 +1,20 @@
...
+
+|      | Solution                                                     | Pros                                                         | Cons                                                         | Effort (T-shirt size) |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | --------------------- |
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                          | S, 5 days             |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Better / straightforward design for how batch endpoints | - Requires both client/server implementation<br />- Need to handle backwards compatibility | M, 10 days            |
```
AndrewCopeland, line 14: I think we should update batch retrieval on the server side. Keep the current behavior unless a specific query parameter is provided. Such as `ignore` or `continueOnFailure`.

The current endpoint looks like:
```
/secrets?variable_ids=myorg:variable:secret1,myorg:variable:secret1
```

The new endpoint should look like:
```
/secrets?continueOnError=true&variable_ids=myorg:variable:secret1,myorg:variable:secret1
```

On the `secrets-provider-for-k8s` side it should be optional. Such as an environment variable `CONJUR_IGNORE_UNSUCCESSFUL_PASSWORD_RETRIEVAL=true` or something like that.

design/milestone_1_2_design_doc.md
```diff
@@ -473,10 +473,27 @@ Note we will also write tests in Conjur for the new batch retrieval endpoint.
...
+To measure it, we will launch 100, 1000 and 10K goroutines, each sending batch secret retrieval API call to the follower repeatedly during 1 minute. We measure the total amount of secrets received during 1 minute for each, and the result is the maximum number of secrets received.
+
+Since this number may vary depending on many variables, we will do this test for each of the following variables independently:
+
+| Variable                                    | Const value while testing other variables | Values to test              |
+| ------------------------------------------- | ----------------------------------------- | --------------------------- |
+| Number of secrets in each batch request     | 20                                        | 1, 20, 100, 500, 2000, 10K  |
+| Number of Secrets Providers                 | 1                                         | 1, 10, 20, 50, 100          |
+| Secret path length                          | 20 chars                                  | 10, 20, 50, 100, 200 chars  |
+| Secret value length                         | 30 chars                                  | 30, 50, 100, 200, 500 chars |
```
sigalsax, line 494: 500 chars, we need to check what the max number of chars conjur allows

sigalsax: (and K8s for the K8s Secret) and use that number as max

design/milestone_1_2_design_doc.md
```diff
@@ -473,10 +473,27 @@ Note we will also write tests in Conjur for the new batch retrieval endpoint.
...
+
+To answer it, we will measure what is the maximum number of secrets 1 follower can serve.
+To measure it, we will launch 100, 1000 and 10K goroutines, each sending batch secret retrieval API call to the follower repeatedly during 1 minute. We measure the total amount of secrets received during 1 minute for each, and the result is the maximum number of secrets received.
+
+Since this number may vary depending on many variables, we will do this test for each of the following variables independently:
+
+| Variable                                    | Const value while testing other variables | Values to test              |
+| ------------------------------------------- | ----------------------------------------- | --------------------------- |
+| Number of secrets in each batch request     | 20                                        | 1, 20, 100, 500, 2000, 10K  |
+| Number of Secrets Providers                 | 1                                         | 1, 10, 20, 50, 100          |
```
sigalsax, line 492: @InbalZilberman wanted 1000 as a max

design/milestone_1_2_design_doc.md
```diff
@@ -366,8 +366,9 @@ Low level design of the solution will be introduced in a separate document.
 
 #### Customer experience
 
-This solution does not introduce any additional requirements for the solution besides using `Deployment` instead of a `Job`.
-We might add some variables, all with reasonable defaults as to ensure no additinal steps are needed to migrate from Milestone 1 to 2.
+This solution introduce only 1 additional requirement for the customer; Use updated Conjur that includes the new API.
```
sigalsax, line 369: introduces*

sigalsax: .  They must updated their Conjur version to access the new API

sigalsax: Nit all requirements are mandatory so maybe use another word?

> This requirement is not mandatory; If this requirement is not fulfilled, the solution will fallback to option (1.a), with efficiency price. The Secrets Provider chart will be using Deployment instead of a Job. 

"
This is not mandatory. If the requirement if not fulfilled, the solution will fall back to option **1.a** but at the price of efficiency.
<ADD new line>
The Secrets Provider Chart
 "

design/batch_retrieval_design.md
Tovli: As talked with @InbalZilberman - We need to support the secrets provider being outside the follower cluster.
This is actually not new, since we needed to support it also before we had the new requirements

abrahamko: @Tovli I don't understand.
This requirement was never introduced as part of this solution, and is unrelated to anything we wrote here.
@InbalZilberman Added this to the [feature doc](https://github.com/cyberark/secrets-provider-for-k8s/issues/102) only yesterday as a minor sentence for testing.
This is not minor, and it is way more than testing, since testing is not a customer requirement; supporting is.
This need to be handled according to regular R&D requirements, and not as part of this solution design.
Just a reminder, the requirements here are support multiple apps, deploy outside of app lifecycle, and support rotation with seamless upgrade between milestones.
These are all Secrets Provider requirements.
There's nothing regarding follower, much less it's location and availability.

design/batch_retrieval_design.md
oburstein-hub: We should consider here how the reply of the request looks currently

design/batch_retrieval_design.md
orenbm: I strongly agree with @abrahamko on this

design/batch_retrieval_design.md
sigalsax: @AndrewCopeland this is could definitely be an option and might save us time.

@abrahamko ^

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,576 @@
...
+
+##### Lifecycle
+
+The lifecycle of the Secrets Provider is independent of the application pods. The lifecycle of the Job is the length of the process. In other words, the time it takes to retrieve the Conjur secrets value and update them in the K8s Secrets. After the Pod completes successfully, [the Job and Pod will remain](https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup) and the customer can fetch logs on the Job. 
+
+To notify the customer that the Secrets Provider has started and finished it's work, the process will be logged.
+
+##### Upgrade
+
+Because we will be using Helm Charts, we can easily supply customers with the Helm Charts for the current Milestone. All the customer would need to do is update their current Secrets Provider Chart. `helm update -f custom-values.yaml secrets-provider secrets-provider-<version>`
```
sigalsax, line 414: TODO: `helm upgrade`

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,576 @@
...
+| --------- | ------------------------------------------------------------ | ------------------------------------------------------------ | --------------------------------------------- | ------------------------------------------------------------ |
+| 1         | *Vanilla flow*, Secret Provider Job successfully updates K8s Secrets | - Conjur is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and K8s Secrets are configured<br />- All mandatory values are defined in `custom-values.yaml`<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job                | - Secrets Provider pod authenticates and fetches Conjur secrets successfully<br />- All K8s Secrets are updated with  Conjur value <br />- App pods receive K8s Secret with Conjur secret values as environment variable<br />- Verify in logs that deployment type is listed <br />- Secrets Provider Job is completion<br />- Verify logs |
```
sigalsax, line 478: >  Secrets Provider Job is completion

 Secrets Provider Job runs to completion

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,576 @@
...
+| 2         | Validate new batch retrieval functionality                   | Secrets Provider host does not have permissions on Conjur secrets / secret is missing / secret value doesn't exist | Secrets Provider runs as a Job                | - Verify in logs that a list of all failed Conjur secrets and those that were skipped are noted |
```
sigalsax, line 479: TODO
> Verify in logs that a list of all failed Conjur secrets and those that were skipped are noted

and those K8s Secrets that were skipped are noted

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,576 @@
...
+| 3         | *Vanilla flow*, non-conflicting Secrets Provider             | Two Secrets Providers have access to different Conjur secret | 2 Secrets Provider Jobs run in same namespace | - All relevant K8s secrets are updated<br />- Verify logs    |
+| 4         | Multiple Secrets Providers in same namespace have access to same K8s Secret | Two Secrets Providers have access to same secret             | 2 Secrets Provider Jobs run in same namespace | - No error will returned and both Secrets Providers will be able to access and update K8s Secret <br />- Verify logs |
```
sigalsax, line 481: TODO
> No error will returned 

No error will be returned

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,576 @@
...
+| 5         | Check the integrity of all user input from `values.yaml` / `custom-values.yaml` |                                                              |                                               | - Validate no malicious input was received and escape/encode |
```
sigalsax, line 482: TODO
> and escape/encode

and all values are escaped/encoded

design/milestone_1_2_design_doc.md
```diff
@@ -0,0 +1,576 @@
...
+| 6         | Service Account does not exist                               | `provideExistingServiceAccount=true` but Service Account provided does not exist | Secrets Provider Chart is installed           | Job will fail because Service Account does not exist         |
+| 7         | Add retry check in all regression tests                      |                                                              |                                               | - Verify retry is also written in logs                       |
+| 8         | *Regression tests*                                           | All regression tests should pass (both Conjur / DAP)         |                                               |                                                              |
+| 9         | *Add support for OC 4.3*                                     | All integration tests should pass on OC 4.3                  |                                               | *Not final*                                                  |
+
```
eladkug, line 487: Need to add test for in correct password, and verify the logs

design/batch_retrieval_design.md
```diff
@@ -0,0 +1,20 @@
...
+| 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                                                       | -                     |
+
+(*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
+
+(**) This solution can be broken up in two: 1. Create a new API endpoint or 2. use existing one (return 409 response)
```
sigalsax, line 19: test

design/milestone_1_2_design_doc.md
sigalsax: As we spoke about yesterday testing if an application fails when trying to connect to its target resources is checking k8s functionality and does not really have to do with our secrets provider

sigalsax:
@abrahamko @orenbm looking for approval to merge the solution doc to master (:

orenbm:
No need for an approval from me, the solution design has passed. Merge away @sigalsax 💯

PR #152, opened 2020-07-20
Remove oc3.10 environment

eladkug:
### What does this PR do?
Removed oc3.10 environment for testing, since some of the tests failed which looks like oc3.10

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,9 +51,10 @@ pipeline {
             tasks["Openshift v3.11, oss"] = {
               sh "./bin/test_integration --docker --oss --oc311"
             }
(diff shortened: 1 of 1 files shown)


orenbm:
@eladkug are we planning to add these tests again in the future? aren't we supporting 3.10 anymore?

garymoon:
@orenbm There is a bug in OC 3.10 that is causing some tests to fail incorrectly, and they're being disabled until we can decide how to approach the problem. Infra's recommendation long-term is that we disable only the problematic tests, and only on OC 3.10, until/unless a better alternative presents itself.

PR #165, opened 2020-07-29
Add Helm Chart for Job deployment

sigalsax:
Adds Helm Charts and dummy applications and secrets to get end-to-end

Instructions for how I got end to end

1. Use `k8s-conjur-deploy`, uncomment dev boostrap env and run `source dev-bootstrap.env`

2. Use the dev env, uncomment dev boostrap and run `./bin/start --dap --dev`

3. In run_with_summon of secrets provider repo and check out add-dev-env branch and repo comment out everything after
```
if [ "${DEV}" = "false"  ]; then
  pushd ./test/test_cases > /dev/null
    ./run_tests.sh
  popd > /dev/null
else
  ./dev/5_load_environment.sh
fi
```
because you'll need the policies that the Secrets Provider loads (for authn, hosts, secrets etc)

4. Inside your conjur namespace inside conjur cli add: `conjur variable values add secrets/test_secret supersecret`

5. `Run: kubectl exec <conjur master> --namespace <conjur-namespace> cat /opt/conjur/etc/ssl/conjur.pem > conjur.pem` to get the SSL Cert and save it in a file. You will need the pem to set the config map.

6. `helm install -f custom-values.yaml secrets-provider ./secrets-provider-test-chart --set-file conjurSslCertificateValue=conjur.pem`
Pass in custom-values and setting the SSL certificate upon installing the chart

7. Drop down into dummy application and verify K8s secret is in env
```
[root@test-env-/]# env

TEST_SECRET=supersecret
```


custom-values.yaml
```diff
@@ -0,0 +1,6 @@
+secretsProviderEnvironment:
```
abrahamko, line 1: This file should be in some test folder, judging by the content. No?

custom-values.yaml
```diff
@@ -0,0 +1,6 @@
...
+  conjurAccount: cucumber
+  conjurApplianceUrl: https://conjur-follower.conjur-d284.svc.cluster.local
+  conjurAuthnUrl: https://conjur-follower.conjur-d284.svc.cluster.local/authn-k8s/authn-dev-env
+  conjurAuthnLogin: host/conjur/authn-k8s/authn-dev-env/apps/local-secrets-provider/*/*
+  k8sSecrets: test-k8s-secret
```
abrahamko, line 6: This should be an array

custom-values.yaml
```diff
@@ -0,0 +1,6 @@
+secretsProviderEnvironment:
+  conjurAccount: cucumber
+  conjurApplianceUrl: https://conjur-follower.conjur-d284.svc.cluster.local
+  conjurAuthnUrl: https://conjur-follower.conjur-d284.svc.cluster.local/authn-k8s/authn-dev-env
+  conjurAuthnLogin: host/conjur/authn-k8s/authn-dev-env/apps/local-secrets-provider/*/*
```
abrahamko, line 5: What are these URLs?
Are they const?

secrets-provider-test-chart/.helmignore
```diff
@@ -0,0 +1,22 @@
+# Patterns to ignore when building packages.
```
abrahamko, line 1: since you do not push package in this PR, what this file means?

secrets-provider-test-chart/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
+name: secrets-provider-test-chart
```
abrahamko, line 4: Why is it called test-chart?
Why is the folder called test-chart?

secrets-provider-test-chart/Chart.yaml
```diff
@@ -0,0 +1,5 @@
...
+version: 1.0.0
```
abrahamko, line 5: We might want to add `kubeVersion` also: https://helm.sh/docs/topics/charts/#the-kubeversion-field

abrahamko, line 5: this is the `secrets-provider` version; should be `1.0`, no?

abrahamko: Actually, this is the version of the chart; `appVersion` is the version of the `secrets provider`.
But it should be the same, I think.

abrahamko: We can also add `home` with `secrets provider` github repo url.

secrets-provider-test-chart/templates/app.yaml
```diff
@@ -0,0 +1,29 @@
+# Not to be shipped. Created for end-to-end testing
```
abrahamko, line 1: If not to be shipped, why it's in the chart?

secrets-provider-test-chart/templates/conjur-cert-config-map.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.secretsProviderEnvironment.conjurSslCertificateName }}
```
abrahamko, line 4: missing namespace

custom-values.yaml
```diff
@@ -0,0 +1,6 @@
+secretsProviderEnvironment:
+  conjurAccount: cucumber
```
abrahamko, line 2: Maybe group all `conjur` prefixed values to `conjur` group?

secrets-provider-test-chart/templates/conjur-cert-config-map.yaml
```diff
@@ -0,0 +1,6 @@
...
+data:
+  ssl-certificate: {{ .Values.conjurSslCertificateValue | quote }}
```
abrahamko, line 6: why `conjurSslCertificateValue` is in the open?
Consider adding `conjur.SslCertificate.Value` and `conjur.SslCertificate.Name`

secrets-provider-test-chart/templates/k8s_secret.yaml
```diff
@@ -0,0 +1,9 @@
+# Not to be shipped. Created for end-to-end testing
```
abrahamko, line 1: remove from chart if it is not part of it.

secrets-provider-test-chart/templates/job.yaml
```diff
@@ -0,0 +1,66 @@
...
+      serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
+      containers:
+      - image: {{ .Values.secretsProvider.image }}
+        imagePullPolicy: Never
+        name: {{ .Values.secretsProvider.name }}
+        env:
+        - name: SECRETS_DESTINATION
+          value: {{ .Values.secretsProviderEnvironment.secretsDestination | quote }}
+
+        # TODO, remove requirement in code
```
abrahamko, line 18: Remove this TODO.
We did not plan to do it, so better keep the code as is.

secrets-provider-test-chart/templates/job.yaml
```diff
@@ -0,0 +1,66 @@
...
+              fieldPath: metadata.name
+
+        - name: MY_POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.namespace
+
+        - name: CONJUR_APPLIANCE_URL
+          value:  {{ required "conjurApplianceUrl is a required configuration" .Values.secretsProviderEnvironment.conjurApplianceUrl | quote }}
```
abrahamko, line 35: Remove all the `required`.
We will do the validation using `JsonSchema` in other task.

secrets-provider-test-chart/templates/job.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+        - name: K8S_SECRETS
+          value: {{ required "k8sSecrets is a required configuration"  .Values.secretsProviderEnvironment.k8sSecrets }}
+
+        {{- if .Values.secretsProvider.debug }}
+        - name: DEBUG
+          value: "true"
+        {{- end }}
+      restartPolicy: Never
+  backoffLimit: 0
```
abrahamko, line 66: These two lines are very hard to read, since they are unindented related to the lines before them.
It's not obvious where they belong, better keep them with lines in the same indentation.

secrets-provider-test-chart/templates/role-binding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Values.global.namespace }}
+  name: {{ .Values.rbac.roleBinding }}
```
abrahamko, line 5: Did you mean `roleBindingName`?

secrets-provider-test-chart/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+{{- if eq .Values.rbac.serviceAccount.create true -}}
```
abrahamko, line 1: I think we shouldn't create and bind the role if we don't create the service account.
So this ccheck should be in both `role` and `rolebinding`.

secrets-provider-test-chart/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.rbac.serviceAccount.name }}
```
abrahamko, line 5: missing `namespace`

secrets-provider-test-chart/templates/role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Values.rbac.roleName }}
```
abrahamko, line 4: missing `namespace`

secrets-provider-test-chart/templates/job.yaml
```diff
@@ -0,0 +1,66 @@
...
+              fieldPath: metadata.namespace
+
+        - name: CONJUR_APPLIANCE_URL
+          value:  {{ required "conjurApplianceUrl is a required configuration" .Values.secretsProviderEnvironment.conjurApplianceUrl | quote }}
+
+        - name: CONJUR_AUTHN_URL
+          value:  {{ required "conjurAuthnUrl is a required configuration"  .Values.secretsProviderEnvironment.conjurAuthnUrl | quote }}
+
+        - name: CONJUR_ACCOUNT
+          value:  {{ required "serviceAccount is a required configuration"  .Values.rbac.serviceAccount.name | quote }}
```
abrahamko, line 41: should be `conjurAccount`, not `serviceAccount`

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
+global:
+  namespace: local-secrets-provider
```
abrahamko, line 2: values should be valid defaults for customers to use.
I would remove the `local-` prefix.

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# K8s permission defaults
+rbac:
+  roleName: secrets-provider-role
+  roleBinding: secrets-provider-role-binding
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: true
+    # The name of the service account to use.
+    # If not set and create is true, a name is generated using the fullname template
```
abrahamko, line 12: I don't understand this comment.
Also I think this is incorrect behavior, we shouldn't generate this name.
We should either use a default, or get one from `custom-values.yaml`.

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+    name: cucumber
```
abrahamko, line 13: Bad default name.

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+secretsProvider:
+  image: cyberark/secrets-provider-for-k8s
```
abrahamko, line 16: image missing label `:1.0`

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  tag: 1.0.0
+  name: cyberark-secrets-provider
+
+secretsProviderEnvironment:
+  conjurSslCertificateValue: ""
```
abrahamko, line 21: Is this a valid value?

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  conjurSslCertificateName: secrets-provider-ssl-config-map
+  secretsDestination: k8s_secrets
```
abrahamko, line 23: What other options we have here?
Since we currently do not support any other than `k8s_secrets`, we shouldn't supply it as variable.

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  debug: false
```
abrahamko, line 24: Remove debug from `valus.yaml`

secrets-provider-test-chart/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  supportRotation: false
```
abrahamko, line 25: Better off without it for now.
We can add it later with default.

secrets-provider-test-chart/templates/job.yaml
sigalsax: This means this will need to be in manifest which is no good. We should have 
```
- name: CONTAINER_MODE
  value: init
```
in our manifests if we aren't running as an init

sigalsax: added task: https://app.zenhub.com/workspaces/palmtree-5d99d900491c060001c85cba/issues/cyberark/secrets-provider-for-k8s/166

secrets-provider-test-chart/templates/app.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: test-env
```
oburstein-hub, line 6: should be renamed from test-env to secret-provider

secrets-provider-test-chart/templates/app.yaml
```diff
@@ -0,0 +1,29 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: test-env
+  template:
+    metadata:
+      labels:
+        app: test-env
+    spec:
+      serviceAccountName: cucumber
```
oburstein-hub, line 18: secviceAccountName should be token from Values

secrets-provider-test-chart/templates/app.yaml
```diff
@@ -0,0 +1,29 @@
...
+      containers:
+      - image: centos:7
+        name: test-app
+        command: ["sleep"]
+        args: ["infinity"]
+        env:
```
oburstein-hub, line 24: what is this sleep infinity for ?

secrets-provider-test-chart/templates/app.yaml
```diff
@@ -0,0 +1,29 @@
...
+          - name: TEST_SECRET
+            valueFrom:
+              secretKeyRef:
+                name: test-k8s-secret
```
oburstein-hub, line 28: please, remove "test" prefix everywhere

secrets-provider-test-chart/templates/conjur-cert-config-map.yaml
oburstein-hub: not sure namespace is needed

secrets-provider-test-chart/Chart.yaml
sigalsax: @eladkug which version of k8s are we supporting? (not oc, k8s)

secrets-provider-test-chart/templates/job.yaml
oburstein-hub: disgree @abrahamko

secrets-provider-test-chart/templates/job.yaml
oburstein-hub: right

secrets-provider-test-chart/values.yaml
oburstein-hub: @sigalsax - are you sure that "create" is valid attribute here ?

secrets-provider-test-chart/values.yaml
oburstein-hub: yes - it is

sigalsax: I will check this one to see if I can move it custom-values

secrets-provider-test-chart/templates/app.yaml
sigalsax: this is a dummy app created for testing purposes. I am removing this (:

secrets-provider-chart/templates/role-binding.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if eq .Values.rbac.serviceAccount.create true -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Values.global.namespace }}
```
oburstein-hub, line 5: Please use .Release.Namespace instead of .Values.global.namespace in all places.
Then you will not need to pass namespace as a parameter - it will be automatically taken from namespace in which you run helm install

secrets-provider-chart/templates/secrets-provider.yaml
```diff
@@ -0,0 +1,62 @@
...
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Values.global.namespace }}
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
+      containers:
+      - image: {{ .Values.secretsProvider.image }}:{{ .Values.secretsProvider.tag }}
+        imagePullPolicy: Always
```
oburstein-hub, line 12: replace 'Always' to 'IfNotPresent'

sigalsax: This will be delivered to the customer @oburstein-hub so I would imagine they wouldn't have the image locally no?

oburstein-hub: This flag means that the image should be pulled to K8S from docker registry on every pod run (Always), or only when not pulled yet (IfNotExists)

sigalsax: > only when not pulled yet

from first run?

oburstein-hub: yes, from some previous run

secrets-provider-chart/ci/test-values-template.yaml
```diff
@@ -0,0 +1,25 @@
+# K8s permission defaults
+rbac:
+  roleName: {{ SECRETS_PROVIDER_ROLE }}
+  roleBindingName: {{ SECRETS_PROVIDER_ROLE_BINDING }}
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: {{ SERVICE_ACCOUNT_CREATE }}
```
orenbm, line 7: i would use `CREATE_SERVICE_ACCOUNT`

secrets-provider-chart/ci/test-values-template.yaml
```diff
@@ -0,0 +1,25 @@
...
+    # The name of the service account to use
+    name: {{ SERVICE_ACCOUNT }}
+
+secretsProvider:
+  image: cyberark/secrets-provider-for-k8s
+  tag: 1.0.0
```
orenbm, line 13: what's this tag? is this the tag of the image? do we want it static?

secrets-provider-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
+{{/*
+Return the most recent RBAC API available
+*/}}
+{{- define "conjur.rbac-api" -}}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
+{{- printf "rbac.authorization.k8s.io/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "v1" }}
+{{- printf "v1" -}}
+{{- end }}
+{{- end }}
```
orenbm, line 10: i am not familiar with helm. should this be indented?

secrets-provider-chart/templates/secrets-provider.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - name: CONJUR_SSL_CERTIFICATE
+          valueFrom:
+            configMapKeyRef:
+              name: {{ .Values.environment.conjur.sslCertificate.name | quote }}
+              key: ssl-certificate
+
+        - name: CONJUR_AUTHN_LOGIN
+          value:  {{ .Values.environment.conjur.authnLogin | quote }}
+
+        - name: SUPPORT_MULTIPLE_APPS
```
orenbm, line 51: what's the purpose of this var?

secrets-provider-chart/values.yaml
```diff
@@ -0,0 +1,28 @@
...
+  roleBindingName: secrets-provider-role-binding
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: true
+    # The name of the service account to use.
+    name: secrets-provider-service-account
+
+secretsProvider:
+  image: cyberark/secrets-provider-for-k8s
+  tag: 1.0.0
```
orenbm, line 13: do we want this static?

secrets-provider-chart/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+appVersion: "1.0.0"
+description: A Helm chart for Kubernetes
```
InbalZilberman, line 3: sound too generic for me. what about A helm chart for deploying CyberArk Secrets Provider for Kubernetes

secrets-provider-chart/Chart.yaml
```diff
@@ -0,0 +1,6 @@
...
+name: secrets-provider-chart
```
InbalZilberman, line 4: Also here the name should be the official name of the product CyberArk Secrets Provider for Kubernetes

custom-values.yaml
```diff
@@ -0,0 +1,9 @@
+environment:
```
InbalZilberman, line 1: @sigalsax can we add a remark here above each param?

secrets-provider-chart/ci/test-values-template.yaml
sigalsax: this tag is something that for each release of a helm chart we would update and ship off to the customer. So I would imagine for Rotation Milestone the version here will be 2.0.0.

Now that I think about it, this sort of limits us from releasing a new secrets provider major.minor.patch version without delivering a new Helm to the customer.. What do you think?

secrets-provider-chart/templates/_helpers.tpl
sigalsax: She Helm's GH repo [here](https://github.com/helm/charts/blob/b05e749688f2c53f76cc5cdcdc06c02acdf24aaf/stable/telegraf/templates/_helpers.tpl#L14)
Only very specific things are indented (range, etc)
I would imagine because every indent is shown once rendered and yaml is indent sensitive

secrets-provider-chart/values.yaml
sigalsax: see my above response to the previous comment

secrets-provider-chart/templates/secrets-provider.yaml
sigalsax: this was supposed to be a distinguishable of the job much like we have for `container_mode = init` so I believe so.

@abrahamko what did we decide on this?

secrets-provider-chart/Chart.yaml
sigalsax: whoops good batch you're absolutely right

custom-values.yaml
sigalsax: I am deleting this file from our chart because this is not something we will be supplying for the user. they will supply us with their own `custom-values.yaml` as part of the install process. We have duplicated values in values.yaml so I will add that explanation there

secrets-provider-chart/Chart.yaml
sigalsax: updating the chart name to `cyberark-secrets-provider-for-kubernetes`

sigalsax: this is the name they will use to install
`helm install cyberark-secrets-provider-for-kubernetes ...`

secrets-provider-chart/templates/secrets-provider.yaml
sigalsax: We need to add a var to distinguish Rotation too so we can already implement it in the code. @abrahamko 
For example: `SUPPORT_ROTATION`

secrets-provider-chart/ci/test-values-template.yaml
oburstein-hub: I suggest to put both image and tag field contents to values.yaml - see example in helm chart I have provided. This way of the user wants he can control them. The defaults of these field in values.yaml should be as you said Sigal

cyberark-secrets-provider-for-kubernetes/values.yaml
```diff
@@ -0,0 +1,36 @@
+# K8s permission defaults
```
sigalsax, line 1: @InbalZilberman see below comments. I would like Shuli to take a look at the comments here as well. I will open a card

custom-values.yaml
sigalsax: added card for shuli and tagged you: https://github.com/cyberark/conjur-docs/issues/1053

secrets-provider-chart/ci/test-values-template.yaml
sigalsax: It already exists there 
```
secretsProvider:
  image: cyberark/secrets-provider-for-k8s
  tag: 1.0.0
  name: cyberark-secrets-provider
```
The ci folder is for our pipeline to run our tests

cyberark-secrets-provider-for-kubernetes/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+appVersion: "1.0.0"
+description: A Helm chart for Kubernetes
+name: cyberark-secrets-provider-for-kubernetes
```
sigalsax, line 4: this will also be the name of the container. We can change this to `cyberark-secrets-provider` as we have it for the init container. what do you think? @orenbm @InbalZilberman 
Screen Shot 2020-08-03 at 9 51 38 AM

cyberark-secrets-provider-for-kubernetes/ci/test-values-template.yaml
```diff
@@ -0,0 +1,25 @@
+# K8s permission defaults
```
abrahamko, line 1: Remove from this PR.
this should be part of the integration tests PR.

cyberark-secrets-provider-for-kubernetes/ci/test-values-template.yaml
```diff
@@ -0,0 +1,25 @@
...
+  serviceAccount:
+    # Specifies whether a service account should be created
+    create: {{ CREATE_SERVICE_ACCOUNT }}
+    # The name of the service account to use
+    name: {{ SERVICE_ACCOUNT }}
+
+secretsProvider:
+  image: cyberark/secrets-provider-for-k8s
+  tag: 1.0.0
+  name: cyberark-secrets-provider
```
abrahamko, line 14: No need to put these if you don't override the defaults.

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
...
+{{- printf "rbac.authorization.k8s.io/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "v1" }}
+{{- printf "v1" -}}
+{{- end }}
+{{- end }}
+
+{{- define "conjur.deploy-api" -}}
+{{- if .Capabilities.APIVersions.Has "batch/v1" }}
+{{- printf "batch/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "apps/v1" }}
```
abrahamko, line 15: Did you check this?
It looks weird to use `apps/v1` for jobs

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
...
+{{- define "conjur.rbac-api" -}}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
+{{- printf "rbac.authorization.k8s.io/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "v1" }}
+{{- printf "v1" -}}
+{{- end }}
+{{- end }}
+
+{{- define "conjur.deploy-api" -}}
+{{- if .Capabilities.APIVersions.Has "batch/v1" }}
```
abrahamko, line 13: If this is for `Job`, I think it is better to check specifically for it.
Like this: 
```suggestion
{{- if .Capabilities.APIVersions.Has "batch/v1/Job" }}
```

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
...
+*/}}
+{{- define "conjur.rbac-api" -}}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
+{{- printf "rbac.authorization.k8s.io/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "v1" }}
+{{- printf "v1" -}}
+{{- end }}
+{{- end }}
+
+{{- define "conjur.deploy-api" -}}
```
abrahamko, line 12: Add comment like the one above the RBAC to explain what this func is for.

secrets-provider-chart/templates/secrets-provider.yaml
orenbm: i don't understand what we will do differently when `SUPPORT_MULTIPLE_APPS` is true and when it's false. can you elaborate please?

cyberark-secrets-provider-for-kubernetes/Chart.yaml
orenbm: i would change `cyberark-secrets-provider` to `cyberark-secrets-provider-for-kubernetes`

orenbm: unless it breaks compatibility in a way that i am missing

sigalsax: I can just hard code that in the manifest

sigalsax: @InbalZilberman  can we call the chart `cyberark-secrets-provider-for-k8s` to save on line space and considering it will sit in the `secrets-provider-for-k8s` repo

InbalZilberman: @sigalsax yes :)

cyberark-secrets-provider-for-kubernetes/values.yaml
InbalZilberman: Thanks

cyberark-secrets-provider-for-kubernetes/Chart.yaml
rafis3: Why is the part `for-kubernetes` needed? Doesn't it go without saying that a Helm chart and a running pod are Kubernetes related? I understand that this is the official name from GitHub, but in GitHub it makes sense because here you don't have the Kubernetes context as you do when it's deployed

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
rafis3: @abrahamko are you sure it's applicable? The API version of a Job is still `batch/v1`.
Please see this [reference](https://www.k8sref.io/docs/workloads/job-v1/).

cyberark-secrets-provider-for-kubernetes/templates/role-binding.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if eq .Values.rbac.serviceAccount.create true -}}
```
rafis3, line 1: I think this should work without specifying `true` because it's a boolean condition

cyberark-secrets-provider-for-kubernetes/templates/role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: {{ include "conjur.rbac-api" . }}
+kind: Role
+metadata:
+  name: {{ .Values.rbac.roleName }}
+  namespace: {{ .Release.namespace }}
```
rafis3, line 5: Does the secrets provider only update secrets in the same namespace in which it runs?

cyberark-secrets-provider-for-kubernetes/templates/secrets-provider.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: {{ include "conjur.deploy-api" . }}
+kind: Job
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.namespace }}
```
rafis3, line 5: I would add an option to add labels or annotations on both the Job definition and the template of the pod.
It's really helpful sometimes for the end user to be able to control certain behaviours of the Job and pod, using labels and annotations. For example, a user could set rules in his/her cluster that pods with certain labels get a certain resource quota.

cyberark-secrets-provider-for-kubernetes/templates/secrets-provider.yaml
```diff
@@ -0,0 +1,62 @@
...
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ .Values.rbac.serviceAccount.name }}
+      containers:
+      - image: {{ .Values.secretsProvider.image }}:{{ .Values.secretsProvider.tag }}
+        imagePullPolicy: IfNotPresent
```
rafis3, line 12: I would make all `imagePullPolicy` configurable. It's a highly useful parameter in my opinion

cyberark-secrets-provider-for-kubernetes/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+environment:
+  # Comma-separated list of Kubernetes Secret names that applications consume and their value is sourced in DAP/Conjur
+  # This key should be in array format. For example, [k8s-secret1,k8s-secret2]
+  k8sSecrets:
+  conjur:
+    # Account name designated to the Conjur/DAP during initial configuration
+    account:
+    # URL of service defined for Conjur/DAP
+    applianceUrl:
+    # Credential service used to log into Conjur/DAP
```
rafis3, line 27: What do you think about adding an example in the comments? I know that we have JSON schema validation, but would be nice for a user to know how does an expected input look like, before it fails in the validation.

cyberark-secrets-provider-for-kubernetes/Chart.yaml
sgnn7: outside of the conversation about which is better to use, we should keep in mind that most k8s resources have a _really_ low maximum length (32 or 64 chars) so brevity is always preferred

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
+{{/*
+Return the most recent RBAC API available
+*/}}
+{{- define "conjur.rbac-api" -}}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
+{{- printf "rbac.authorization.k8s.io/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "v1" }}
+{{- printf "v1" -}}
```
sgnn7, line 8: Since you use the `{{- blah -}}` pattern everywhere you can use regular indentation to make the logic blocks not looks so cluttered

cyberark-secrets-provider-for-kubernetes/templates/secrets-provider.yaml
sgnn7: I'd strongly +1 this

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
sigalsax: batch.v1 is for jobs

secrets-provider-chart/templates/secrets-provider.yaml
sigalsax: SUPPORT_MULTIPLE_APPS=true is the separately deployment Job that supports multiple apps
Ideally we wouldn't be expecting this value to be false for now because we have CONTAINER_MODE init for that

cyberark-secrets-provider-for-kubernetes/Chart.yaml
sigalsax: @InbalZilberman  ^ please see above thread

cyberark-secrets-provider-for-kubernetes/templates/role.yaml
sigalsax: Yes indeed (:

secrets-provider-chart/templates/secrets-provider.yaml
oburstein-hub: @sigalsax it is not clear to me for what we need this variable. Please, explain what is the logic that it adds

cyberark-secrets-provider-for-kubernetes/templates/secrets-provider.yaml
sigalsax: Okay, I will add 
`imagePullPolicy: IfNotPresent` in the values.yaml and the user will be able to override this one via `custom-values.yaml`

cyberark-secrets-provider-for-kubernetes/Chart.yaml
oburstein-hub: I would suggest to call it secret-provider-for-k8s, because there is no convention of adding cyberark prefix in other things

cyberark-secrets-provider-for-kubernetes/values.yaml
sigalsax: We also plan to write documentation with examples. Do you think that that will be sufficient or do you still feel examples in the comments give a better experience for the user?

oburstein-hub: I think documentation is better, because values.yaml arrives to the customer in many cases inside hel.tgz file and hence the customer will not see it

rafis3: I disagree. The `values.yaml` is the file that the end user edits to customize the Helm chart values for his/her setup. If the user will look at any file in the Helm chart, this would be the one.

secrets-provider-chart/templates/secrets-provider.yaml
sigalsax: For init deployment we have CONTAINER_MODE=init, now we cannot use that anymore because we are no longer init container but we still need to identity the type of deployment so we thought `SUPPORT_MULTIPLE_APPS` is most representative of deployment _flow_

cyberark-secrets-provider-for-kubernetes/ci/test-values-template.yaml
sigalsax: @abrahamko don't understand your comment here. Why?

cyberark-secrets-provider-for-kubernetes/ci/test-values-template.yaml
sigalsax: We might if we want to test different versions

cyberark-secrets-provider-for-kubernetes/values.yaml
oburstein-hub: this file is called usually custom-values.yaml @rafis3 
The file that is called values.yaml is part of helmchart itself.

secrets-provider-chart/templates/secrets-provider.yaml
oburstein-hub: so it is going to be instead of the variable CONTAINER_MODE ?

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
sigalsax: I think the above comment also applies here. Or do you have an idea?
```
{{/*
Return the most recent Deployment Configuration API available
*/}}
```
?

cyberark-secrets-provider-for-kubernetes/values.yaml
oburstein-hub: following to our offline conversation with Rafi, we advice that the required format of input will be displayed as an output of helm install. Maybe it os done automatically by json schema mechanism. @sigalsax please check it - and if it is not - it must be added explicitly to the output

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
sigalsax: I wanted to follow the[ helm/chart](https://github.com/helm/charts/blob/886cac277eaaada4ac2c1495c81a11f025c588b9/incubator/solr/templates/_helpers.tpl) pattern as they have it

cyberark-secrets-provider-for-kubernetes/templates/secrets-provider.yaml
sigalsax: You got it!

secrets-provider-chart/templates/secrets-provider.yaml
sigalsax: we aren't deleting container_mode for backwards compatibility

cyberark-secrets-provider-for-kubernetes/values.schema.json
```diff
@@ -0,0 +1,126 @@
...
+              ],
+              "properties": {
+                "name": {
+                  "type": "string",
+                  "minLength": 1
+                },
+                "value": {
+                  "type": "string",
+                  "minLength": 1
+                }
```
sigalsax, line 118: @abrahamko @eladkug we should also check for maxlength

cyberark-secrets-provider-for-kubernetes/values.schema.json
```diff
@@ -0,0 +1,126 @@
...
+  "$schema": "http://json-schema.org/schema",
+  "required": [
+    "rbac",
+    "secretsProvider",
+    "environment"
+  ],
+  "properties": {
+    "rbac": {
+      "required": [
+        "roleName",
```
sigalsax, line 11: @abrahamko @eladkug we should check min and max length of these field for k8s. Is 1 character enough to be valid?

cyberark-secrets-provider-for-kubernetes/values.schema.json
```diff
@@ -0,0 +1,126 @@
...
+        "roleBindingName",
+        "serviceAccount",
+        "create"
```
sigalsax, line 14: is there a way to limit input "true" or "false"

cyberark-secrets-provider-for-kubernetes/values.schema.json
sigalsax: Will a space (ascii, etc) character be a valid service/role name?

secrets-provider/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+    name: secrets-provider-service-account
+
+secretsProvider:
+  image: cyberark/secrets-provider-for-k8s
+  pullPolicy: IfNotPresent
+  tag: 1.0.0
+  name: cyberark-secrets-provider
+
+# Additional labels to apply to all resources
+labels: {}
```
sigalsax, line 22: @abrahamko can you add json schema for labels and annotations? 🙏

secrets-provider/templates/secrets-provider.yaml
```diff
@@ -75,5 +75,15 @@ spec:
         - name: DEBUG
           value: "true"
         {{- end }}
+
+        {{- if .Values.environment.syncIntervalTime }}
```
sigalsax, line 79: @abrahamko can you add json schema for syncIntervalTime and retryCount? 🙏

secrets-provider/templates/secrets-provider.yaml
```diff
@@ -75,5 +75,15 @@ spec:
...
+        - name: SYNC_INTERVAL_TIME
+          value: {{ .Values.environment.syncIntervalTime }}
```
sigalsax, line 81: @oburstein-hub @eladkug what would be the mainstream flow for using SyncIntervalTime and RetryCount? 
1️⃣ What should the details be in code if the user does not provide us with this?
2️⃣ What if the user only provides `syncIntervalTime` and not `retryCount`? Are they a package deal or would the normal flow be taking the configured syncIntervalTime and use the default retryCount if not provided also?

cyberark-secrets-provider-for-kubernetes/values.schema.json
```diff
@@ -0,0 +1,126 @@
...
+      "required": [
+        "k8sSecrets",
+        "conjur"
+      ],
+      "properties": {
+        "k8sSecrets": {
+          "type": "array",
+          "minLength": 1,
+          "items": {
+            "type": "string",
```
sigalsax, line 73: @abrahamko is this enough? do we also want to escape the string?

secrets-provider/templates/secrets-provider.yaml
oburstein-hub: 1. The defaults should be as follows: SyncIntervalTime: 3 seconds, retryCount: 3 - @InbalZilberman - please comment
2. User can provide each of above parameter separately. The unprovided parameter will stay with default value

sigalsax: to clarify, these params are for our retry mechanism if there was a failure (connecting to follower, fetching conjur secret, updating k8s secret, etc)

sigalsax: @oburstein-hub  great thank you! Want to note that these default will **not** be forward facing rather be in the code itself

secrets-provider/templates/secrets-provider.yaml
sigalsax: with validation of `> 0`. In other words, cannot be 0 or less

cyberark-secrets-provider-for-kubernetes/values.schema.json
abrahamko: I disagree.
Let k8s do its own validations, I think we shouldn't do validations on behalf of k8s.

secrets-provider/templates/secrets-provider.yaml
```diff
@@ -48,15 +65,12 @@ spec:
         - name: CONJUR_AUTHN_LOGIN
           value:  {{ .Values.environment.conjur.authnLogin | quote }}
 
-        - name: SUPPORT_MULTIPLE_APPS
```
sigalsax: removing in order to avoid introducing env variables without the code. Will bring this back in when we make this addition in the code

secrets-provider/ci/test-values-template.yaml
```diff
@@ -0,0 +1,34 @@
+# Values for secrets-provider Chart. All missing values need to be supplied via environment variable.
```
abrahamko, line 1: I think it should be part of integration tests.
There's no context in this PR to review this.

secrets-provider/templates/_helpers.tpl
```diff
@@ -0,0 +1,21 @@
...
+{{- define "conjur.rbac-api" -}}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
+{{- printf "rbac.authorization.k8s.io/v1" -}}
+{{- else if .Capabilities.APIVersions.Has "v1" }}
+{{- printf "v1" -}}
+{{- end }}
+{{- end }}
+
+{{/*
+Return the most recent Workload API available
```
abrahamko, line 13: This is dedicated to Job, so be explicit in the comment and in the name of the function

secrets-provider/templates/role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: {{ include "conjur.rbac-api" . }}
```
abrahamko, line 1: Missing `{{- if .Values.rbac.create -}}`

secrets-provider/templates/_helpers.tpl
sigalsax: This isn't dedicated to Job. This is for deployment and Job.
If Deployment it will take apps/v1, if Job, it will take batch/v1

cyberark-secrets-provider-for-kubernetes/templates/_helpers.tpl
abrahamko: @rafis3 Yes, I'm sure, you can see in `helm` docs here: https://helm.sh/docs/chart_template_guide/builtin_objects/
`
Capabilities.APIVersions.Has $version indicates whether a version (e.g., batch/v1) or resource (e.g., apps/v1/Deployment) is available on the cluster.
`

rafis3: Cool, thanks for sharing!

secrets-provider/values.schema.json
```diff
@@ -0,0 +1,137 @@
...
+          "type": "string",
+          "minLength": 1
+        }
+      }
+    },
+    "labels": {
+      "type": "object"
+    },
+    "annotations": {
+      "type": "object"
```
sigalsax, line 72: @abrahamko Should we also caste to strings?

sigalsax: Seems they are interpreted as strings https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

secrets-provider/templates/role.yaml
sigalsax: 😲 you are right!

secrets-provider/templates/_helpers.tpl
sigalsax: I am deleted this because we don't yet need this. this would be needed for rotation to switch back and forth

secrets-provider/values.schema.json
```diff
@@ -0,0 +1,137 @@
...
+          "minLength": 1
+        },
+        "tag": {
+          "type": ["string", "number"],
+          "minLength": 1,
+          "pattern": "^\\d+(\\.\\d+){0,2}$"
+        },
+        "pullPolicy": {
+          "type": "string",
+          "enum": ["Always", "IfNotPresent", "Never"]
```
sigalsax, line 60: nice

secrets-provider/values.schema.json
```diff
@@ -0,0 +1,137 @@
...
+        "image": {
+          "type": "string",
+          "minLength": 1
+        },
+        "tag": {
+          "type": ["string", "number"],
+          "minLength": 1,
+          "pattern": "^\\d+(\\.\\d+){0,2}$"
+        },
+        "pullPolicy": {
```
rafis3, line 58: Since in K8s, the parameter name is `imagePullPolicy`, I think it would be nice to keep the same name to provide a known language the the user already familiar with.

sigalsax: you got it!

cyberark-secrets-provider-for-kubernetes/values.schema.json
oburstein-hub: I think it is done automatically without need to code

secrets-provider/templates/secrets-provider.yaml
oburstein-hub: good idea

PR #177, opened 2020-08-10
Add logs to codebase

sigalsax:
These logs are added as part of Milestone Job. Based off conversations with Shuli, we are adding new logs and correcting old ones

See docs card here: https://github.com/cyberark/conjur-docs/issues/1046

Some changes made after discussion with TW:
k8s secret -> Kubernetes Secret
Conjur -> DAP/Conjur
Kubernetes Secrets Provider -> Secrets Provider for Kubernetes

** Retry mechanism log will be added when that functionality has been implemented


cmd/secrets-provider/main.go
```diff
@@ -63,40 +63,28 @@ func main() {
...
+		err = provideConjurSecrets(accessToken)
+		if err != nil {
+			return log.RecordedError(messages.CSPFK016E)
+		}
+
+		err = accessToken.Delete()
+		if err != nil {
+			return log.RecordedError(messages.CSPFK003E, err.Error())
 		}
+		log.Info(messages.CSPFK009I)
```
abrahamko, line 86:
```suggestion

		log.Info(messages.CSPFK009I)
```
Add new line

pkg/log/messages/info_messages.go
```diff
@@ -13,10 +13,12 @@ package messages
...
-const CSPFK002I string = "CSPFK002I Creating Conjur client..."
-const CSPFK003I string = "CSPFK003I Retrieving following secrets from Conjur: %v"
-const CSPFK004I string = "CSPFK004I Creating Kubernetes client..."
-const CSPFK005I string = "CSPFK005I Retrieving Kubernetes secret '%s' from namespace '%s'..."
+const CSPFK002I string = "CSPFK002I Creating DAP/Conjur client"
+const CSPFK003I string = "CSPFK003I Retrieving following secrets from DAP/Conjur: %v"
+const CSPFK004I string = "CSPFK004I Creating Kubernetes client"
+const CSPFK005I string = "CSPFK005I Retrieving Kubernetes secret '%s' from namespace '%s'"
 const CSPFK006I string = "CSPFK006I Updating Kubernetes secret '%s' in namespace '%s'"
 const CSPFK007I string = "CSPFK007I Waiting for %s to re-authenticate and fetch secrets."
```
abrahamko, line 21: You deleted this log line in `main()`.
Is it still in use?

pkg/log/messages/info_messages.go
```diff
@@ -13,10 +13,12 @@ package messages
...
-const CSPFK008I string = "CSPFK008I Kubernetes Secrets Provider v%s starting up..."
+const CSPFK008I string = "CSPFK008I CyberArk Secrets Provider for Kubernetes v%s starting up"
+const CSPFK009I string = "CSPFK009I Kubernetes Secrets updated successfully"
```
abrahamko, line 23: `Kubernetes secrets` is only when target is `k8s-secrets`.
Consider:
```suggestion
const CSPFK009I string = "CSPFK009I Conjur secrets updated in Kubernetes successfully"
```

pkg/log/messages/info_messages.go
```diff
@@ -13,10 +13,12 @@ package messages
 */
 
 const CSPFK001I string = "CSPFK001I Authenticating as user '%s'"
-const CSPFK002I string = "CSPFK002I Creating Conjur client..."
-const CSPFK003I string = "CSPFK003I Retrieving following secrets from Conjur: %v"
-const CSPFK004I string = "CSPFK004I Creating Kubernetes client..."
-const CSPFK005I string = "CSPFK005I Retrieving Kubernetes secret '%s' from namespace '%s'..."
+const CSPFK002I string = "CSPFK002I Creating DAP/Conjur client"
```
orenbm, line 16: is this something that PO requested?

cmd/secrets-provider/main.go
```diff
@@ -63,40 +63,28 @@ func main() {
 	expBackoff.MaxElapsedTime = 2 * time.Minute
 
 	err = backoff.Retry(func() error {
-		for {
```
orenbm, line 66: 1. why did the `for` go away?
2. did anything else change here apart from removing the `for`?

pkg/log/messages/info_messages.go
sigalsax: Yes we needed to find a way to support both flows. 
@InbalZilberman

cmd/secrets-provider/main.go
sigalsax: :one: because we don't need it since we aren't a sidecar and we don't want to restart the process since we are an init container / job.

:two: Yes we removed unreachable code 
```
if authnConfig.ContainerMode == "init" {
  os.Exit(0)
}

// Reset exponential backoff
expBackoff.Reset()

log.Info(messages.CSPFK007I, authn.Config.TokenRefreshTimeout)

fmt.Println()
time.Sleep(authn.Config.TokenRefreshTimeout)
```

sigalsax: unreachable because the process hits, it will exit anyways because it is always ContainerMode init
```
if authnConfig.ContainerMode == "init" {
  os.Exit(0)
}
```

pkg/log/messages/info_messages.go
sigalsax: This was purposeful. I want to leave this in because we may find it valuable in a later milestone

pkg/log/messages/info_messages.go
sigalsax: Good point being that we may look to files in the future

cmd/secrets-provider/main.go
orenbm: well, this PR does a LOT more than just adding logs to the codebase.

if i understand correctly, after this change we don't use the `CONTAINER_MODE` parameter. will we not use it anytime in the future? won't we ever run as a sidecar? Also, we need to remove it from the docs in that case.

seems that `CSPFK007I` is never used.

orenbm: don't forget that the default value for `authnConfig.ContainerMode` is `sidecar`. i know that we are now not querying this value but what will happen if we choose to support sidecar in the future? how will we do so _and_ maintain backwards compatibility?

sigalsax: I think we should still use `CONTAINER_MODE` / have customer's declare it because it allows us to determine which flow we will use init or Job. 

We need it for this check
```
if secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode != "init" {
   printErrorAndExit(messages.CSPFK007E)
}
```
Here is where we will determine which flow the customer defined for us and limit them to either init (ContainerMode) or Job (SupportMultipleApps)

If we choose to move to sidecar deployments then we will add this back in but IMO it makes the code a lot less readable because we have places that are just not reachable.

sigalsax:
@orenbm added it back to the way it was

pkg/log/messages/info_messages.go
orenbm: yuck

InbalZilberman:
All logs reviewed and approved

Issue #166, opened 2020-07-29
Secrets Provider - Update CONTAINER_MODE init restriction


sigalsax:
We are required to list the following envs in our secrets provider manifests
If `SECRETS_DESTINATION = k8s_secrets` we demand that `CONTAINER_MODE` be init. 

```
- name: SECRETS_DESTINATION
  value: k8s_secrets

- name: CONTAINER_MODE
  value: init
```

Error returned:
```
ERROR: 2020/07/29 11:13:10 main.go:113: CSPFK007E Setting SECRETS_DESTINATION environment variable to 'k8s_secrets' must run as init container
```

We must update the following to support also supportingMultipleApps
```
if secretsConfig.StoreType == secretsConfigProvider.K8S && (authnConfig.ContainerMode != "init" || authnConfig.SupportMultipleApps != “true”  {
	printErrorAndExit(messages.CSPFK007E)
}
```
Update `defaultConfig` to check for either `ContainerMode` or `SupportMultipleApps`.

and update the log to:
```
CSPFK007E Setting SECRETS_DESTINATION environment variable to 'k8s_secrets' must run as init container or must support multiple applications”
```

DOD
- [x] Update code to not require `CONTAINER_MODE` init when `SECRETS_DESTINATION` is k8s_Secret. Check if SUPPORT_MULTIPLE_APPS is set as well.
- [x] Remove CONTAINER_MODE restriction in Helm chart and add `SUPPORT_MULTIPLE_APPS`
- [x] Add JSON schema for `SUPPORT_MULTIPLE_APPS`

PR #179, opened 2020-08-12
Support CONTAINER_MODE running as an application

sigalsax:
### What does this PR do?
SECRETS_DESTINATION=k8s_secrets was bound to `CONTAINER_MODE=init`. Because we are expanding and supporting the Job flow (and not running as an init container) `CONTAINER_MODE=init` no longer suits our needs. 

In this PR we are introducing a new env variable to support the Job flow (`CONTAINER_MODE=application`)

Logs were approved by Shuli <3

Connected to #166

See image below for example of what will be returned if user inputs a value we don't support
Screen Shot 2020-08-13 at 11 01 47 AM

### DIFF
--- a/cmd/secrets-provider/main.go
+++ b/cmd/secrets-provider/main.go
@@ -34,10 +34,7 @@ func main() {
 		printErrorAndExit(messages.CSPFK015E)
 	}
 
-	// verify we don't run with a K8s secrets storage configuration in a side car
-	if secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode != "init" {
-		printErrorAndExit(messages.CSPFK007E)
-	}
+	validateContainerMode(authnConfig.ContainerMode)
 
 	provideConjurSecrets, err := secrets.GetProvideConjurSecretFunc(secretsConfig.StoreType)
 	if err != nil {
@@ -85,8 +82,8 @@ func main() {
 				return log.RecordedError(messages.CSPFK003E, err.Error())
 			}
 
-			log.Info(messages.CSPFK009I)
-			if authnConfig.ContainerMode == "init" {
+			if authnConfig.ContainerMode == "init" || authnConfig.ContainerMode == "application" {
+				log.Info(messages.CSPFK010I)
 				os.Exit(0)
 			}
 
@@ -126,5 +123,23 @@ func configureLogLevel() {
 	} else if val != "" {
 		// In case "DEBUG" is configured with incorrect value
 		log.Warn(messages.CSPFK001W, val, validVal)
+	}
+}
+
+func validateContainerMode(containerMode string) {
+	validContainerModes := []string{
+		"init",
+		"application",
+	}
+
+	isValidContainerMode := false
+	for _, validContainerModeType := range validContainerModes {
+		if containerMode == validContainerModeType {
+			isValidContainerMode = true
+		}
+	}
+
+	if !isValidContainerMode {
+		printErrorAndExit(fmt.Sprintf(messages.CSPFK007E, containerMode, validContainerModes))
 	}
 }
--- a/pkg/log/messages/error_messages.go
+++ b/pkg/log/messages/error_messages.go
@@ -20,7 +20,7 @@ const CSPFK003E string = "CSPFK003E AccessToken failed to delete access token da
 // Environment variables
 const CSPFK004E string = "CSPFK004E Environment variable '%s' must be provided"
 const CSPFK005E string = "CSPFK005E Provided incorrect value for environment variable %s"
-const CSPFK007E string = "CSPFK007E Setting SECRETS_DESTINATION environment variable to 'k8s_secrets' must run as init container"
+const CSPFK007E string = "CSPFK007E CONTAINER_MODE '%s' is not supported. Supported values are: %v"
 
 // Authenticator
 const CSPFK008E string = "CSPFK008E Failed to instantiate authenticator configuration"
--- a/helm/secrets-provider/templates/secrets-provider.yaml
+++ b/helm/secrets-provider/templates/secrets-provider.yaml
@@ -33,7 +33,7 @@ spec:
           value: k8s_secrets
 
         - name: CONTAINER_MODE
-          value: init
+          value: application
 
         - name: MY_POD_NAME
           valueFrom:


cmd/secrets-provider/main.go
```diff
@@ -34,8 +34,14 @@ func main() {
 		printErrorAndExit(messages.CSPFK015E)
 	}
 
+	// verify CONTAINER_MODE and SUPPORT_MULTIPLE_APPS are not both set
+	envValMultiApps := os.Getenv("SUPPORT_MULTIPLE_APPS")
+	if authnConfig.ContainerMode != "" && (envValMultiApps != "false" && envValMultiApps != "") {
```
orenbm, line 39: why can't both live together?

cmd/secrets-provider/main.go
```diff
@@ -85,7 +91,7 @@ func main() {
 				return log.RecordedError(messages.CSPFK003E, err.Error())
 			}
 
-			if authnConfig.ContainerMode == "init" {
+			if authnConfig.ContainerMode == "init" || envValMultiApps == "true" {
```
orenbm, line 94: so `SUPPORT_MULTIPLE_APPS` indicates that the secrets-provider is running in its own pod?

orenbm: here's an idea. the `CONTAINER_MODE` variable can get 3 values:
- `init`
- `sidecar`
- `app`

this still makes sense in our purpose as what you want to say is "if the secrets-provider runs as an init container or if it runs as an app container, we should exit the code after one cycle". this way the only code change will be:
```
if authnConfig.ContainerMode == "init" || authnConfig.ContainerMode == "app" {
  os.Exit(0)
}
```

no extra variable that is not really needed and we get a logical solution to our issue.

what say you?

orenbm: From k8s docs:
```
As well as application containers, a Pod can contain init containers that run during Pod startup
```
so we have the following container types:
- application
- init
- sidecar

it all fits

sigalsax: I think it makes it much more clear! Thanks for the input! <3

cmd/secrets-provider/main.go
sigalsax: updated to your idea but to answer you here. container_mode init and support multiple apps = true cannot run together because they are different flows. One is for init, the other a Job

cmd/secrets-provider/main.go
```diff
@@ -35,7 +35,7 @@ func main() {
 	}
 
 	// verify we don't run with a K8s secrets storage configuration in a side car
-	if secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode != "init" {
+	if secretsConfig.StoreType == secretsConfigProvider.K8S && (authnConfig.ContainerMode != "init" && authnConfig.ContainerMode != "application") {
```
orenbm, line 38: do you want to change it to `secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode == "sidecar"`?

it is what is written in the comment above (and we can remove the comment)

pkg/log/messages/error_messages.go
```diff
@@ -20,7 +20,7 @@ const CSPFK003E string = "CSPFK003E AccessToken failed to delete access token da
 // Environment variables
 const CSPFK004E string = "CSPFK004E Environment variable '%s' must be provided"
 const CSPFK005E string = "CSPFK005E Provided incorrect value for environment variable %s"
-const CSPFK007E string = "CSPFK007E Setting SECRETS_DESTINATION environment variable to 'k8s_secrets' must run as init container"
+const CSPFK007E string = "CSPFK007E Setting SECRETS_DESTINATION environment variable to 'k8s_secrets' must run as init or application container"
```
orenbm, line 23: i would change this to `must not run as a sidecar container`

sigalsax: ideally the user can put
```
- name: CONTAINER_MODE
   value: sigal
```
So I want to put feedback on the types of input we do accept

cmd/secrets-provider/main.go
sigalsax: i will remove the comment because I want to be very clear with what types of input we accept

orenbm: and what will happen if we add another container mode in the future? will we add it to this `if` clause too? if we are trying to not support k8s store-type alongside a side-car container mode let's do that.

pkg/log/messages/error_messages.go
orenbm: would love to know what the `sigal` container mode gives you!

but i really think that we should say what we do not allow as that is the issue here

sigalsax: I disagree with you here. Because we don't know if they inputted "sidecar". It could ideally be anything and we are assuming sidecar. IMO it is better to guide the user to tell them what we do support

cmd/secrets-provider/main.go
sigalsax: If it is anything else but sidecar, we won't catch it. I want to fail here when it is anything except init or application.

orenbm: not quite. what we are trying to do, as mentioned in the comment, is to `verify we don't run with a K8s secrets storage configuration in a side car`.

prior to your change, container_mode that is not `init` was `sidecar` and that is why we wrote the `if` condition as it is. However, i think that it was written in a wrong way and that it should have been `if secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode == "sidecar"`

pkg/log/messages/error_messages.go
sigalsax: > would love to know what the sigal container mode gives you!

LOTS of avocados

cmd/secrets-provider/main.go
```diff
@@ -34,10 +34,8 @@ func main() {
 		printErrorAndExit(messages.CSPFK015E)
 	}
 
-	// verify we don't run with a K8s secrets storage configuration in a side car
-	if secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode != "init" {
-		printErrorAndExit(messages.CSPFK007E)
-	}
+	// verify we run only with the supported modes of deployment
```
orenbm, line 37: the comment is redundant

cmd/secrets-provider/main.go
```diff
@@ -128,3 +126,21 @@ func configureLogLevel() {
...
+	isValidContainerMode := false
+	for _, validContainerModeType := range validContainerModes {
+		if containerMode == validContainerModeType {
+			isValidContainerMode = true
+		}
+	}
+
+	if !isValidContainerMode {
+		printErrorAndExit(fmt.Sprintf(messages.CSPFK007E, containerMode, validContainerModes))
+	}
```
orenbm, line 145: LOVE this code :)

pkg/log/messages/error_messages.go
```diff
@@ -20,7 +20,7 @@ const CSPFK003E string = "CSPFK003E AccessToken failed to delete access token da
 // Environment variables
 const CSPFK004E string = "CSPFK004E Environment variable '%s' must be provided"
 const CSPFK005E string = "CSPFK005E Provided incorrect value for environment variable %s"
-const CSPFK007E string = "CSPFK007E Setting SECRETS_DESTINATION environment variable to 'k8s_secrets' must run as init container"
+const CSPFK007E string = "CSPFK007E CONTAINER_MODE '%s' is not supported. Supported values are: %v"
```
orenbm, line 23: can you check if this log code is written in the docs?

pkg/log/messages/error_messages.go
orenbm: image
`container_mode=sigal`, illustration

orenbm: which one is the sidecar?

pkg/log/messages/error_messages.go
sigalsax: Screen Shot 2020-08-13 at 11 01 47 AM

it is

sigalsax: Screen Shot 2020-08-13 at 11 49 33 AM
it is

pkg/log/messages/error_messages.go
sigalsax: Such a philosophical question because the avocados in the plastic box are totally containers and the box is the pod. SO I would say the water surrounding the cados is the sidecar (why is there even water in the plastic bin..?)

pkg/log/messages/error_messages.go
orenbm: so we need to update that. can you file an issue that changes the issue and resolution?

sigalsax: Here we are: https://github.com/cyberark/conjur-docs/issues/1079

sigalsax:
Link to doc issue for updating our docs with the new log: https://github.com/cyberark/conjur-docs/issues/1079

cmd/secrets-provider/main.go
abrahamko: I'm 😱 !
Just found out there's no way to find an element in an array in Go!
Can't believe it...

orenbm: yeah i had the same shock around 2 weeks ago. crazy that they don't have a built-in method

Issue #174, opened 2020-08-09
Secrets Provider - Add configurable retry mechanism


sigalsax:
For this card we need to make our retry mechanism configurable for better troubleshooting and to support network lag.

DOD
- [x] Add `RETRY_COUNT` and `TIMEOUT` Helm chart ({{- if .Values.environment.timeout }})
- [x] Update Json schema accordingly
- [x] Make the retry mechanism configurable using env vars passed from defaults in the custom-values.yaml (default should be defined in code TIMEOUT: 30 seconds, RETRY_COUNT: 3
- [x] Add proper, last log from here https://github.com/cyberark/conjur-docs/issues/1046

alexkalish:
@sigalsax: Could you provide some details about why this change was made?  What motivated it?  Trying to update the release notes and would like to understand it a bit better.  In the future, it would be great if the user value of a change could be captured in the story and the CHANGELOG entry, whenever possible.  Thanks!

alexkalish:
@sigalsax: Two more questions:
1. How are these values configured?  Looking at the PR diff, they appear to be ENV variables that can be set in the Helm chart.  Is that right?
2. Are these configuration values and their proper use documented?

sigalsax:
@alexkalish to start, there was a problem in our release process pushing to dockerhub due to an infra issue so this bump was not successful. The problem has since been fixed and we plan to release a new version when our documentation will be ready (which is in progress) @eladkug and I [created this release doc](https://ca-il-confluence.il.cyber-ark.com/display/rndp/%5B11.7%5D+CyberArk+Secret+Provider+for+K8s+Release) which you might find helpful in the release process. 

> it would be great if the user value of a change could be captured in the story and the CHANGELOG entry, whenever possible

Noted! No problem 💯 I want to note that the retry mechanism wasn't intended to be forward facing/documented and be used for troubleshooting purposes. @InbalZilberman feel free to expand here. 

The idea for this feature came from the [secrets provider epic](https://app.zenhub.com/workspaces/palmtree-5d99d900491c060001c85cba/issues/cyberark/secrets-provider-for-k8s/163) created by @InbalZilberman and decided to stay as an internal feature.

Regarding your questions:

> How are these values configured? Looking at the PR diff, they appear to be ENV variables that can be set in the Helm chart. Is that right?

[...]

PR #182, opened 2020-08-13
Add configurable retry mechanism for Secrets Provider deployment

sigalsax:
### What does this PR do?
- Removes exponential backoff and replaces with constant backoff with defaults
- Update Helm chart and JSON schema, adding RETRY_COUNT and SYNC_INTERVAL_TIME variables for a configurable retry

Connected to #174 174

#### Change log


cmd/secrets-provider/main.go
```diff
@@ -35,7 +35,7 @@ func main() {
 	}
 
 	// verify we don't run with a K8s secrets storage configuration in a side car
-	if secretsConfig.StoreType == secretsConfigProvider.K8S && authnConfig.ContainerMode != "init" {
+	if secretsConfig.StoreType == secretsConfigProvider.K8S && (authnConfig.ContainerMode != "init" && authnConfig.ContainerMode != "application") {
```
orenbm, line 38: i think you need to rebase on master. please do so and ping me when the PR is ready.

sigalsax: @orenbm we should be good to go

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Red Hat certified version of the secrets-provider-for-k8s image ([cyberark/secrets-provider-for-k8s#93](https://github.com/cyberark/secrets-provider-for-k8s/pull/93))
 
 ### Changed
+- The retry backoff for the Secrets Provider is now constant instead of exponential ([cyberark/secrets-provider-for-k8s#182](https://github.com/cyberark/secrets-provider-for-k8s/pull/182))
```
orenbm, line 13: do we have an issue for this? it is best to link to issues in the CHANGELOG rather than PRs.

cmd/secrets-provider/main.go
```diff
@@ -51,51 +52,55 @@ func main() {
...
 	}
 
-	// Configure exponential backoff
-	expBackoff := backoff.NewExponentialBackOff()
-	expBackoff.InitialInterval = 2 * time.Second
-	expBackoff.RandomizationFactor = 0.5
-	expBackoff.Multiplier = 2
-	expBackoff.MaxInterval = 15 * time.Second
-	expBackoff.MaxElapsedTime = 2 * time.Minute
+	syncIntervalTime := 3
```
orenbm, line 55: do we want to have this as some costant?

cmd/secrets-provider/main.go
```diff
@@ -51,51 +52,55 @@ func main() {
...
+	if envSyncIntervalVal := os.Getenv("SYNC_INTERVAL_TIME"); envSyncIntervalVal != "3" && envSyncIntervalVal != "" {
+		syncIntervalTime, _ = strconv.Atoi(envSyncIntervalVal)
+	}
```
orenbm, line 58: it is best to move this to the config file. that is the place where we read the env to set the config

cmd/secrets-provider/main.go
```diff
@@ -51,51 +52,55 @@ func main() {
...
+
+	retryCount := 3
+	if envRetryCount := os.Getenv("RETRY_COUNT"); envRetryCount != "3" && envRetryCount != "" {
+		retryCount, _ = strconv.Atoi(envRetryCount)
+	}
```
orenbm, line 63: same as above. constant + read env in other file

helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -29,12 +29,6 @@ spec:
         imagePullPolicy: {{ .Values.secretsProvider.imagePullPolicy }}
         name: {{ .Values.secretsProvider.name }}
         env:
-        - name: SECRETS_DESTINATION
-          value: k8s_secrets
-
-        - name: CONTAINER_MODE
-          value: application
```
orenbm, line 36: why was this removed?

pkg/log/messages/info_messages.go
```diff
@@ -18,7 +18,6 @@ const CSPFK003I string = "CSPFK003I Retrieving following secrets from DAP/Conjur
 const CSPFK004I string = "CSPFK004I Creating Kubernetes client"
 const CSPFK005I string = "CSPFK005I Retrieving Kubernetes secret '%s' from namespace '%s'"
 const CSPFK006I string = "CSPFK006I Updating Kubernetes secret '%s' in namespace '%s'"
-const CSPFK007I string = "CSPFK007I Waiting for %s to re-authenticate and fetch secrets."
 const CSPFK008I string = "CSPFK008I CyberArk Secrets Provider for Kubernetes v%s starting up"
 const CSPFK009I string = "CSPFK009I DAP/Conjur Secrets updated in Kubernetes successfully"
-const CSPFK010I string = "CSPFK010I Secrets Provider for Kubernetes completed successfully"
+const CSPFK010I string = "CSPFK010I Updating Kubernetes Secrets: %d retries out of %d"
```
orenbm, line 23: @shulifink please review

pkg/log/messages/error_messages.go
```diff
@@ -54,4 +54,5 @@ const CSPFK036E string = "CSPFK036E Variable ID '%s' is not in the format '<acco
 const CSPFK037E string = "CSPFK037E Failed to parse DAP/Conjur variable IDs"
 
 // General
+const CSPFK039E string = "CSPFK039E Secrets Provider for Kubernetes failed to update Kubernetes Secrets"
```
orenbm, line 57: can you move it to be after 38?

pkg/log/messages/error_messages.go
```diff
@@ -54,4 +54,5 @@ const CSPFK036E string = "CSPFK036E Variable ID '%s' is not in the format '<acco
...
 const CSPFK038E string = "CSPFK038E Retransmission backoff exhausted"
```
orenbm, line 58: is this still used?

cmd/secrets-provider/main.go
sigalsax: Yes, this is a default

helm/secrets-provider/templates/secrets-provider.yaml
sigalsax: wasn't removed, it was pushed below with all the other configs

cmd/secrets-provider/main.go
sigalsax: they are defaults not constants

pkg/log/messages/error_messages.go
sigalsax: yes if after 3 retries, the user should hit this log

cmd/secrets-provider/main.go
orenbm: i meant that we add a constant instead of having the `3` spreading around the code in case the default is changed. you can do this in the config file

pkg/log/messages/error_messages.go
sigalsax: I used `printErrorAndExit` instead of log.error which is wrong. I will update this

cmd/secrets-provider/main.go
sigalsax: const list will now look like
```
const (
	K8S            = "k8s_secrets"
	CONJUR_MAP_KEY = "conjur-map"
	DEFAULT_RETRY_COUNT = 3
	DEFAULT_SYNC_INTERVAL = 3
)
```

pkg/log/messages/info_messages.go
sigalsax: this was already reviewed

cmd/secrets-provider/main.go
sigalsax: added to consts

pkg/log/messages/error_messages.go
sigalsax: jumped all the logic up to after count = retry count so the deleting of accessToken still takes place

cmd/secrets-provider/main.go
```diff
@@ -51,63 +51,52 @@ func main() {
...
 
-	// Configure exponential backoff
-	expBackoff := backoff.NewExponentialBackOff()
-	expBackoff.InitialInterval = 2 * time.Second
-	expBackoff.RandomizationFactor = 0.5
-	expBackoff.Multiplier = 2
-	expBackoff.MaxInterval = 15 * time.Second
-	expBackoff.MaxElapsedTime = 2 * time.Minute
-
+	// Configure constant backoff for application container
```
orenbm, line 54: what do you mean by `for application container`? this will not happen also from init container?

cmd/secrets-provider/main.go
```diff
@@ -51,63 +51,52 @@ func main() {
...
+	constantBackOff := backoff.NewConstantBackOff(time.Duration(secretsConfig.SyncIntervalTime) * time.Minute)
+	count := 0
 	err = backoff.Retry(func() error {
```
orenbm, line 57: i may be missing something but it seems you are not doing anything with the returned `err` object. what happens if an error is returned somewhere inside the backoff loop? will this block still run:
```
/ Deleting the retrieved Conjur access token in case we got an error after retrieval.
			// if the access token is already deleted the action should not fail
			err = accessToken.Delete()
			if err != nil {
				log.Error(messages.CSPFK003E, err)
			}
			log.Error(messages.CSPFK038E)
			printErrorAndExit(messages.CSPFK039E)
```

cmd/secrets-provider/main.go
```diff
@@ -51,63 +51,52 @@ func main() {
...
-			if err != nil {
-				return log.RecordedError(messages.CSPFK011E)
-			}
-
-			err = provideConjurSecrets(accessToken)
-			if err != nil {
-				return log.RecordedError(messages.CSPFK016E)
-			}
-
+		if count > secretsConfig.RetryCount {
```
orenbm, line 58: i have a suggestion that i think will make the code clearer. currently we return different errors in case they occurred and catch them after the backoff loop. Then, we clean up the token and print the errors.

However, not you add a new behaviour where you handle the error (retry count limit reached) inside the loop itself.

What you can do is return an error in this case and still catch all the errors outside of the loop.

helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -65,9 +59,25 @@ spec:
...
         - name: K8S_SECRETS
           value: {{ .Values.environment.k8sSecrets | join "," }}
 
+        {{- if .Values.environment.conjur.syncIntervalTime }}
+        - name: SYNC_INTERVAL_TIME
+          value: {{- .Values.environment.conjur.syncIntervalTime }}
+        {{- end }}
+
+        {{- if .Values.environment.conjur.retryCount }}
+        - name: RETRY_COUNT
```
orenbm, line 77: how about `RETRY_COUNT_LIMIT`? or something like that. `count` sounds like an ongoing procedure

cmd/secrets-provider/main.go
sigalsax: small left over. removing

pkg/secrets/config/config.go
```diff
@@ -49,10 +54,22 @@ func NewFromEnv() (*Config, error) {
 		return nil, err
 	}
 
+	syncIntervalTime := DEFAULT_RETRY_COUNT
+	if envSyncIntervalVal := os.Getenv("SYNC_INTERVAL_TIME"); envSyncIntervalVal != "3" && envSyncIntervalVal != "" {
```
orenbm, line 58: you can use `DEFAULT_RETRY_COUNT`

cmd/secrets-provider/main.go
sigalsax: I just removed the `err =` 
if there is an error in the backoff loop, it will try again and again until that if is true
```
if count > secretsConfig.RetryCount {
}
```
Once the above is true we will process will exit with 1 error code and the AccessToken will be deleted and "Secrets Provider for Kubernetes failed to update Kubernetes Secrets" will be added to logs

Some background, I needed a way to break out of the loop because and that above if statement does just that

cmd/secrets-provider/main.go
```diff
@@ -51,63 +51,51 @@ func main() {
...
+			log.Info(fmt.Sprintf(messages.CSPFK010I, count, secretsConfig.RetryCount))
+		}
+		count++
 
-			// Reset exponential backoff
-			expBackoff.Reset()
+		log.Info(fmt.Sprintf(messages.CSPFK001I, authn.Config.Username))
+		authnResp, err := authn.Authenticate()
+		if err != nil {
+			return log.RecordedError(messages.CSPFK010E)
```
orenbm, line 75: i may be missing something but it seems that you are returning this error but it is never used. am i wrong?

abrahamko: Returning this error triggers the retry loop.

helm/secrets-provider/values.schema.json
```diff
@@ -112,6 +112,16 @@
...
               "minLength": 1
             },
+            "timeout": {
+              "type": "number",
+              "minLength": 1,
+              "minimum": 5
+            },
+            "retryCountLimit": {
+              "type": "number",
+              "minLength": 1,
```
abrahamko, line 122: remove, only valid for strings and arrays.

helm/secrets-provider/values.schema.json
```diff
@@ -112,6 +112,16 @@
               "type": "string",
               "minLength": 1
             },
+            "timeout": {
+              "type": "number",
+              "minLength": 1,
```
abrahamko, line 117: remove, only valid for strings and arrays.

helm/secrets-provider/values.schema.json
```diff
@@ -112,6 +112,16 @@
...
+              "minimum": 5
+            },
+            "retryCountLimit": {
+              "type": "number",
+              "minLength": 1,
+              "minimum": 1
```
abrahamko, line 123: Why 0 is not a valid retry limit?
I may want to not retry at all.

helm/secrets-provider/values.schema.json
```diff
@@ -112,6 +112,16 @@
               "type": "string",
               "minLength": 1
             },
+            "timeout": {
+              "type": "number",
+              "minLength": 1,
+              "minimum": 5
```
abrahamko, line 118: 5 what? minutes? seconds? milliseconds?
Better call it `timeoutSec` to ensure readability.

abrahamko, line 118: why minimum 5?
I would assume non-negative is enough, although in many places `timeout=-1` means forever, which is also valid.
What say you?
@oburstein-hub do you have something to add? should we allow 0 or even -1 as valid timeout values?

pkg/secrets/config/config.go
```diff
@@ -49,10 +54,22 @@ func NewFromEnv() (*Config, error) {
...
 
+	timeout := DEFAULT_TIMEOUT
+	if envTimeout, _ := strconv.Atoi(os.Getenv("TIMEOUT")); envTimeout != DEFAULT_TIMEOUT && envTimeout != 0 {
+		timeout = envTimeout
+	}
+
+	retryCountLimit := DEFAULT_RETRY_COUNT_LIMIT
+	if envRetryCountLimit, _ := strconv.Atoi(os.Getenv("RETRY_COUNT_LIMIT")); envRetryCountLimit != DEFAULT_RETRY_COUNT_LIMIT && envRetryCountLimit != 0 {
+		retryCountLimit = envRetryCountLimit
+	}
```
abrahamko, line 65: You can extract this method into something like:
```
ParseIntFromEnvOrDefault("RETRY_COUNT_LIMIT", DEFAULT_RETRY_COUNT_LIMIT)
```
Then you can reuse it for both env vars.

pkg/secrets/config/config.go
```diff
@@ -49,10 +54,22 @@ func NewFromEnv() (*Config, error) {
 		return nil, err
 	}
 
+	timeout := DEFAULT_TIMEOUT
+	if envTimeout, _ := strconv.Atoi(os.Getenv("TIMEOUT")); envTimeout != DEFAULT_TIMEOUT && envTimeout != 0 {
+		timeout = envTimeout
+	}
+
+	retryCountLimit := DEFAULT_RETRY_COUNT_LIMIT
+	if envRetryCountLimit, _ := strconv.Atoi(os.Getenv("RETRY_COUNT_LIMIT")); envRetryCountLimit != DEFAULT_RETRY_COUNT_LIMIT && envRetryCountLimit != 0 {
```
abrahamko, line 63: why do you check for value != 0?
Is this some kind of input validation?
If so, it should be clear. Since retry = 0 might be a valid input.
If it is to avoid the default value in case of parsing error, better check for `err != nil`, then use the default.

helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -65,9 +59,24 @@ spec:
...
+
+        # Enables the support of multiple Kubernetes applications
+        - name: CONTAINER_MODE
+          value: application
+
         - name: K8S_SECRETS
           value: {{ .Values.environment.k8sSecrets | join "," }}
 
+        - name: TIMEOUT
+          value: {{- .Values.environment.conjur.timeout }}
```
abrahamko, line 73: Since they are not mandatory, and do not have default values, what will be the value of this if user does not supply it? (which he doesn't even know he can...)

cmd/secrets-provider/main.go
```diff
@@ -51,51 +51,42 @@ func main() {
...
-
-			fmt.Println()
-			time.Sleep(authn.Config.TokenRefreshTimeout)
+		err = retryCondition(count, secretsConfig.RetryCountLimit)
+		if err != nil {
+			return err
+		}
+		count++
+
+		log.Info(fmt.Sprintf(messages.CSPFK001I, authn.Config.Username))
```
abrahamko, line 63: Extract to function from this line to the end of the anonymous function.
Something like: `provideSecretsToTarget` or anything else you like.

cmd/secrets-provider/main.go
```diff
@@ -51,51 +51,42 @@ func main() {
...
+		err = accessToken.Delete()
+		if err != nil {
+			return log.RecordedError(messages.CSPFK003E, err.Error())
 		}
-	}, expBackoff)
+
+		if authnConfig.ContainerMode == "init" || authnConfig.ContainerMode == "application" {
+			log.Info(messages.CSPFK009I)
+			os.Exit(0)
+		}
```
abrahamko, line 87: Why do you need this?
the next line `return nil` will exit the process anyway.

helm/secrets-provider/values.schema.json
sigalsax: @oburstein-hub ^

helm/secrets-provider/values.schema.json
sigalsax: name changed to retryInterval because it wasn't really a timeout

helm/secrets-provider/templates/secrets-provider.yaml
sigalsax: this was updated to 
```

        {{- if .Values.environment.conjur.retryInterval }}
        - name: RETRY_INTERVAL
          value: {{- .Values.environment.conjur.retryInterval }}
        {{- end }}

        {{- if .Values.environment.conjur.retryCountLimit }}
        - name: RETRY_COUNT_LIMIT
          value: {{ .Values.environment.conjur.retryCountLimit }}
        {{- end }}

```

pkg/secrets/config/config.go
sigalsax: if the user didn't provide in these env vars value they would programmatically equate to 0 so this is to ensure that this doesn't happen

cmd/secrets-provider/main.go
oburstein-hub: Please, turn "init" and "application" to constant String parameters

cmd/secrets-provider/main.go
oburstein-hub: Why do you need comparison to 3 here ? Atoi will handle it anyway, right ?

cmd/secrets-provider/main.go
oburstein-hub: you dont need comparison with "3" here either

cmd/secrets-provider/main.go
oburstein-hub: why are they multiplied by 15 and by 2 ?

helm/secrets-provider/values.schema.json
oburstein-hub: because it will create busy waiting

helm/secrets-provider/templates/secrets-provider.yaml
oburstein-hub: why are you putting these "ifs" there ? the parameters are always in Values, arent they ?

helm/secrets-provider/values.schema.json
sigalsax: @oburstein-hub ^

cmd/secrets-provider/main.go
sigalsax: the red means it was removed @oburstein-hub

helm/secrets-provider/templates/secrets-provider.yaml
sigalsax: no they are optional. If the user doesn't supply them we take defaults that are defined in code

helm/secrets-provider/values.schema.json
oburstein-hub: it is not convinient to put timeout 0, although -1 is a common way to sign infinite timeout. 1 can be a minimal value

helm/secrets-provider/templates/secrets-provider.yaml
oburstein-hub: I think the better way it always to provide these parameters from helm chart. If the user does not want to provide them he will not put them in custom-values.yaml and defaults from values.yaml will be used

pkg/secrets/config/config.go
```diff
@@ -49,13 +54,27 @@ func NewFromEnv() (*Config, error) {
...
 		RequiredK8sSecrets: requiredK8sSecrets,
+		RetryCountLimit:    retryCountLimit,
+		RetryInterval:      retryInterval,
 		StoreType:          storeType,
 	}, nil
 }
 
+func ParseIntFromEnvOrDefault(environmentVariable string, defaultVariable int) int {
+	retry := defaultVariable
+	if envVar, _ := strconv.Atoi(os.Getenv(environmentVariable)); envVar != defaultVariable && envVar != 0 {
```
abrahamko, line 72: instead of `envVar != 0`, which is a valid value, use `err != nil`, which indicated parsing error occurred.
Also, I suggest to document that the we failed to parse the value, what do you think?

helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -63,11 +57,28 @@ spec:
...
+
+        # Enables the support of multiple Kubernetes applications
+        - name: CONTAINER_MODE
+          value: application
 
         - name: K8S_SECRETS
           value: {{ .Values.environment.k8sSecrets | join "," }}
 
+        {{- if .Values.environment.conjur.retryInterval }}
+        - name: RETRY_INTERVAL
```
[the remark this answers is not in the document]
sigalsax: adding

helm/secrets-provider/values.schema.json
sigalsax: adding

PR #187, opened 2020-08-17
Add fix for backoff for proper log placement

sigalsax:
Previously, the retry loop was printing upon failure. With this fix, the retry log will only be printed if there is a failure

Before:
Screen Shot 2020-08-17 at 4 25 27 PM

After:
Screen Shot 2020-08-17 at 4 30 38 PM


sigalsax:
closing because this change was not needed

PR #205, opened 2020-08-25
Add batch retrieval solution design

sigalsax:
### What does this PR do?
Add solution design for new batch retrieval API endpoint

### Background
*Previously*, for the init container solution, each Secrets Provider was deployed to the same Pod as the customer application containers. Because of this, when a batch retrieval request was made, the requests were dispersed across many call to the Conjur server. When making the batch retreival request and a permission error occured for example, then the batch request would fail and only that Pod would not spin up successfully. 

*Now*, with the Job Milestone, the Secrets Provider is deployed outside of the customers' Pods, supporting multiple customer pods. With this, a single batch retrieval request will be made for *all* applications to fetch the necessary secrets to deploy properly. Therefore, if there is a failure in receiving one of the values, then the whole request will fail. In other words, **no application** Pod will spin up in the namespace. This leads to a faulty user experience.

Connected to -

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/design/batch_retrieval_design.md
+++ b/design/batch_retrieval_design.md
@@ -73,23 +73,23 @@ As a small recap, the following explains the Secrets Provider init container flo
 
 4. The `cyberark-secrets-provider-for-k8s` init container runs to completion.
 
-The new Batch Retrieval API will effect #2a. 
+The new Batch Retrieval API will effect #2a. This will effect both init and application container (Milestone Job) flows.
 
 ### URI
 
-##### POST /secrets/batch/v2
+##### POST /secrets
 
 This method retuns a list of secrets, their status code, and their value if an error did not occur. For differences between proposed API and existing API see [Proposed API vs Current API](#proposed-api-vs-current-api) section below.
 
 ```
-POST /secrets/batch/v2
+POST /secrets
 ```
 
 ### Request Body
 
 The request body is a list of all variable IDs (paths to secrets in Conjur/DAP) that need to be retrieved. For example: 
 
-```
+```json
 {
   "variables": [
     {"id": "var/path/conjurSecret"},
@@ -110,7 +110,7 @@ Authorization: Token token="eyJkYX...Rhb="
 
 #### Batch Response
 
-```
+```json
 # Headers
 HTTP/1.1
 Content-Type: application/json
@@ -128,35 +128,38 @@ Content-Type: application/json
   },
   {
     "id": "var/path/name2",
-    "status": 403
+    "status": 403,
     "error": {
       "message": "Forbidden"
     }
   },
   {
     "id": "var/path/name3",
-    "status": 404
+    "status": 404,
     "error": {
       "message": "Not Found"
  		}
   },
   {
     "id": "var/path/name4",
-    "status": 200
+    "status": 200,
     "value": "superSuperSecret"    
   }
  ]
 }
 ```
 
-| Code            | Description                                                  |
-| --------------- | ------------------------------------------------------------ |
-| 200 OK          | All secret values were retrieved successfully                |
-| 400 Bad Request | Server cannot process request due to client error (unparsable request body) |
+| Code             | Description                                                  |
+| ---------------- | ------------------------------------------------------------ |
+| 200 OK           | All secret values were retrieved successfully                |
+| 400 Bad Request  | Server cannot process request due to client error (unparsable request body) |
+| 401 Unauthorized | The request lacks valid authentication credentials           |
 
 If no Conjur/DAP secrets were able to be retrieved, the response will still return a 200 OK because the [request body is parsable]([https://docs.microsoft.com/en-us/graph/json-batching#:~:text=The%20status%20code%20on%20a,requests%20inside%20the%20batch%20succeeded](https://docs.microsoft.com/en-us/graph/json-batching#:~:text=The status code on a,requests inside the batch succeeded)) and request was handled.
 
 If an invalid request body was supplied, a 400 Bad Request will be returned.
+
+If the request was sent with an invalid access token, a 401 Unauthorized response will be returned.
 
 #### Response body
 
@@ -173,18 +176,17 @@ All secret values will be treated as individual entities. If a single secret fai
 
 The following are the possible error codes that can be returned for individual secrets.
 
-| Status/Message   | Description                                                  |
-| ---------------- | ------------------------------------------------------------ |
-| 400 Bad Request  | Server cannot process request due to client error (invalid input parameters or logical validations). For example: adding `<>` in path to secret in Conjur/DAP |
-| 401 Unauthorized | The request lacks valid authentication credentials           |
-| 403 Forbidden    | The authenticated user lacks the necessary privilege         |
-| 404 Not Found    | Variable does not exist or variable does not have secret value |
+| Status/Message  | Description                                                  |
+| --------------- | ------------------------------------------------------------ |
+| 400 Bad Request | Server cannot process request due to client error (invalid input parameters or logical validations). For example: adding `<>` in path to secret in Conjur/DAP |
+| 403 Forbidden   | The authenticated user lacks the necessary privilege         |
+| 404 Not Found   | Variable does not exist or variable does not have secret value |
 
 ### Proposed API vs Current API
 
 |               | Current                                                      | Proposed                                                     |
 | ------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
-| URI           | `GET /secrets{?variable_ids}}`                               | `POST /secrets/batch/v2`                                     |
+| URI           | `GET /secrets{?variable_ids}}`                               | `POST /secrets`                                              |
 | Request Body  | -                                                            | See **Request Body** section                                 |
 | Response Body | JSON of ***only*** successfully retrieved secrets and their values. Single failure, fails whole request and no secret is returned. | JSON of **all** secrets, their status code/messages, and values for successfully retrieved secrets |
 | Behavior      | Request fails when individual request fails                  | Request does not fail when sub/individual requests fail      |
@@ -201,12 +203,153 @@ Now that it is easy to brute-force multiple variables at a time in the body of t
 
 ### Audit
 
-A new audit entry will made per secret fetched from Conjur/DAP as is done in the current Batch retrieval API endpoint.
+A new audit entry will made per secret fetched from Conjur/DAP as is done in the current Batch retrieval API endpoint. Also for all secrets that are attempted to be fetched and that the batch retrieval request was made.
+
+### Test Plan
+
+#### Conjur
+
+The following integration tests will be implemented in Conjur
+
+|      | Scenario                                                     | Given                                                        | When                                                         | Then                                                         | Manual/Automated |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ---------------- |
+| 1    | Vanilla flow                                                 | Given I am an authenticated, privileged user/host<br />With proper permissions on requested secrets | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/anotherConjurSecret` in request body | - The JSON response should be (*see below* #1)<br />- Status response code should be `200 OK`<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 2    | User/Host authenticated but does not have `execute` privilege | Given I am a privileged, authenticated user/host<br />With proper permissions on ***one*** of requested secrets | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/anotherConjurSecret` in request body | - The JSON response should be (*see below* #2)<br />- Status response code should be `200 OK`<br />- Existing secret should have `200`. Non-existing secret should have `403` status error<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 3    | User/Host authenticated but secret doesn't exist             | Given I am a privileged, authenticated user/host<br />With proper `execute` permissions<br />Requesting secret does not exist | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/nonExistingSecret` in request body | - The JSON response should be (*see below #3*)<br />- Status response code should be `200 OK`<br />- Existing secret should have `200`. Non-existing secret should have `404 ` status error<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 4    | User/Host authenticated but secret does not have value       | Given I am a privileged, authenticated user/host<br />With proper permissions<br />Requesting secret *value* does not exist | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/nonSecretValue` in request body | - The JSON response should be (*see below #4)<br />- Status response code should be `200 OK`<br />- Existing secret should have `200`. Non-existing secret value should have `404 ` status error<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 5    | Invalid request body                                         | Given I am a privileged, authenticated user/host<br />With proper permissions<br />Requesting secret in invalid format | I send `POST /secrets` with invalid JSON body with `var/path/conjurSecret` and  `var/path/<>` | - The JSON response should be (see below #5)<br />- Status response code should be `400 `<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 6    | Security Mitigation tests                                    |                                                              |                                                              |                                                              |                  |
+
+*Test 1 response body*
+
+```json
+200 OK
+
+{
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+  {
+    "id": "var/path/anotherConjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  }
+ ] 
+}
+```
+
+*Test 2 response body*
+
+```json
+200 OK
+
+{
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+    {
+    "id": "var/path/anotherConjurSecret",
+    "status": 403,
+    "error": {
+      "message": "Forbidden"
+ 		}
+  },
+ ] 
+}
+```
+
+*Test 3 response body*
+
+```json
+200 OK
+
+{
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+    {
+    "id": "var/path/nonexistingSecret",
+    "status": 404,
+    "error": {
+      "message": "Not Found"
+ 		}
+  },
+ ] 
+}
+```
+
+*Test 4 response body*
+
+```json
+200 OK
+
+{
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+    {
+    "id": "var/path/nonSecretValue",
+    "status": 404,
+    "error": {
+      "message": "Not Found"
+ 		}
+  },
+ ] 
+}
+```
+
+*Test 5 response*
+
+```json
+400 Bad Request
+```
+
+
+
+#### Secrets Provider for K8s
+
+The following integration tests will be implemented in Secrets Provider for K8s
+
+|      | Scenario                                                     | Given                                                        | When                           | Then                                                         | Manual/Automated |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------ | ------------------------------------------------------------ | ---------------- |
+| 1    | *Vanilla flow*, Secret Provider Job successfully updates K8s Secrets | - Conjur is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and mapped in K8s Secrets<br />- All mandatory values are defined in `custom-values.yaml`<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job | - Secrets Provider pod authenticates and fetches Conjur secrets successfully<br />- All K8s Secrets are updated with  Conjur value <br />- Secrets Provider Job runs to completion<br />- Verify logs | Automated        |
+| 2    | Partial success                                              | - Conjur is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and mapped in K8s Secrets<br />- All mandatory values are defined in `custom-values.yaml`<br />- Host defined to Secrets Provider in `custom-values.yaml` does not have access to one of the Conjur secrets<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job | - Secrets Provider pod authenticates and attempts to fetch Conjur secrets <br />- All K8s Secrets are updated with those Conjur values successfully fetched <br />- Secrets Provider Job runs to exhaustion<br />- Verify logs #1/2/3 appear | Automated        |
+| 3    | *Vanilla test,* Conjur server is not up-to-date              | - Outdated Conjur server is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and mapped in K8s Secrets<br />- All mandatory values are defined in `custom-values.yaml`<br />- Host defined to Secrets Provider in `custom-values.yaml` <br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job | - Secrets Provider pod authenticates and attempts to fetch Conjur secrets<br />- Application deploys successfully<br />- Verify logs #4 appear | Automated        |
+| 4    | Conjur server is not up-to-date                              | - Outdated Conjur server is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and mapped in K8s Secrets<br />- All mandatory values are defined in `custom-values.yaml`<br />- Host defined to Secrets Provider in `custom-values.yaml` does not have access to one of the Conjur secrets<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job | - Secrets Provider pod authenticates and attempts to fetch Conjur secrets<br /><br />- Secrets Provider Job runs to exhaustion<br />- Application fails to be deployed<br />- Verify logs #1/4 appear | Automated        |
+
+## Logs
+
+|      | **Scenario**                                                 | **Log message**                                              | Log level |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | --------- |
+| 1    | Secrets Provider batch retrieval has partial success (Conjur) | Failed to retrieve a portion of the Conjur/DAP secrets: Reason: '%s' | Error     |
+| 2    | Secrets Provider batch retrieval total failure (Conjur)      | Failed to retrieve Conjur/DAP secrets: Reason: '%s'          | Error     |
+| 3    | Secrets Provider batch retrieval has partial success (Secrets Provider) | Skipped Kubernetes Secrets '%s'. Reason: Failed to retrieve DAP/Conjur secrets | Debug     |
+| 4    | Secrets Provider success                                     | Successfully updated '%d' out of '%d' Kubernetes Secrets     | Info      |
+| 5    | Old batch retrieval endpoint                                 | Warning: Secrets Provider cannot efficiently run because DAP/Conjur is not up to date. Please consider upgrading to the latest version | Warn      |
+
+## Performance tests
+
+@Abraham add here
+
+
 
 ### Open questions
 
 1. Do we want this duplicated API documented? At this stage this API endpoint will be for internal use only.
 2. Do we ultimately want to move to a single Batch Request API and depricate the existing one?
+3. Do we want to support v4?
 
 ### Future enhancements
 
@@ -215,3 +358,23 @@ The following are ideas for possible extension of the batch retrieval endpoint A
 1. Add secret versions as a filtering query parameter so we can get all secrets after a certain version
 2. Fetch all secrets that the user/host has permissions to
 3. Fetch all secrets after `/var/path/*` that the user/host has permissions to
+
+### Delivery Plan
+
+- [ ] Fix flaky tests *2 days*
+- [ ] Implement new API endpoint in Conjur *5 days*
+- [ ] Add integration tests in Conjur *4 days*
+- [ ] Golang SDK
+  - [ ] Implement batch logic *2 days*
+  - [ ] Add unit tests *2 days*
+- [ ] Implement tests in Golang SDK *3 days*
+- [ ] Add change in Secrets Provider *3 days*
+- [ ] Implement test plan
+  - [ ] Add integration tests in Secrets Provider *3 days*
+- [ ] Performance tests
+  - [ ] In Conjur *TBD*
+  - [ ] In Secrets Provider *TBD*
+- [ ] Team/cross-team review cycle *3 days*
+- [x] Logs reviewed by TW *1 day*
+
+Total: 29 days ~6 weeks
--- a/design/batch_retrieval_design.md
+++ b/design/batch_retrieval_design.md
@@ -7,7 +7,7 @@
 + [Request Body](#request-body)
   - [Headers](#headers)
 + [Response](#response)
-+ [Proposed API vs Old API](#proposed-api-vs-old-api)
++ [Proposed API vs Current API](#proposed-api-vs-current-api)
 + [Permissions](#permissions)
 + [Open questions](#open-questions)
 
@@ -22,6 +22,11 @@ This document will cover the options for enhancing Batch Retrieval requests.
 | Secrets Provider Job Milestone - Solution Design | https://github.com/cyberark/secrets-provider-for-k8s/blob/master/design/milestone_1_2_design_doc.md |
 | Secrets Provider - Feature Doc                   | https://app.zenhub.com/workspaces/palmtree-5d99d900491c060001c85cba/issues/cyberark/secrets-provider-for-k8s/163 |
 
+### References
+
+- https://docs.microsoft.com/en-us/graph/json-batching#first-json-batch-request
+- https://docs.microsoft.com/en-us/graph/json-batching#response-format
+
 ### Motivation
 
 *Previously*, for the init container solution, each Secrets Provider was deployed to the same Pod as the customer application containers. Because of this, when a batch retrieval request was made, the requests were dispersed across many call to the Conjur server. When making the batch retreival request and a permission error occured for example, then the batch request would fail and only that Pod would not spin up successfully. 
@@ -32,9 +37,9 @@ The following table outlines possible solutions to overcome this:
 
 |      | Solution                                                     | Pros                                                         | Cons                                               | Effort (T-shirt size) |
 | ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | -------------------------------------------------- | --------------------- |
-| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                | S, 5 days             |
-| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Best practice / straightforward design for batch endpoint | - Requires both client/server implementation<br /> | M, 10 days            |
-| 3    | Stay as is                                                   | No additional work needed                                    | - Bad UX<br />- Performance issues                 | -                     |
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes                                     | Load on server with extra calls (*)                | S, 5 days             |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their individual responses (success/failure) (**) | - Best practice / straightforward design for batch endpoint<br />- Will help us during rotation for Milestone 2<br /> | - Requires both client/server implementation<br /> | M, 10 days            |
+| 3    | Stay as is                                                   | No additional work needed                                    | - Bad UX<br />                                     | -                     |
 
 (*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
 
@@ -43,47 +48,41 @@ The following table outlines possible solutions to overcome this:
 *Decision:* We decided to go with #2 and create our own API. We decided to create a new batch retrieval API because of the following reasons:
 
 1. Adding onto the existing batch endpoint would lead to changing behaviors for our needs in the Secrets Provider which would have cascading impacts on other teams. 
-2. The two APIs (current and new) offer separate use-cases;  partial success (new) and fail fast (old)
-3. The old batch retrieval API accepts variable IDs as query parameters. We anticipate a long string of variableIDs to be passed in for secret retrieval and this impacts performance.
+2. The two APIs (current and new) offer separate use-cases;  partial success (new) and fail fast (current)
+3. The current batch retrieval API accepts variable IDs as query parameters. We anticipate a long string of variableIDs to be passed in for secret retrieval and this impacts performance.
 4. Query parameters for HTTP GET requests are limited in length (2,048 characters) which is problematic if a customer has long secrets paths.
 5. The new route offers is more useful if we choose to expand its functionality in the future. By using POST instead of GET we can not only opt to use query parameters but also utilize the body of the request.
 
+At this stage this new batch retrieval API endpoint will be for internal use only.
+
 ### Design
 
-#### Secrets Provider Job flow
+#### Secrets Provider init container flow
 
-As a small recap, the following explains the Secrets Provider Job flow
+As a small recap, the following explains the Secrets Provider init container flow
 
-1. The `cyberark-secrets-provider-for-k8s` runs as a Job and authenticates with the Conjur/DAP Follower using the Kubernetes Authenticator (`authn-k8s`).
+1. The `cyberark-secrets-provider-for-k8s` runs as an init container and authenticates with the Conjur/DAP Follower using the Kubernetes Authenticator (`authn-k8s`).
 
 2. The `cyberark-secrets-provider-for-k8s` reads all Kubernetes secrets required by the applications in the same namespace.
 
-3. For each mapped Kubernetes secret, 
+3. For each mapped Kubernetes secret, the `cyberark-secrets-provider-for-k8s` 
 
-   a. The `cyberark-secrets-provider-for-k8s` performs a Batch Retrieval request and retrieves Conjur/DAP secrets.
+   a. Performs a Batch Retrieval request and retrieves Conjur/DAP secrets.
 
-   b. It updates the Kubernetes secret with the Conjur/DAP secret value.
+   b. Updates the Kubernetes secret with the Conjur/DAP secret value.
 
-4. The `cyberark-secrets-provider-for-k8s` Job is completed.
+4. The `cyberark-secrets-provider-for-k8s` init container runs to completion.
 
 The new Batch Retrieval API will effect #2a. 
 
 ### URI
 
-##### POST /secrets/batch
+##### POST /secrets/batch/v2
 
-This method retuns a list of secrets, their status code, and their value if an error did not occur. See difference between proposed API and existing API see [Proposed API vs Old API](#proposed-api-vs-old-api) below.
+This method retuns a list of secrets, their status code, and their value if an error did not occur. For differences between proposed API and existing API see [Proposed API vs Current API](#proposed-api-vs-current-api) section below.
 
 ```
-POST /secrets/batch
-
-{
-  "variables": [
-    {"name": "var/path/conjurSecret"},
-    {"name": "var/path/anotherConjurSecret"},
-    ...
-  ]
-}
+POST /secrets/batch/v2
 ```
 
 ### Request Body
@@ -93,8 +92,8 @@ The request body is a list of all variable IDs (paths to secrets in Conjur/DAP) 
 ```
 {
   "variables": [
-    {"name": "var/path/conjurSecret"},
-    {"name": "var/path/anotherConjurSecret"},
+    {"id": "var/path/conjurSecret"},
+    {"id": "var/path/anotherConjurSecret"},
     ...
   ]
 }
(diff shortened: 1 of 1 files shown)


design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                | S, 5 days             |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Best practice / straightforward design for batch endpoint | - Requires both client/server implementation<br /> | M, 10 days            |
+| 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                                             | -                     |
 
 (*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
 
 (**) This solution can be broken up in two: 1. Create a new API endpoint or 2. use existing one (return 409 response)
 
+*Decision:* We decided to go with #2 and create our own API. The new batch retrieval API endpoint will not be forward facing and be for **internal use only**.
```
orenbm, line 30: so the new endpoint will be fully developed but it won't be in the docs? other than not being in the docs, will it have any difference from the other endpoints?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+The new Batch Retrieval API will effect #2a. 
+
+### URI
+
+##### POST /secrets/batch
+
+This method retuns a list of secrets, their status code, and their value if an error did not occur. 
+
+```
+POST /secrets/batch
```
orenbm, line 59: is this the new API endpoint? so we'll now have:
- endpoint `/secrets` that performs a batch retrieval and fails if one secrets fails
- endpoint `/secrets/batch` that performs a batch retrieval and does not fail if one secrets fails

did i get it right?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Best practice / straightforward design for batch endpoint | - Requires both client/server implementation<br /> | M, 10 days            |
+| 3    | Stay as is                                                   | No additional work needed                                    | Bad UX                                             | -                     |
 
 (*) Fallback solution: use this solution as a safety net, only if the original Batch Retrieval request fails
 
 (**) This solution can be broken up in two: 1. Create a new API endpoint or 2. use existing one (return 409 response)
 
+*Decision:* We decided to go with #2 and create our own API. The new batch retrieval API endpoint will not be forward facing and be for **internal use only**. 
+
+### Design
```
orenbm, line 32: i think that this design is lacking info for how the new endpoint will be implemented. did you consider using the current batch retrieval api? if so, why did you decide not to use it?

design/batch_retrieval_design.md
orenbm: will both APIs have the query param `variable_ids`?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+    {"name": "var/path/conjurSecret"},
+    {"name": "var/path/anotherConjurSecret"},
+    ...
+  ]
+}
+```
+
+### Request Body
+
+The request body is a list of all variable IDs (paths to secrets in Conjur/DAP) that need to be retrieved. For example:
```
orenbm, line 72: in the existing batch retrieval api we use the query param `variable_ids` to say which secrets are required. why did you change this behaviour for this api?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+```
+
+### Response
+
+| Code             | Description                                                  |
+| ---------------- | ------------------------------------------------------------ |
+| 200 OK           | All secret values were retrieved successfully                |
+| 400 Bad Request  | Server cannot process request due to client error (invalid input parameters or logical validations) |
+| 401 Unauthorized | The request lacks valid authentication credentials           |
+| 403 Forbidden    | The authenticated user lacks the necessary privilege         |
```
orenbm, line 97: for all the secrets? what happens if the user has permissions for one secret but not for another?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+| 404 Not Found    | At least one variable does not exist, or at least one variable does not have any secret values |
```
orenbm, line 98: in this case will the secrets that were found be returned in the response body?

if so - isn't it wrong to return 404 but still return data?
if not - isn't this exactly what you wanted to fix?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+  "sucessSecret1":[
+     {
+       "code": 200
+       "value": "superSecret"
+     }
+  ]
+  "failSecret1":[
+    {
+      "code": 401
+      "value": "Unauthorized"
```
orenbm, line 115: i wouldn't put this as "value" but would rather add a new key "error" which will be optional (it will be present only if `code != 200`)

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+      "code": 200
+      "value": "superSuperSecret"    
+    }
+  ]
+}
+```
+
+All secret values will be treated as individual entities. If a single secret fails to be retrieved, this will not impact the integrity of the request.
+
+If no Conjur/DAP secrets were able to be retrieved (no 200 code), the Batch Retreival response fail with a 5xx error code.
```
orenbm, line 135: why 5xx?

design/batch_retrieval_design.md
```diff
@@ -2,19 +2,140 @@
...
+
+### Permissions
+
+Only an authorized Host with `read` and `execute` privileges on the Conjur/DAP secrets will be able to fetch the secret values. Otherwise, the proper 401 Unauthorized error code will be returned for `code` and `value` of the response respectively.
```
orenbm, line 139: 403 seems more relevant

design/batch_retrieval_design.md
orenbm: also - why isn't it customer facing?

sigalsax: We originally decided not to document because we already have a batch retrieval endpoint and so introducing another one is confusing because on paper, not much is different expect partial success. We thought to add onto the existing batch endpoint but that would lead to changing behaviors for our needs which is not what we wanted to accomplish

@abrahamko mentioned that from what he read it is better to document, even if there is duplication involved.

After convos on slack, it seems that this will be revisited. I will add this to an open question in the doc

design/batch_retrieval_design.md
sigalsax: Exactamundo!

design/batch_retrieval_design.md
sigalsax: > i think that this design is lacking info for how the new endpoint will be implemented.

As in adding it to `routes.rb` etc? Based on other docs I have seen on confluence, I actually think that is a bit too low level. Do you think differently?

> did you consider using the current batch retrieval api? if so, why did you decide not to use it?

Indeed! We chose to add another Batch Retrieval API for the following reasons:
1. Add onto the existing batch endpoint would lead to changing behaviors for our needs which would have cascading impacts on other teams which is not what we wanted to accomplish. Also they are two separate use-cases 1️⃣  partial success or 2️⃣ fail fast
2. In terms of performance, having a long string of variableIds as query params impacts performance verses sending those variableIds in the body of the request Improves.  From what I read, query params are used for filtering (search/pagination or `?withComments`).  Since we are anticipating heavy parameter usage, we decided to send all this data in the body of a POST request. 

This is probably a question a lot of people will ask so I will add this to the doc

design/batch_retrieval_design.md
sigalsax: See comment above for answer

design/batch_retrieval_design.md
sigalsax: I think I explained this below but all secret values will be treated as individual entities. If a single secret fails to be retrieved, this will not impact the integrity of the request. I will add this statement in this Response section.

design/batch_retrieval_design.md
sigalsax: > in this case will the secrets that were found be returned in the response body?

Yes

> if so - isn't it wrong to return 404 but still return data?
Data meaning the error message if a 404 response was returned. Upon success, the value of the secret will be returned. Upon failure, the error message (Not Found) will be substituted for the the value, not the data of the secret.
Why is this wrong? Do you expect value to be `""` instead of `Not Found`?

design/batch_retrieval_design.md
sigalsax: and `value` would be `""` if there was an error?

design/batch_retrieval_design.md
sigalsax: I realize now that this is wrong. The response will still return a 200 OK because the request was handled even if there was a problem in getting the value of the secret (404 / 401)

design/batch_retrieval_design.md
sigalsax: yes would be correct. So ideally, 401 wouldnt be a possible response for batch right? because we are dealing with authentication right?

design/batch_retrieval_design.md
orenbm: i think that this is quite drunk. i would think that the difference should be in the url itself. both `/secrets` and `/secrets/batch` perform batch retrieval for secrets. 

i will add another comment with my thoughts on this.

design/batch_retrieval_design.md
orenbm: i still think that we can use the existing API. there are 2 differences between the two APIs:
1. the required variables are passed in the body instead of in a query param - this can be handled by reading the body _and_ reading the query param. we can decide if we want to merge the lists or reject a request that has both inputs but we can still use the existing one with backwards compatibility.
2. The response body is quite the same iirc, where the only difference is that we don't return only the value but also a `status` field. So the object is still a key-value json where the key is the variable-id. This response code will also change accordingly as the new API will not fail if only part of the variables will fail. This can be filtered by a query_param (e.g `partial`) that if present we will return the new API response. This is also backwards-compatible.

design/batch_retrieval_design.md
orenbm: please add a [link to the comment](https://github.com/cyberark/secrets-provider-for-k8s/pull/205/files#r477064995) in cases like this as it was not the comment above.

design/batch_retrieval_design.md
orenbm: so when will we return 403? only if all the secrets are forbidden? can you emphasize this in the table above?

orenbm: also - you return 404 if at least one is not found but 403 if all the secrets are not found. i think this is lacking consistency.

design/batch_retrieval_design.md
orenbm: just to be clear. a user may get a response with code 404 and the following body:
```
{
  "successSecret1":[
     {
       "status": 200
       "value": "superSecret"
     }
  ]
  "failSecret1":[
    {
      "status": 404
      "value": "NotFound"     
    }
  ]
}
```

it seems weird to me that i get a response body with secrets where the error code is 404.

orenbm: but this is also a reasonable response:
```
HTTP/1.1 200 OK
{
  "successSecret1":[
     {
       "status": 200
       "value": "superSecret"
     }
  ]
  "failSecret1":[
    {
      "status": 403
      "value": "Forbidden"     
    }
  ]
}
```

so 200 in case one is forbidden but 404 in case one is not found.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,176 @@
...
+```
+HTTP/1.1 200 OK
+
+{
+  "successSecret1":[
+     {
+       "status": 200
+       "value": "superSecret"
+     }
+  ]
```
orenbm, line 136: why do we need the list? can't it be
```
"successSecret1":{
   "status": 200
   "value": "superSecret"
}
```

design/batch_retrieval_design.md
sigalsax: @abrahamko ^

design/batch_retrieval_design.md
sigalsax: No no if we can parse the body of the request, the response will be 200 even if the sub-requests have failed with 401 / 404

design/batch_retrieval_design.md
sigalsax: See here: https://docs.microsoft.com/en-us/graph/json-batching#:~:text=The%20status%20code%20on%20a,requests%20inside%20the%20batch%20succeeded

Regardless of the outcome of the secrets that are being fetched, if we can parse the body of the response we return a 200 status code because the body was parsable and "a 200 status code on the batch response does not indicate that the individual requests inside the batch succeeded."

design/batch_retrieval_design.md
orenbm: then i'll ask again. when will we return 403 - when all the secrets are forbidden? when at least one of the secrets is forbidden?

can you state this in the table?

design/batch_retrieval_design.md
orenbm: so this goes for 403 as well or only for 404? the behaviour is inconsistent between the 2.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,176 @@
...
+    {"name": "var/path/anotherConjurSecret"},
+    ...
+  ]
+}
+```
+
+#### Headers
+
+```
+Content-Type: application/json
```
abrahamko, line 107: Missing authentication token header

design/batch_retrieval_design.md
abrahamko, line 146: I think the response body should be similar to the request, in case we want to add something to the response which is not specific to any of the variables:
```
{
  "variables": [
    {
      "name": "var/path/name1",
       "status": 200,
       "value": "superSecret"
    },
    {
      "name": "var/path/name2",
      "status": 404,
      "value": "Not Found"
    },
...
  ]
}
```

abrahamko (changes requested):
Good doc!
please see comments.

design/batch_retrieval_design.md
InbalZilberman: Why do we need the statuses for each variable?
Few questions regarding this:
1. I believe that 403 status is problematic from security aspect, no? I have permissions we need to 
2. Generally, is secret provider will act differently for specific statues? or is it only for supportability?

design/batch_retrieval_design.md
Tovli: Have you consider using filtering instead of a list of ids? so a user with 1,000,000 variables wouldn't need to send a massive body?
In that case you can have the filtering in the query string as it expect to be a rather short string

sigalsax: Can you explain a bit more about your idea above @Tovli Filtering on what quality of the variableID?

design/batch_retrieval_design.md
sigalsax: @InbalZilberman 
1️⃣  From our docs, I see that Batch can also [returns 403](https://docs.conjur.org/Latest/en/Content/Developer/Conjur_API_Batch_Retrieve.htm). I am not saying this is right behavior because we did it before. But it is important to note that only the nested, sub-requests return the error codes **NOT** the batch request itself. It is possible that each secret returns 403 error but the batch request itself will return 200 status OK. IMO since we are just an API endpoint, it is up to the צורך to decide if he wants to cascade that error to his application. @sashaCher what say you?

2️⃣ To know if we have failed or succeeded and on which exact secret. I think it gives nice insight into the status of each secret that was retrieved from Conjur. By adding `status` in the response, it makes it easy for the Secrets Provider to just extract the error code and percolate up that error, manipulating the result how we see fit (returning a log "Updating x out of y secrets", etc)

design/batch_retrieval_design.md
sigalsax: They are already separated like so:
### Response
## Batch Response
 # error codes (200 400)

## Individual secret responses
# error codes (200, 400, 401, 403, 404)

Was that not clear enough?

design/batch_retrieval_design.md
sigalsax: As we spoke on the phone, there are two different responses for this new batch endpoint. The overall response and the internal sub-responses

design/batch_retrieval_design.md
sigalsax: Updating

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,176 @@
...
+```
+POST /secrets/batch
+
+{
+  "variables": [
+    {"name": "var/path/conjurSecret"},
+    {"name": "var/path/anotherConjurSecret"},
+    ...
+  ]
+}
```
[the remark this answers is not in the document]
sigalsax: purposefully. One is the overall request and the other is a dissection of each part

design/batch_retrieval_design.md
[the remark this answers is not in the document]
sigalsax: great thanks!

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,176 @@
...
+```
+
+### Request Body
+
+The request body is a list of all variable IDs (paths to secrets in Conjur/DAP) that need to be retrieved. For example: 
+
+```
+{
+  "variables": [
```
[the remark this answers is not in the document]
sigalsax: It needs to be variables here no? because in conjur secrets are referenced to as variables

design/batch_retrieval_design.md
sigalsax: Changed the body to resemble @orenbm comment [here](https://github.com/cyberark/secrets-provider-for-k8s/pull/205#discussion_r477111213)
```
{
  "successSecret1":{
    "status": 200,
    "value": "superSecret"
  },
  "failSecret1":{
    "status": 403,
    "value": "Forbidden"     
  }
```

sigalsax: I can wrap it in variables object though

sigalsax: I think your suggestion makes sense because it is possible that we will add filtering as part of the query param

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,183 @@
...
+##### POST /secrets/batch
+
+This method retuns a list of secrets, their status code, and their value if an error did not occur. See difference between proposed API and existing API see [Proposed API vs Old API](#proposed-api-vs-old-api) below.
+
+```
+POST /secrets/batch
+
+{
+  "variables": [
+    {"name": "var/path/conjurSecret"},
```
sgnn7, line 82: I believe that we use "id" instead of "name" in most other places so if we're doing a new endpoint, we should probably fix that up.

design/batch_retrieval_design.md
sgnn7: I'd +1 to this. Maybe something like `/secrets/batch-v2` or `/secrets/batch/v2` would make more sense?

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,183 @@
...
+
+4. The `cyberark-secrets-provider-for-k8s` Job is completed.
+
+The new Batch Retrieval API will effect #2a. 
+
+### URI
+
+##### POST /secrets/batch
+
+This method retuns a list of secrets, their status code, and their value if an error did not occur. See difference between proposed API and existing API see [Proposed API vs Old API](#proposed-api-vs-old-api) below.
```
sgnn7, line 75: Given that it's now easy to brute-force multiple variables at a time, we probably need to implement mitigations for that.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,183 @@
...
+
+All secret values will be treated as individual entities. If a single secret fails to be retrieved, this will not impact the integrity of the request. Each secret has it's own response code.
+
+| Code             | Description                                                  |
+| ---------------- | ------------------------------------------------------------ |
+| 200 OK           | All secret values were retrieved successfully                |
+| 400 Bad Request  | Server cannot process request due to client error (invalid input parameters or logical validations) |
+| 401 Unauthorized | The request lacks valid authentication credentials           |
+| 403 Forbidden    | The authenticated user lacks the necessary privilege         |
+| 404 Not Found    | Variable does not exist or variable does not have secret value |
```
sgnn7, line 133: I don't think these messages are in line with standard JSON responses. I think more appropriate would be something like:

Worked OK:
```json
{
  "id": "foo/bar",
  "success": true,
  "value": "somevalue"
}
```

And for errors:
```json
{
  "id": "foo/bar",
  "success": false,
  "error": {
    "code": 404,
    "message": "Variable not found"
  }
}
```

Google uses this pattern quite a bit and you can find more info about it here: https://www.jsonrpc.org/specification#error_object

design/batch_retrieval_design.md
sigalsax: We might want to deprecate the old batch api. Does it still make sense to add the `v2`?

design/batch_retrieval_design.md
sigalsax: updating

design/batch_retrieval_design.md
sigalsax: Noted. Created a mitigation section

design/batch_retrieval_design.md
sigalsax: Do you think there is added value of having an extra field for success named `"code": 200` instead of or alongside `"success": true`?

sigalsax: So it would be:
```
  "variables": [
  	{
      "id": "var/path/name1",
      "code": 200,
      "value": "superSecret"
    },
    {
      "id": "var/path/name2",
      "error": {
        "code": 403,
        "message": "Forbidden"
      }
    },
...
```
where in-place of the `"success": true` we have error or no error.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,196 @@
...
+
+#### Batch Response
+
+If no Conjur/DAP secrets were able to be retrieved (no 200 code), the response will still return a 200 OK because the [request body is parsable]([https://docs.microsoft.com/en-us/graph/json-batching#:~:text=The%20status%20code%20on%20a,requests%20inside%20the%20batch%20succeeded](https://docs.microsoft.com/en-us/graph/json-batching#:~:text=The status code on a,requests inside the batch succeeded)) and request was handled.
+
+If an invalid request body was supplied, a 400 Bad Request will be returned
+
+| Code            | Description                                                  |
+| --------------- | ------------------------------------------------------------ |
+| 200 OK          | All secret values were retrieved successfully                |
```
nessiLahav, line 122: What about to separate the use cases of "All secret values were retrieved successfully" and to "partial successes" , [here](https://stackoverflow.com/questions/8472935/http-status-code-for-a-partial-successful-request) is an example to the same issue, where they used `207 Multi-Status` response code for partial success

sigalsax: What I gathered from [here](https://docs.microsoft.com/en-us/graph/json-batching#response-format) is that there are two types of status for the overall batch request 200 or 400. 
200 is always returned even if an internal response was a failure. 200 is returned if the body of the request is parsable. 
Do you still feel that a partial status, non-200 status code should be returned?

sigalsax: Also found some good explanations here from Arik's blog post https://conjurhq.slack.com/files/U5VEFK01G/F01680UV11C/blog_-_how__and_why__to_build_killer_bulk_apis_-_part_2.pdf

design/batch_retrieval_design.md
abrahamko: @Tovli I think filtering is an advanced feature and can be added later if we see fit.
Thanks for the suggestion.

design/batch_retrieval_design.md
abrahamko: @sigalsax I meant to create a different table for each.
Since Batch Responses are the API response code, and the Individual responses are in the response body, they are inherently different.
So they should be described differently.
The Batch Responses should be described in the API response code section, but the Individual responses should be described in the Response Body section. and a table for each for their relevant values and meaning.

design/batch_retrieval_design.md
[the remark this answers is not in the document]
abrahamko: Still, it doesn't look good. think of a way to improve it.

design/batch_retrieval_design.md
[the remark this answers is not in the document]
abrahamko: I think I got the definition of `variables` in Conjur.
variables are Conjur entities which hold secrets as their value.
They may contain other info such as attributes etc.
So, `secrets` API should return only the variables **values**, as correctly done by the existing `secrets` API.
But this proposed new API returns more info than just the values, so maybe it should not be called `secrets`.
Maybe it should be variables, because it can bring more info on the variables besides the secret (which is the variable value).

design/batch_retrieval_design.md
oburstein-hub: There was no explicit request from any customer to have additional format of batch retrieval  -and it is basically wrong to provide 2 different ways of doing things

design/batch_retrieval_design.md
oburstein-hub: do we have a v2 concept there for other REST APIs in conjur ?
If not - we should not add v2 only for one specific API. V2 is usually used when rewriting all APIs to some new concept

design/batch_retrieval_design.md
abrahamko: You have the code in the base in case of success, but in the error object in case of failure.
This is inconsistent.

design/batch_retrieval_design.md
oburstein-hub: I agree that @sashaCher should provide a guidance for it

oburstein-hub:
Nice document

design/batch_retrieval_design.md
oburstein-hub: I think its OK to return 200 in the body of the request in case of partial failure

design/batch_retrieval_design.md
sigalsax: do we not see value in returning the code if success in general?

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,216 @@
...
+
+4. The `cyberark-secrets-provider-for-k8s` Job is completed.
+
+The new Batch Retrieval API will effect #2a. 
+
+### URI
+
+##### POST /secrets/batch/v2
+
+This method retuns a list of secrets, their status code, and their value if an error did not occur. See difference between proposed API and existing API see [Proposed API vs Current API](#proposed-api-vs-current-api) below.
```
sigalsax, line 77: TODO remove "status code"
status code if error occurs

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,215 @@
...
 
 (**) This solution can be broken up in two: 1. Create a new API endpoint or 2. use existing one (return 409 response)
 
+*Decision:* We decided to go with #2 and create our own API. We decided to create a new batch retrieval API because of the following reasons:
+
+1. Adding onto the existing batch endpoint would lead to changing behaviors for our needs in the Secrets Provider which would have cascading impacts on other teams. 
+2. The two APIs (current and new) offer separate use-cases;  partial success (new) and fail fast (current)
+3. The current batch retrieval API accepts variable IDs as query parameters. We anticipate a long string of variableIDs to be passed in for secret retrieval and this impacts performance.
+4. Query parameters for HTTP GET requests are limited in length (2,048 characters) which is problematic if a customer has long secrets paths.
+5. The new route offers is more useful if we choose to expand its functionality in the future. By using POST instead of GET we can not only opt to use query parameters but also utilize the body of the request.
```
jonahx, line 54: According to [this](https://stackoverflow.com/a/983458), it looks like the latest relevant RFC does allow semantically meaningfully bodies in `GET` requests.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,215 @@
...
+
+|      | Solution                                                     | Pros                                                         | Cons                                               | Effort (T-shirt size) |
+| ---- | ------------------------------------------------------------ | ------------------------------------------------------------ | -------------------------------------------------- | --------------------- |
+| 1    | *Client side:* <br />For each K8s Secret, perform a new Batch request | - No server side changes / no breaking changes               | Load on server with extra calls (*)                | S, 5 days             |
+| 2    | *Server side:*<br />Update Batch retrieval to return list of variables **and** their response (success/failure) (**) | - Will help us during rotation for Milestone 2<br />- Best practice / straightforward design for batch endpoint | - Requires both client/server implementation<br /> | M, 10 days            |
+| 3    | Stay as is                                                   | No additional work needed                                    | - Bad UX<br />- Performance issues                 | -                     |
```
jonahx, line 42: > "Load on server with extra calls"

Do we have any data indicating that this would be a problem?  In _theory_ I understand it could, but it's worth making sure we aren't optimizing for a problem that won't exist in practice.  

To be clear I'm not saying the problem wouldn't exist... I'm just asking: How do we know?

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,215 @@
...
+
+### Open questions
+
+1. Do we want this duplicated API documented? At this stage this API endpoint will be for internal use only.
+2. Do we ultimately want to move to a single Batch Request API and depricate the existing one?
+
+### Future enhancements
+
+1. Add secret versions as a filtering query parameter so we can get all secrets after a certain version
+2. Fetch all secrets that the user/host has permissions to
```
jonahx, line 214: This caught my eye...

As I understand it (correct me if I'm wrong), a given host in Conjur will always be used for a specific application (running on k8s).  If this is the case, and since we already have to configure Conjur (via policy) with the knowledge of what secrets a host can access, I wonder if the whole idea here might be redundant in a way....

That is, could we just have a `/host-secrets` endpoint that gets everything at once, since it already knows all the secrets the application will need?  Then the secrets-provider-for-k8s just sends its host id, gets back a structure with all the secrets, and loops through them, hydrating the k8s secrets with the mapped keys?

I guess what I'm asking is: Are we duplicating the knowledge of what secrets an application needs?  Putting that knowledge in both Conjur and in this batch request?

design/batch_retrieval_design.md
sigalsax: Also a theory which doesn't seem too convincing but at this time, hasn't been tested so I will just remove that

design/batch_retrieval_design.md
sigalsax: Let me preface that this section is an accumulating of future intentions for this batch retrieval API endpoint that surpasses the work done in the secrets provider. We ultimately intend to make this endpoint forward facing and used across our teams. So for this point, I didn't think about this in the scope of the secrets provider.

> a given host in Conjur will always be used for a specific application (running on k8s)

A given host identity will be used by the secrets provider application container to fetch the secrets, not per customer application.

If we were to look at this idea in the scope of the secrets provider, it is not totally redundant because the customer may seek to get a secret from Conjur (by defining the path in its K8s Secret) that it actually doesn't have access to. Especially if they reuse a host for the secrets provider.

I am missing what the difference would be between the proposed batch extension idea (2️⃣ ) and `/host-secrets` here.

jonahx: Not totally sure I understood... Let me ask the question a different way:

Which piece of the system is responsible for holding the knowledge about which secrets an application has access to?  That is, I am a developer and write an application to be deployed on k8s needing secrets A, B, and C.  How does that information flow up the chain so that the k8s secrets provider:
1. Uses a host identity with access to those secrets.
2. Knows to request those secrets from Conjur and hydrate the corresponding k8 secrets values.

sigalsax: > Which piece of the system is responsible for holding the knowledge about which secrets an application has access to?

Conjur. All defined in policy

1. The user defines the secrets they need A B C in the configuration of the secrets provider. For Helm it is done [here](https://github.com/cyberark/secrets-provider-for-k8s/blob/master/helm/secrets-provider/values.yaml#L26).
2. We ask the customer to add `conjur-map` key to the K8s Secrets A B C which holds the name of the individual secrets and their path in conjur. For example
```
apiVersion: v1
kind: Secret
metadata:
  name: test-k8s-secret
type: Opaque
stringData:
  conjur-map: |-
    secret: secrets/someSortOfSecret
    anotherSecret: secrets/anotherSomeSortOfSecret
```
3. As part of this configuration we also ask them to supply the `environment.conjur.authnLogin` [here](https://github.com/cyberark/secrets-provider-for-k8s/blob/master/helm/secrets-provider/values.yaml#L35) which Conjur will use to determine if host used has access to the secrets defined in the conjur-map path (from above example)
4. If so, we successfully update the K8s Secrets defined for us the k8sSecret parameter from step 1 with their values from Conjur.

Does that make it more sense?

jonahx: Starting to understand, but still a couple more questions.  Re: the above, and also your earlier comment:

> A given host identity will be used by the secrets provider application container to fetch the secrets, not per customer application.

it looks like we have:

1. A host, with access to a bunch of secrets.
2. Applications, each of which has access to some subset of those secrets.
3. A k8s secrets config (like the example in step 2. above) which defines the Conjur -> K8s name mappings for those secrets.  

Questions:

Re: 3, is there just a single file for the host?  Is there anything stopping one app from accessing k8s secrets that have been hydrated for another app?  If not, that seems like a security concern.  We want access restricted by application, but have to settle for the granularity of host level access, where a host has the authority to get the secrets for multiple applications?  Am I understanding that right?  Or is there a separate authnLogin for each application?

design/batch_retrieval_design.md
sashaCher: `403` is acceptable error code **if and only if** the role has `READ` permission on given variable but has no `EXECUTE` one.
`404` is the right error code for all other use cases.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,352 @@
...
+| error.message | Response message of the error. See table below     |
+
+##### Individual secret responses
+
+The following are the possible error codes that can be returned for individual secrets.
+
+| Status/Message   | Description                                                  |
+| ---------------- | ------------------------------------------------------------ |
+| 400 Bad Request  | Server cannot process request due to client error (invalid input parameters or logical validations). For example: adding `<>` in path to secret in Conjur/DAP |
+| 401 Unauthorized | The request lacks valid authentication credentials           |
```
sashaCher, line 179: `401` is a result of invalid Access Token.
Access Token validation occurs once even before flow reaches a controller code (please fix me if I'm wrong).
So `401` is a return code of the whole API rather than an "Individual secret response".

design/batch_retrieval_design.md
sashaCher: How individual status codes are translated to the API status code is a functional decision.
From my perspective it should be 200 as far as there's a secret in response that has 200...

design/batch_retrieval_design.md
sigalsax: @oburstein-hub are you saying we recommend against creating an additional api endpoint?

design/batch_retrieval_design.md
sigalsax: changed to `POST /secrets` as discussed in the meeting

design/batch_retrieval_design.md
sigalsax: >  is there just a single file for the host?

As in Conjur policy file? If so, yes. The user can define the level of granularity / resource restrictions in the scope of the namespace, service account, deployment (not for application containers b/c they run as a Job), etc. If I choose to restrict the secrets provider by service account, that means I am restricting all communication with Conjur at the level of the service account. In other words, only a Secrets Provider Pod that runs with that service account (and defined in policy) can authenticate/fetch secrets from Conjur.

>  Is there anything stopping one app from accessing k8s secrets that have been hydrated for another app?

In K8s, the K8s Secrets granularity is at the namespace level so all applications have access to all K8s secrets within the same namespace. So to answer your question, no. Our thinking was that the security measures taken within K8s environments is up to K8s to control. We restrict access and update K8s secrets with Conjur secrets that the Conjur host has access to. The rest is up for K8s to decide. 
The value in the Secrets Provider is its enhancement UX. The Secrets Provider allows customers to still use the K8s secrets struct but their secret's lifecycle can be managed in one place, Conjur.
Also, it is known that K8s secrets are not 💯 the best option for security. If the customer wants more security, usually they would look to summon etc.

> We want access restricted by application, but have to settle for the granularity of host level access, where a host has the authority to get the secrets for multiple applications?

Yes. I was to clarify, the host is only for the secrets provider, not for customer applications. If the customer wants to restrict access a bit more, they can deploy multiple secrets providers (application container solution - aka the work being done in this Milestone) or give the secrets provider that sits in the init containers different hosts for their secrets provider with different restrictions. But again, once it leaves our hands, it is up to K8s to enforce least privilege.

> Or is there a separate authnLogin for each application?

authnLogin is not for each customer application, rather for each secrets provider pod.

design/batch_retrieval_design.md
sigalsax: You are correct. Good catch

design/batch_retrieval_design.md
oburstein-hub: I am saying that there should be only one endpoint exposed to customers

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,354 @@
...
+| 1    | Vanilla flow                                                 | Given I am an authenticated, privileged user/host<br />With proper permissions on requested secrets | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/anotherConjurSecret` in request body | - The JSON response should be (*see below* #1)<br />- Status response code should be `200 OK`<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 2    | User/Host authenticated but does not have `execute` privilege | Given I am a privileged, authenticated user/host<br />With proper permissions on ***one*** of requested secrets | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/anotherConjurSecret` in request body | - The JSON response should be (*see below* #2)<br />- Status response code should be `200 OK`<br />- Existing secret should have `200`. Non-existing secret should have 404 status error<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
```
eladkug, line 217: Non-existing secret should have 404 <-> The authenticated user lacks the necessary privilege 403

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,354 @@
...
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+    {
+    "id": "var/path/anotherConjurSecret",
+    "status": 403,
+    "value": "superSecret"
```
eladkug, line 259: should replace with:
"error": {
      "message": "Forbidden"
    }

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,354 @@
...
+{
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+    {
+    "id": "var/path/nonexistingSecret",
+    "status": 404,
```
eladkug, line 279: Should replace with :
"error": {
      "message": "Not Found"
 		}

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,354 @@
...
+{
+  "variables": [
+  {
+    "id": "var/path/conjurSecret",
+    "status": 200,
+    "value": "superSecret"
+  },
+    {
+    "id": "var/path/nonSecretValue",
+    "status": 404,
```
eladkug, line 300: Should replace with :
"error": {
      "message": "Not Found"
 		}

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,354 @@
...
+| 4    | User/Host authenticated but secret does not have value       | Given I am a privileged, authenticated user/host<br />With proper permissions<br />Requesting secret *value* does not exist | I send `POST /secrets` with valid JSON body with  `var/path/conjurSecret` and `var/path/nonSecretValue` in request body | - The JSON response should be (*see below #4)<br />- Status response code should be `200 OK`<br />- Existing secret should have `200`. Non-existing secret value should have `404 ` status error<br />- The proper audit record should appear for the whole request and each individual secret | Automated        |
+| 5    | Invalid request body                                         | Given I am a privileged, authenticated user/host<br />With proper permissions<br />Requesting secret in invalid format | I send `POST /secrets` with invalid JSON body with `var/path/conjurSecret` and  `var/path/<>` | - The JSON response should be (see below #5)<br />- Status response code should be `400 `<br />- The proper audit record should appear for the whole request and each individual secret<br />- **@Elad should the whole request fail?** | Automated        |
```
eladkug, line 220: Suppose to fail since we have problem with the request

oburstein-hub: agree

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,354 @@
...
+| 3    | Conjur server is not up-to-date                              | - Outdated Conjur server is running<br />- Authenticator is defined<br />- Secrets defined in Conjur and mapped in K8s Secrets<br />- All mandatory values are defined in `custom-values.yaml`<br />- Host defined to Secrets Provider in `custom-values.yaml` does not have access to one of the Conjur secrets<br />- Customer installs Secrets provider Chart | Secrets Provider runs as a Job | - Secrets Provider pod authenticates and attempts to fetch Conjur secrets<br /><br />- Secrets Provider Job runs to exhaustion<br />- Application fails to be deployed<br />- Verify logs #1/4 appear | Automated        |
```
eladkug, line 315: i think it suppose to be error and not warning

design/batch_retrieval_design.md
jonahx: Great explanation.  Thanks @sigalsax!

Final comments:

1. Returning to my original thought, now that I have a better understanding, it still seems to me like what we really want here is a single point, "get all my secrets as a batch," that an authenticated host can post to, especially since these hosts seem to exist for the sole purpose of identifying a k8s secrets provider.  It seems redundant that they have to specify the secrets they want in the request, when we can lookup and return all the secrets they own.

2. There may be some reason I'm not thinking of that prevents this suggestion, but.... it would be nice if even the mapping to k8s secrets was optional.  That is, by default, map to the same names that the secrets have in Conjur.  Now, I understand we may still want the list of secrets and their mapped values to exist within the k8s infrastructure in the file you pasted above:
```
apiVersion: v1
kind: Secret
metadata:
  name: test-k8s-secret
type: Opaque
stringData:
  conjur-map: |-
    secret: secrets/someSortOfSecret
    anotherSecret: secrets/anotherSomeSortOfSecret
```
since the admins managing k8s may be different from those managing Conjur.  However, we could still auto-generate the above file (using same name to same name mappings) from Conjur policy, in theory.  I think?  That way you'd have a single source of truth and cut down on the boilerplate files users have to write.

sigalsax: My only concern with your point in 1 is we would be basically be dumping the Conjur DB. _Especially_ if the Conjur host is used for the secrets provider and reused for some other entity that require a different set of secrets. The consequence of that would be what I like to call "over sharing"; returning more secrets than asked for.

Regarding your second point, can you send me an example of the flow you are envisioning. I didn't quite understand
>  by default, map to the same names that the secrets have in Conjur.

@abrahamko FYI

design/batch_retrieval_design.md
sigalsax: fail the _whole_ request? Adding #5 (:

design/batch_retrieval_design.md
orenbm: @jonahx i think that your suggestion for the host's secrets API is worth looking at but it can be added later. This design of @sigalsax and @abrahamko is for improving the current batch secrets retrieval (that will assist them in the secrets-provider). i think that both APIs can live together and have different value.

design/batch_retrieval_design.md
sigalsax: yes updating

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,378 @@
...
+The following are ideas for possible extension of the batch retrieval endpoint API and not necessary directly related to the Secrets Provider project.
+
+1. Add secret versions as a filtering query parameter so we can get all secrets after a certain version
+2. Fetch all secrets that the user/host has permissions to
+3. Fetch all secrets after `/var/path/*` that the user/host has permissions to
+
+### Delivery Plan
+
+- [ ] Fix flaky tests *2 days*
+- [ ] Implement new API endpoint in Conjur *10 days*
```
abrahamko, line 365: Shouldn't be more than 3 days.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,378 @@
...
+- [ ] Add integration tests in Conjur *4 days*
+- [ ] Update Golang SDK *4 days*
```
abrahamko, line 367: 2 days top, I think.

design/batch_retrieval_design.md
```diff
@@ -1,20 +1,378 @@
...
+- [ ] Implement tests in Golang SDK *3 days*
```
abrahamko, line 368: What kind of testing env does it have?
Does it have integration tests?
If not, let's do the testing in secrets provider, and implement few validation unit tests.
up to 2 days, IMO.

design/batch_retrieval_design.md
jonahx: @sigalsax @orenbm 

> My only concern with your point in 1 is we would be basically be dumping the Conjur DB. Especially if the Conjur host is used for the secrets provider and reused for some other entity that require a different set of secrets. 

I agree you don't want to overshare but shouldn't that problem be solved at the host level?  That is, to follow the principle of least privilege, you need to create hosts with properly scoped permissions.  If you create a profligate host and then reign it in with each request, you asking for trouble: every client/admin using the host has a new opportunity to fail and share something incorrectly.  Whereas if the host is created itself with just the permissions it needs, the task is done one time, by the Conjur admin, and then no one else can inadvertently mess it up.

> This design of @sigalsax and @abrahamko is for improving the current batch secrets retrieval (that will assist them in the secrets-provider)

Yeah I understand what I am saying is technically out of scope given the current proposal.  But I am suggesting that / asking if maybe the current proposal is not the ideal approach, both because of the repeated work and security concerns I mentioned.

design/batch_retrieval_design.md
sigalsax: When I give estimates I don't only think of the work done on our end but also try to factor in cross-team work (review from our own team and from conjur-core, etc). I don't know if that is the best way to estimate but I find that we haven't been meeting our estimations because of that factor. 
I agree that this shouldn't be 10 but 3 seems a little too low. Maybe 5 or 6

design/batch_retrieval_design.md
sigalsax: Again see my [above comment](https://github.com/cyberark/secrets-provider-for-k8s/pull/205#discussion_r484022823). It still applies

design/batch_retrieval_design.md
sigalsax: looks like unit tests https://github.com/cyberark/conjur-api-go/blob/master/conjurapi/variable_test.go

design/batch_retrieval_design.md
sigalsax: updating the doc to break it up in terms of implementation and unit tests, 2 days for each

design/batch_retrieval_design.md
sigalsax: agreed. Updating doc

design/batch_retrieval_design.md
sigalsax: As discussed in a private channel we will split implementation and conjur-core reviews into two separate tickets

design/batch_retrieval_design.md
```diff
@@ -288,7 +288,31 @@ The following integration tests will be implemented in Secrets Provider for K8s
 
 ## Performance tests
 
---
+The following performance tests will be run using *Jmeter*.
```
sigalsax, line 291: @abrahamko why are we falling back to Jmeter? Does the Conjur env already exist in Jmeter?

design/batch_retrieval_design.md
```diff
@@ -288,7 +288,31 @@ The following integration tests will be implemented in Secrets Provider for K8s
...
+These tests will be run for both the new API and the existing one, and compare the results between them.
+
+#### Benchmark
+
+| Section     | Parameter                      | Value          |
+| ----------- | ------------------------------ | -------------- |
+| Conjur      | \# of standbys                 | 2              |
+| Conjur      | \# of followers                | 1              |
+| Environment | Kubernetes                     | OpenShift 3.11 |
```
sigalsax, line 300: Why is Parameter Kubernetes and the Value Openshift here? Should it be Environment with both Openshift and Kubernetes values?

design/batch_retrieval_design.md
```diff
@@ -288,7 +288,31 @@ The following integration tests will be implemented in Secrets Provider for K8s
...
+| Environment | \# of cores                    | 16             |
+| Environment | \# of nodes                    | 1              |
+| Test        | \# of requesting threads       | 20             |
+| Test        | \# of variables                | 1025           |
+| Test        | \# of variabels per policy     | 5              |
+| Test        | \# of policies                 | 205            |
+| Test        | Variable ID length             | 50             |
+| Test        | Variable value length (secret) | 50             |
+
```
sigalsax, line 309: @abrahamko do you plan to add a benchmark chart for the Secrets Provider performance tests? I am filling the SP performance test flows. For SP will we also be using Jmeter? How many SP will be deployed?

design/batch_retrieval_design.md
```diff
@@ -288,7 +288,31 @@ The following integration tests will be implemented in Secrets Provider for K8s
...
+| ---- | ------------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ---------------- |
+| 1    | Response time | - Conjur is running<br />- Host is defined<br />- 1024 Secrets are defined<br /> | For X=1, X<=1024, X=X\*2:<br />    Concurrently retrieve X secrets for 5 minutes, measure average response time.<br />    Repeat each step 3 times.<br /> | Calculate the average response time between the 3 cycles for each step.<br />Plot the average response time per number of secrets provider. | Manual           |
+| 2    | Throughput    | - Conjur is running<br />- Host is defined<br />- 1024 Secrets are defined<br /> | For X=1, X<=1024, X=X\*2:<br />    Concurrently retrieve X secrets for 5 minutes, measure throughput.<br />    Repeat each step 3 times.<br /> | Calculate the throughput between the 3 cycles for each step.<br />Plot the throughput per number of secrets provider. | Manual           |
```
sigalsax, line 315: so we will be testing 1, 1024, 512 secrets in the body of batch request?

design/batch_retrieval_design.md
abrahamko: We do not "fall back" to jmeter.
Jmeter is the tool to use to run the tests and analyze the results.
The conjur should still be deployed separately.

design/batch_retrieval_design.md
abrahamko: OpenShift is a type of Kubernetes deployment.

design/batch_retrieval_design.md
abrahamko: For SP, we should run the same performance tests that were defined for SLA.

design/batch_retrieval_design.md
abrahamko: We will be testing 1,2,4,8,16,32,64,128,256,512,1024 secrets in the body. each as its own test, and plot the results.

abrahamko (approved):

design/batch_retrieval_design.md
sigalsax: this should be stated in the file either copy and pasted or if the file exists provide the link to the file

design/batch_retrieval_design.md
sigalsax: I understand that.. but that is not what I am saying. I am asking if we are testing on both k8s and openshift. If so, that is not clear and I ask you list them both

design/batch_retrieval_design.md
sigalsax: When I say fall back I am saying that this is not how we did it for SP so I ask why we chose Jmeter. Suggesting to write a one-liner explaining this choice. For example "We chose to use Jmeter because the test environment is already setup and is the easiest to get end-to-end"

design/batch_retrieval_design.md
sigalsax: That is not clear here. It seems to me that we are only running with 3 types of X values (1, 1024, 512 secrets). Please highlight that. It is imperative we are clear about what we are testing otherwise it will lead to misunderstandings and delay

sigalsax:
reopening, close accidentally

PR #206, opened 2020-08-30
Update retry defaults after concern on init container solution

sigalsax:
### What does this PR do?
During a team meeting it was decided to update the retry mechanism defaults to avoid impacting the current init container solution UX.

Connected to #207 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
+++ b/deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
@@ -18,8 +18,8 @@ pushd ../../
   # A parameter that will force a failure
   export CONJUR_AUTHN_URL="https://conjur-follower.${CONJUR_NAMESPACE_NAME}.svc.cluster.local/api/authn-k8s/${AUTHENTICATOR_ID}xyz"  # Configure retry mechanism with overriding defaults
 
-  DEFAULT_RETRY_INTERVAL_SEC=30
-  DEFAULT_RETRY_COUNT_LIMIT=3
+  DEFAULT_RETRY_INTERVAL_SEC=1
+  DEFAULT_RETRY_COUNT_LIMIT=5
 
   fill_helm_chart_test_image
   fill_helm_chart_no_override_defaults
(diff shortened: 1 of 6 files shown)


orenbm (approved):
LGTM! Do we want to release a new version with this?

sigalsax:
> LGTM! Do we want to release a new version with this?

@orenbm 
Not yet. I just updated the go.sum because I was failing on dependencies. After docs are released we will be pushing a release (:

deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
```diff
@@ -36,18 +36,18 @@ $cli_with_timeout "logs $pod_name | grep 'CSPFK010E Failed to authenticate'"
...
 
 duration=$(( SECONDS - start ))
 # Since we are testing retry in scripts we must determine an acceptable range that retry should have taken place
 # If the duration falls within that range, then we can determine the retry mechanism works as expected
-retryIntervalMin=`echo "scale=2; $DEFAULT_RETRY_INTERVAL_SEC/100*80" | bc | cut -d "." -f 1 | cut -d "," -f 1`
-retryIntervalMax=`echo "scale=2; $DEFAULT_RETRY_INTERVAL_SEC/100*120" | bc | cut -d "." -f 1 | cut -d "," -f 1`
-if (( $duration >= $retryIntervalMin && $duration <= $retryIntervalMax )); then
-  echo 0
+retryIntervalMin=`echo "scale=3; 1/100*80" | bc -l | awk '{print ($0-int($0)<0.499)?int($0):int($0)+1}'`
+retryIntervalMax=`echo "scale=3; 1/100*120" | bc -l | awk '{print ($0-int($0)<0.499)?int($0):int($0)+1}'`
```
eladkug, line 46: why not using $RETRY_INTERVAL_SEC/100...

eladkug (commented):
We changed 7 different files to update those variables, maybe its better to put it in global place and pull it from there?

deploy/utils.sh
```diff
@@ -219,8 +219,8 @@ fill_helm_chart() {
     -e "s#{{ TAG }}# ${TAG:-"latest"}#g" \
     -e "s#{{ LABELS }}# ${LABELS:-"app: test-helm"}#g" \
     -e "s#{{ DEBUG }}# ${DEBUG:-"false"}#g" \
-    -e "s#{{ RETRY_COUNT_LIMIT }}# ${RETRY_COUNT_LIMIT:-"3"}#g" \
-    -e "s#{{ RETRY_INTERVAL_SEC }}# ${RETRY_INTERVAL_SEC:-"30"}#g" \
+    -e "s#{{ RETRY_COUNT_LIMIT }}# ${RETRY_COUNT_LIMIT:-"5"}#g" \
```
abrahamko, line 222: Why not defining the defaults at `utils.sh` and reuse it wherever you need?

pkg/utils/limited_backoff_test.go
```diff
@@ -13,7 +13,7 @@ func TestLimitedBackOff(t *testing.T) {
 	Convey("Subject: Using limited backoff", t, func() {
 		const (
 			interval   = time.Second
-			retryLimit = 3
```
abrahamko, line 16: revert this. It has nothing to do with the defaults.

abrahamko (commented):
just a small fix and you're good to go.

deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
sigalsax: great catch

deploy/utils.sh
sigalsax: It is a bit more complicated than that because the existence of the count limit and retry sec is conditional so for me to manipulate them, they need to exist in the environment. So I needed a way to mock defaults because they aren't accessible to me in the environment. 
The places that set them as  `DEFAULT_RETRY_COUNT_LIMIT` in test 25 for example, do not use this function (fill_helm_chart and instead, `fill_helm_chart_no_override_defaults`). `fill_helm_chart_no_override_defaults` is to check that the defaults are taken. Fill_helm_chart function is used to override those defaults

Issue #218, opened 2020-09-09
Failed to find file  during the test(TEST_ID_25)


eladkug:
During some of the tests we see the following error :
Error: open ../helm/secrets-provider/ci/take-default-test-values.yaml: no such file or directory 
[link](https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--secrets-provider-for-k8s/detail/master/513/pipeline/71) [Logs](https://app.zenhub.com/files/206022444/330e6188-c15d-4d69-a033-87d660cfe032/download)

eladkug:
Fixed as part of this [ticket](https://github.com/cyberark/secrets-provider-for-k8s/pull/220)

PR #220, opened 2020-09-13
R&D Boost - Flaky tests

sigalsax:
We have been experiencing a number of failings across multiple Secrets Provider pipeline runs in Jenkins

This PR fixes:
1. Helm File existence #218 
2. SSL certificate existence upon loading

### DIFF
--- a/deploy/test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
+++ b/deploy/test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
@@ -9,7 +9,8 @@ echo "Create test-env pod. SECRETS_DESTINATION is with invalid value 'incorrect_
 export SECRETS_DESTINATION_KEY_VALUE=$KEY_VALUE_NOT_EXIST
 deploy_init_env
 
-pod_name=$(cli_get_pods_test_env | awk '{print $1}')
+$cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers | wc -l | tr -d ' ' | grep '^1$'"
+pod_name=$($cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers" | awk '{print $1}')
 
 echo "Expecting secrets provider to fail with error 'CSPFK004E Environment variable 'SECRETS_DESTINATION' must be provided'"
 $cli_with_timeout "logs $pod_name -c cyberark-secrets-provider-for-k8s | grep CSPFK004E"
--- a/deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
+++ b/deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
@@ -28,8 +28,8 @@ pushd ../../
 
   fill_helm_chart_no_override_defaults
 
-  helm install -f "../helm/secrets-provider/ci/take-default-test-values.yaml" \
-    -f "../helm/secrets-provider/ci/take-image-values.yaml" \
+  helm install -f "../helm/secrets-provider/ci/take-default-test-values-$UNIQUE_TEST_ID.yaml" \
+    -f "../helm/secrets-provider/ci/take-image-values-$UNIQUE_TEST_ID.yaml" \
     secrets-provider ../helm/secrets-provider \
     --set-file environment.conjur.sslCertificate.value="test/test_cases/conjur.pem"
 popd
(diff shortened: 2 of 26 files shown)


deploy/test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
```diff
@@ -9,6 +9,7 @@ echo "Create test-env pod. SECRETS_DESTINATION is with invalid value 'incorrect_
 export SECRETS_DESTINATION_KEY_VALUE=$KEY_VALUE_NOT_EXIST
 deploy_init_env
 
+cli_get_pods_test_env
```
eladkug, line 12: i understand we have some failures because of this, but we need to find better way.

deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
```diff
@@ -22,29 +22,28 @@ pushd ../../
...
-# This sleep allows all logs to arrive to avoid platform failures
-sleep 5
-
 # Find initial authentication error that should trigger the retry
+SPACER=1
 $cli_with_timeout "logs $pod_name | grep 'CSPFK010E Failed to authenticate'"
-
-# Start the timer for retry interval
-start=$SECONDS
+failure_time=$($cli_without_timeout logs $pod_name | grep 'CSPFK010E Failed to authenticate' | head -1 | awk '{ print $3 }' | awk -F: '{ print ($1 * 3600) + ($2 * 60) + $3 }')
```
eladkug, line 40: This is much better and will give us better results !

deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
```diff
@@ -22,29 +22,28 @@ pushd ../../
...
 
 echo "Expecting Secrets Provider retry configurations to take defaults RETRY_INTERVAL_SEC $DEFAULT_RETRY_INTERVAL_SEC and RETRY_COUNT_LIMIT $DEFAULT_RETRY_COUNT_LIMIT"
 $cli_with_timeout "logs $pod_name | grep 'CSPFK010I Updating Kubernetes Secrets: 1 retries out of $DEFAULT_RETRY_COUNT_LIMIT'"
+retry_time=$($cli_without_timeout logs $pod_name | grep "CSPFK010I Updating Kubernetes Secrets: 1 retries out of $DEFAULT_RETRY_COUNT_LIMIT" | head -1 | awk '{ print $3 }' | awk -F: '{ print ($1 * 3600) + ($2 * 60) + $3 }')
```
eladkug, line 44: can't we run it in one line which contain cli_with_timeout and will update the variable? like we have for example in test_id_18:
`pod_name=$($cli_with_timeout get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers | awk '{print $1}' | head -1)`

deploy/utils.sh
```diff
@@ -28,9 +27,11 @@ wait_for_it() {
 
 if [ $PLATFORM = "kubernetes" ]; then
     cli_with_timeout="wait_for_it 600 kubectl"
+    cli_with_timeout_job="wait_for_it 200 kubectl"
```
eladkug, line 30: i think we need better solution for this by giving the time as a value **but** since i dont want it will be part of this PR(It will require a lot of file changes), please open a ticket and we will handle it better later.

deploy/test/test_cases/TEST_ID_10_SECRETS_DESTINATION_not_exist.sh
```diff
@@ -9,7 +9,8 @@ echo "Create test-env pod. SECRETS_DESTINATION is with invalid value 'incorrect_
 export SECRETS_DESTINATION_KEY_VALUE=$KEY_VALUE_NOT_EXIST
 deploy_init_env
 
-pod_name=$(cli_get_pods_test_env | awk '{print $1}')
+$cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers | wc -l | tr -d ' ' | grep '^1$'"
+pod_name=$($cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers" | awk '{print $1}')
```
eladkug, line 13: We have to run it twice ? Did we try to run the second line only? if it didnt work then why we need to run the "get pods command" twice with timeout.. 
The second one can be without timeout.

deploy/test/test_cases/TEST_ID_25_helm_default_retry_successful.sh
sigalsax: it does not account for the timeout. I rather leave it as is

eladkug (approved):
LGTM. 
Moving to BDD will improve some of the code and will give us much more flexibility that we dont have today.

Issue #221, opened 2020-09-13
Use logger from authn k8s client


abrahamko:
## Is your feature request related to a problem? Please describe.

I would like to use the same logger as in authn client.
All logger features i secrets provider were moved to authn client, and now the secrets provider logger should be removed.
This allows benefiting from the authn client logger that adds milliseconds precision to log timestamp.

## Describe the solution you would like

Use [logger from Authn client](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/pkg/log/logger.go)

PR #222, opened 2020-09-13
Add microseconds to logs time

abrahamko:
### What does this PR do?
Since secrets provider operation can be finished in less than a second, logs time resolution of seconds is not enough to understand steps duration.
Adding microseconds to logs time resolution allow better investigation of steps duration.

Connected to #221

#### Change log
- [x] The CHANGELOG has been updated

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Improve log messages to include also microseconds instead of only seconds.
 
 ## [1.1.0] - 2020-08-20
 ### Added
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -8,8 +8,8 @@ import (
 	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
 )
 
-var stdoutLogger = log.New(os.Stdout, "INFO:  ", log.LUTC|log.Ldate|log.Ltime|log.Lshortfile)
-var errorLogger = log.New(os.Stderr, "ERROR: ", log.LUTC|log.Ldate|log.Ltime|log.Lshortfile)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Add nanoseconds to the log
```
orenbm, line 9: `Log messages are now showing microseconds`. (@sigalsax ?)

please also add a link to an issue (and create one if there isn't one.

pkg/log/logger.go
```diff
@@ -8,8 +8,8 @@ import (
 	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
 )
 
-var stdoutLogger = log.New(os.Stdout, "INFO:  ", log.LUTC|log.Ldate|log.Ltime|log.Lshortfile)
-var errorLogger = log.New(os.Stderr, "ERROR: ", log.LUTC|log.Ldate|log.Ltime|log.Lshortfile)
+var stdoutLogger = log.New(os.Stdout, "INFO:  ", log.LUTC|log.Ldate|log.Ltime|log.Lmicroseconds|log.Lshortfile)
```
sigalsax, line 11: @abrahamko if we do this we might need to change test 25 and 26 which relies on time between failure and retry. Please ensure that these tests still pass.

pkg/log/logger.go
```diff
@@ -8,8 +8,8 @@ import (
...
+var errorLogger = log.New(os.Stderr, "ERROR: ", log.LUTC|log.Ldate|log.Ltime|log.Lmicroseconds|log.Lshortfile)
```
sigalsax, line 12: @abrahamko Can you attach a picture (before/after) to the PR to see what this looks like

CHANGELOG.md
sigalsax: @abrahamko @orenbm `Log messages are now showing microseconds to allow for clarity and easier troubleshooting` 

IMO we should also add the motivation behind each CHANGELOG entry not just _what_ was changed

sigalsax:
@abrahamko do you intend for this to be merge before our release?

Issue #233, opened 2020-09-24
Secrets Provider - Release automation


sigalsax:
After conversations with Geri on the SP automation we came up with the following items to implement

- [x] Every green master build on Jenkins will trigger an automated push to an edge tag
- [x] Automate the packaging of helm charts and save the artifact as part of Jenkins pipeline.
- [x] Update README to outline the types of builds/releases we push to Dockerhub and add instructions for taking release artifacts and open PR in helm-charts repo
- [ ] Talk to infra about how we might be able to automate the pushing of our Jenkins artifacts to another repo (helm-chart)

PR #234, opened 2020-09-24
Add release automation

sigalsax:
This PR is part of a bigger effort in automating our release process
- Package helm chart on every build
- Add ability to push an edge tag on every master build
- Update tag and release process in docs

See generated helm package here: https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--secrets-provider-for-k8s/detail/automate-release/31/artifacts

Connected to #233 

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -186,13 +186,6 @@ follow the instructions in this section.
     1. [Version file](pkg/secrets/version.go)
     1. [Chart version](helm/secrets-provider/Chart.yaml)
     1. [Default deployed version](helm/secrets-provider/values.yaml)
-    
-1. Create a Helm package by running the following command from the repo root: `helm package helm/secrets-provider`.
-   The Helm package will be saved to the current folder and will resemble `secrets-provider-<version>.tgz`.
-1. Clone the repo [helm-charts](https://github.com/cyberark/helm-charts) and do the following:
-    1. Move the Helm package file created in the previous step to the *docs* folder in the `helm-charts` repo.
-    1. Go to the `helm-charts` repo root folder and execute the `reindex.sh` script file located there.
-    1. Create a PR with those changes.
 1. Review the git log and ensure the [changelog](CHANGELOG.md) contains all
    relevant recent changes with references to GitHub issues or PRs, if possible.
 1. Review the changes since the last tag, and if the dependencies have changed
@@ -210,6 +203,14 @@ follow the instructions in this section.
 1. Push the tag: `git push vx.y.z` (or `git push origin vx.y.z` if you are working
    from your local machine).
 
+### Push Helm package
+1. Every build packages the Secrets Provider Helm chart for us. The package can be found under the 'Artifacts' tab of the Jenkins build and will resemble `secrets-provider-<version>.tgz`. 
+Navigate to the 'Artifacts' tab of the _tagged version_ build and save this file. You will need it for the next step.
+1. Clone the repo [helm-charts](https://github.com/cyberark/helm-charts) and do the following:
+    1. Move the Helm package file created in the previous step to the *docs* folder in the `helm-charts` repo.
+    1. Go to the `helm-charts` repo root folder and execute the `reindex.sh` script file located there.
+    1. Create a PR with those changes.
+    
 ### Publish the git release
 1. In the GitHub UI, create a release from the new tag and copy the change log
 for the new version into the GitHub release description. The Jenkins pipeline 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,23 +79,38 @@ pipeline {
       }
     }
 
-    stage('Publish client Docker image') {
-      steps {
-        withCredentials(
-          [
-            usernamePassword(
-              credentialsId: 'conjur-jenkins-api',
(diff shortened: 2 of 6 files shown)


Jenkinsfile
```diff
@@ -79,29 +79,49 @@ pipeline {
...
-                git fetch --tags
-                export GIT_DESCRIPTION=$(git describe --tags)
-                summon ./bin/publish
-            '''
+    stage('Release') {
+      parallel {
+        stage('Push Images') {
+          steps {
+            script {
+              BRANCH_NAME=env.BRANCH_NAME
```
hughsaunders, line 87: I'm not sure what this for, may as well use ${env.BRANCH_NAME} on L102?

bootstrap.env
```diff
@@ -1,4 +1,4 @@
-export UNIQUE_TEST_ID="$(uuidgen | tr "[:upper:]" "[:lower:]" | head -c 10)"
+export UNIQUE_TEST_ID="$(cat UNIQUE_TEST_ID 2>/dev/null || uuidgen | tr "[:upper:]" "[:lower:]" | head -c 10|tee UNIQUE_TEST_ID)"
```
hughsaunders, line 1: I know its my fail but all the other pipes have spaces around them.
```suggestion
export UNIQUE_TEST_ID="$(cat UNIQUE_TEST_ID 2>/dev/null || uuidgen | tr "[:upper:]" "[:lower:]" | head -c 10 | tee UNIQUE_TEST_ID)"
```

hughsaunders (commented):
Would be good to see a test of the package/release stages, maybe comment out then `when` clause for a test?

hughsaunders:
I missed this build which does test the package stage: https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--secrets-provider-for-k8s/detail/automate-release/31/pipeline/39

Jenkinsfile
sigalsax: I need the BRANCH_NAME inside publish to determine when to push our edge tag. We want to only push when on master so I need the `BRANCH_NAME=env.BRANCH_NAME` from the script and output the result. I tried via scripts but I would also get HEAD instead of the actual branch

CONTRIBUTING.md
```diff
@@ -210,6 +203,14 @@ follow the instructions in this section.
 1. Push the tag: `git push vx.y.z` (or `git push origin vx.y.z` if you are working
    from your local machine).
 
+### Push Helm package
+1. The tagged build will package the Secrets Provider Helm chart for us. The package will appear under the 'Artifacts' tab of the Jenkins build and will resemble `secrets-provider-<version>.tgz`. 
+Save this file you will need it for the next step.
```
eladkug, line 208: comma is missing i think.
Save this file **,** you will need it for the next step.

bin/publish
```diff
@@ -43,4 +43,8 @@ if [ "$GIT_DESCRIPTION" = "v${VERSION}" ]; then
     echo 'Failed to log in to scan.connect.redhat.com'
     exit 1
   fi
+elif [ "$BRANCH_NAME" = "master" ]; then
```
eladkug, line 46: Its not relevant for redhat registry?

sigalsax: Good point but I don't think so. I don't know if we would want to be pushing edge tags to RH because the publishing of tags is not an automated process. Our current policy for RH registry is that someone needs to push "publish" for tags to be public facing

bootstrap.env
sigalsax: No longer relevant because I removed this logic. Thank you though <3

CONTRIBUTING.md
sigalsax: great catch! I added a "because"

ci/jenkins_build
```diff
@@ -0,0 +1,11 @@
...
+set -euo pipefail
+
+source bin/build_utils
+helm_version=3.3.0
+
+docker run --rm \
+  -v $PWD/helm/secrets-provider:/root/helm/secrets-provider \
+  -v $PWD/helm-artifacts/:/root/helm-artifacts \
+  --workdir /root/helm-artifacts \
+   alpine/helm:${helm_version} package ../helm/secrets-provider
```
eladkug, line 11: need to align

sigalsax: should be fixed now

eranha (approved):
LGTM, well done.

eladkug (approved):
LGTM

ci/jenkins_build
```diff
@@ -0,0 +1,11 @@
...
+set -euo pipefail
+
+source bin/build_utils
+helm_version=3.3.0
+
+docker run --rm \
+  -v $PWD/helm/secrets-provider:/root/helm/secrets-provider \
+  -v $PWD/helm-artifacts/:/root/helm-artifacts \
+  --workdir /root/helm-artifacts \
+  alpine/helm:${helm_version} package ../helm/secrets-provider
```
abrahamko, line 11: that's the stuff!
😉

PR #243, opened 2020-10-02
Add release solution design

sigalsax:
### What does this PR do?
Add solution design for automated release

Connected to #233 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/design/automate_release_design.md
+++ b/design/automate_release_design.md
@@ -1,0 +1,212 @@
+# Solution Design - Secrets Provider Release
+
+## Table of Contents
+
+[//]: # "You can use this tool to generate a TOC - https://ecotrust-canada.github.io/markdown-toc/"
+
+## Glossary
+
+[//]: # "Describe terms that will be used throughout the design"
+[//]: # "You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#"
+
+| **Term**                                                     | **Description**                                              |
+| ------------------------------------------------------------ | ------------------------------------------------------------ |
+| [Github actions](https://docs.github.com/en/free-pro-team@latest/actions) | CI/CD platform that is event-driven to help automate task and execute workflows in repositories |
+| [repostiory_dispatch](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows)  webhook event | a webhook event used when you want to trigger a workflow for activity that happens outside of Github |
+
+## Useful links
+
+[//]: # "Add links that may be useful for the reader"
+
+Release documentation
+
+- [Secrets Provider release documentation](https://github.com/cyberark/secrets-provider-for-k8s/blob/master/CONTRIBUTING.md#releases)
+
+Open issues:
+
+- [Secrets Provider - Release automation](https://app.zenhub.com/workspaces/palmtree-5d99d900491c060001c85cba/issues/cyberark/secrets-provider-for-k8s/233)
+
+- [Secrets Provider - Github action for dispatching events](https://github.com/cyberark/secrets-provider-for-k8s/issues/239)
+
+- [Community - Guidelines and tools for releasing artifacts to separate suite components](https://github.com/cyberark/community/issues/77)
+
+PR:
+
+- https://github.com/cyberark/secrets-provider-for-k8s/pulls
+
+## Background
+
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
+
+The Secrets Provider release process has become a long and heavily manual process. This has become more evident with the introduction of Helm to the project which now adds additional steps to the release process. 
+
+Goals for this effort:
+
+1. Automate the parts we can.
+2. Knowledge sharing by providing guidelines for event-driven actions across multiple repositories.
+3. Push 'edge' release on every master build for easier testing and to better align ourselves with other open source projects
+
+## Solution
+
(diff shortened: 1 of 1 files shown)


design/automate_release_design.md
```diff
@@ -0,0 +1,204 @@
...
+
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
+
+The Secrets Provider release process has become a long and heavily manual process. This has become more evident with the introduction of Helm to the project which now adds additional steps to the release process. 
+
+Goals for this effort:
+
+1. Automate the parts we can.
+2. Knowledge sharing by providing guidelines for event-driven actions across multiple repositories.
+3. Push tag on every master build for easier testing and to better align ourselves with other open source projects
```
orenbm, line 46: it may be described in the rest of the doc but in case not - do we want also to add a tar with the image to the release? I created [this issue](https://github.com/cyberark/conjur-authn-k8s-client/issues/168) for the authn-client which we can do for this repo as well

orenbm: also - you don't need to actually list the numbers in `md` files. instead, you can just use `1` and the UI will show numbers. It can help for adding new entries in the middle without the need to change the numbers

design/automate_release_design.md
```diff
@@ -0,0 +1,204 @@
...
+
+As part of our Secrets Provider for K8s [release process](https://github.com/cyberark/secrets-provider-for-k8s/blob/master/CONTRIBUTING.md#releases) we need to manually package our Secrets Provider Helm Chart and open a PR attaching the artifact in our [cyberark/helm-charts](cyberark/helm-charts) repository. This process can be automated by introducing Github actions into our workflows and will be addressed in the below solution.
+
+The solution will be twofold and impact both the `cyberark/secrets-provider-for-k8s` Jenkins pipeline and `cyberark/helm-chart` repository. The flow at a high-level is as follows:
+
+`cyberark/secrets-provider-for-k8s` Jenkins pipeline
+
+1. Package the Secrets Provider Helm chart as part of Jenkins pipeline
+2. Dispatch an `version-release` event by sending a curl request with Helm package as part of request
```
orenbm, line 59: an -> a

orenbm: will the secrets-provider pipeline know if the `cyberark/helm-charts` failed its flow?

design/automate_release_design.md
```diff
@@ -0,0 +1,204 @@
...
+            BRANCH_PREFIX: "release-/"
+            PULL_REQUEST_FROM_BRANCH: "release-v110"
+            PULL_REQUEST_BRANCH: master
+            PULL_REQUEST_TITLE: "Push Helm chart package for new version release"
+            PULL_REQUEST_ASSIGNEES: sigalsax     	
+```
+
+### Push `edge` release on successful master build
+
+Another key addition part of the release automation effort is pushing an `edge` release to Dockerhub on every green master build. `edge` releases are intended for frequent pushes of content and used for closely monitoring development. This tag should not be given to our customer's to use in production.
```
orenbm, line 153: when will we consume this image? where will it be used?

design/automate_release_design.md
```diff
@@ -0,0 +1,204 @@
...
+### Additions to `cyberark/helm-chart` Github repository
+
+In the `cyberark/helm-chart` repository we will need to create a Github workflow that listens for `version-release` events. Once received this will trigger a series of actions in response. The following is an incomplete Github workflow that will be added to the `cyberark/helm-chart` repository:
+
+```pseudocode
+name: Push Helm package workflow
+
+on:
+  repository_dispatch:
+    types: [version-release]
```
izgeri, line 127: we will need this in other projects too (eg for `homebrew-tools`). if `version-release` events are the kind of event that the `helm-charts` repo is listening for, should these instead be  called `helm-version-release` or `helm-chart-release`, so that the `helm-charts` project _knows_ its an event that it should be listening for?

to make this clearer, if I release secretless and want to dispatch a `version-release` event to update `homebrew-tools`, is the `helm-charts` repo going to also catch that event and try to do something with it, even though there is no helm chart update needed?

design/automate_release_design.md
sigalsax: @orenbm This is a neat idea thanks for suggesting! Why do we add a tar for the image in a release btw? I don't remember doing this in our other SP releases. Is this standard?

design/automate_release_design.md
sigalsax: I don't think it is smart enough yet to know but this can be something that can be explored. Isn't it up to the permission who is reviewing to approval or fail?

If the helm-chart repo has a failure what do you think the logical action would be? Just failing the tagged release build?

design/automate_release_design.md
sigalsax: We will consume the image when we want to test the latest features without pushing a release. It can also be given to SEs for early testing and it can be used by @eladkug to have a reliable image without having to manually build it. In our experience this leads to conflicting images and extra work differentiating which image is the most recent build

design/automate_release_design.md
orenbm: iirc we had a case where the customer wanted to get the code base of the image.

design/automate_release_design.md
sigalsax: I agree on the first part that the event should be helm-specific.

For your second part are you concerned that if we don't specify that there was a helm chart release event then another repo will catch an incorrect event?

design/automate_release_design.md
sigalsax: Currently the action I envisioned does not create a new release in the SP rather open a PR in a completely separate repo. 
The one you are suggesting here handles a separate task in the release process (also important because that is still a manual) that could definitely be something to add in future developments section once I conquer the first obstacle. IMO your idea is much easier to implement then my original idea :)

For example, if a green bumped version bump:
1. grab content from changelog as the release description
2. Push the image and helm chart tarball
3. Create draft release. That way, there is still human involved to give the final OK

WDYT?

PR #257, opened 2020-11-10
Correctly set the tag in the project's version

### LINKED ISSUE #256: SP does not print <version>-dev on startup
On startup of the Secrets Provider deployment, we print the following:

`Secrets Provider for Kubernetes v... starting up`

During a customer case investigation while attempting to solve separate issue, it was found that we are printing <version>-dev which is no bueno. dev should only appear if the user manually builds the feature. See [here](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/Dockerfile#L23) how we do it for authn-client.

Currently
```
INFO: 2020/11/10 14:32:29 main.go:23: CSPFK008I CyberArk Secrets Provider for Kubernetes v1.1.0-dev starting up
```
Expected
```
INFO: 2020/11/10 14:32:29 main.go:23: CSPFK008I CyberArk Secrets Provider for Kubernetes v1.1.0 starting up
```

orenbm:
### What does this PR do?
When the secrets-provider starts it writes its version to the log.
Currently, it always writes
```
CSPFK008I CyberArk Secrets Provider for Kubernetes v1.1.0-dev starting up
```

Note that the `dev` is hard-coded.

After this commit, we correctly write the git tag of the version:
```
CSPFK008I CyberArk Secrets Provider for Kubernetes v1.1.0-f463bf8 starting up
```

Resolves #256

#### Change log
- [x] The CHANGELOG has been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   which adds some fixes around cert injection failure (see also changes in 
   [0.18.1](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/CHANGELOG.md#0181---2020-09-13)). [cyberark/secrets-provider-for-k8s#247](https://github.com/cyberark/secrets-provider-for-k8s/pull/247)
 
+### Fixed
+- The version that is printed at the product's startup now includes the git commit
+  hash instead of a hard-coded 'dev' string.
+  [cyberark/secrets-provider-for-k8s#256](https://github.com/cyberark/secrets-provider-for-k8s/issues/256)
+
 ## [1.1.0] - 2020-09-15
 ### Added
 - Helm chart to deploy the Secrets Provider Job using Helm ([cyberark/secrets-provider-for-k8s#165](https://github.com/cyberark/secrets-provider-for-k8s/pulls/165))
--- a/bin/build
+++ b/bin/build
@@ -27,6 +27,7 @@ if [ $DEBUG = true ]; then
   echo "Building secrets-provider-debug Docker image"
 
   docker build \
+    --build-arg TAG="debug" \
     --tag "secrets-provider-for-k8s-debug:latest" \
     --target "secrets-provider-debug" \
     .
@@ -34,6 +35,7 @@ else
   echo "Building secrets-provider-for-k8s:$FULL_VERSION_TAG Docker image"
 
   docker build \
+    --build-arg TAG=$(git_tag) \
     --tag "secrets-provider-for-k8s:dev" \
     --tag "secrets-provider-for-k8s:${FULL_VERSION_TAG}" \
     --tag "secrets-provider-for-k8s:latest" \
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   which adds some fixes around cert injection failure (see also changes in 
   [0.18.1](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/CHANGELOG.md#0181---2020-09-13)). [cyberark/secrets-provider-for-k8s#247](https://github.com/cyberark/secrets-provider-for-k8s/pull/247)
 
+### Fixed
+- The version that is printed at the product's startup now includes the git commit
```
sigalsax, line 21: do you still want the git commit's hash after the convo with @izgeri  in slack?

> we want the version for tags _just_ to be the tag

bin/build
```diff
@@ -27,13 +27,15 @@ if [ $DEBUG = true ]; then
...
   docker build \
+    --build-arg TAG="dev" \
     --tag "secrets-provider-for-k8s-debug:latest" \
     --target "secrets-provider-debug" \
     .
 else
   echo "Building secrets-provider-for-k8s:$FULL_VERSION_TAG Docker image"
 
   docker build \
+    --build-arg TAG=$(git_tag) \
```
sigalsax, line 38: if you want to follow Geri's note then just use short_version_tag found here https://github.com/cyberark/secrets-provider-for-k8s/blob/master/bin/build_utils#L11

CHANGELOG.md
orenbm: yes, but that effort will be bigger. i created an issue for that: https://github.com/cyberark/secrets-provider-for-k8s/issues/258

I think that we should first make this quick fix so that we don't have the hard-coded `dev` tag. the next fix will take more time and we can address that when we have the time.

bin/build
orenbm: this will output the following message:
```
CSPFK008I CyberArk Secrets Provider for Kubernetes v1.1.0-1.1.0 starting up
```

which is not what we want. what is this "Geri's note" that you're referring to?

sigalsax: See [here](https://conjurhq.slack.com/archives/C2V5T1NRZ/p1605023744116100?thread_ts=1605022517.115600&cid=C2V5T1NRZ)

orenbm: yes, i saw that and referred to it in the other comment. do you still think we should use `short_version_tag` here? i want to first remove the hard-coded `dev` tag and then properly fix it for static versions.

sigalsax: but wont we get `1.1.0-`? 
I actually think the commit hash would be nice so we know exactly which commit a customer is working with, not just the version which is much easier to troubleshoot. 
If you agree then tag should be `(git_tag)`

orenbm: closed as in merge the PR? because this PR adds the `git_tag`.

I think that we need the commit hash as long as it is not a released version. official versions should not include the commit hash but that will be taken care of in another issue.

sigalsax (approved):
IMPO LGTM!

PR #266, opened 2020-11-18
Skip build and test for markdown files

### LINKED ISSUE #264: Jenkins pipeline validation is running although only .md files were changed

abrahamko:
This will skip building the appliance and running all of the tests if the
branch only contains changes to markdown files.
It will still run the changelog parser so operations on markdown files still
happen, but blocks off the build and test portion of the build.

### What does this PR do?
Skip jenkins pipeline for PRs that include only `.md` files changes.
This allows easier working on docs in the repo, that does not have impact on code.

Similar changes in `conjur`: https://github.com/cyberark/conjur/pull/1939

Resolves #264 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,117 +22,126 @@ pipeline {
       }
     }
 
-    stage('Build client Docker image') {
-      steps {
-        sh './bin/build'
+    stage('Build and test Secrets Provider') {
+      when {
+        expression {
+         sh(returnStatus: true, script: 'git diff origin/master --name-only | grep -v "^.*\\.md$" > /dev/null') == 0
+        }
       }
-    }
-
-    stage('Scan Docker Image') {
-      parallel {
-        stage("Scan Docker Image for fixable issues") {
+      stages {
+        stage('Build client Docker image') {
           steps {
-            // Adding the false parameter to scanAndReport causes trivy to
-            // ignore vulnerabilities for which no fix is available. We'll
-            // only fail the build if we can actually fix the vulnerability
-            // right now.
-            scanAndReport('secrets-provider-for-k8s:latest', "HIGH", false)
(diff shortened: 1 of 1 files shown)


ismarc:
I'd hold off for a minute.  Hugh updated the regex to be more accurate for what we're l actually looking for, and with this approach there may be a way to extract pieces of it so it's clearer/easier to do in multiple repositories.

abrahamko:
@ismarc Could you please elaborate what we're waiting for?
Is Hugh work on adding an infra function to be consumed by all Jenkinsfiles?

ismarc:
Just for the PR on conjur to get merged (which it has).  I'd prefer to have the same regex used in each of the repos.  For the long-term, I think the expression itself can be lifted to the pipeline library, but the structure would have to stay the same.  Jonah is working on a restructure of the conjur Jenkinsfile that should confirm this as part of it, but it likely won't be a fast thing to get done.

Jenkinsfile
```diff
@@ -22,117 +22,126 @@ pipeline {
...
+        expression {
+         sh(returnStatus: true, script: 'git diff origin/master --name-only | grep -v "^.*\\.md$" > /dev/null') == 0
+        }
       }
-    }
-
-    stage('Scan Docker Image') {
-      parallel {
-        stage("Scan Docker Image for fixable issues") {
+      stages {
```
eladkug, line 31: why stages?

abrahamko: I did similar changes to the `conjur` PR: https://github.com/cyberark/conjur/pull/1939
In short, it needs a `stages` group to add the `when` on all steps need skipping.
These are all testing, excluding changelog step that still need running.

eladkug (approved):
LGTM

PR #268, opened 2020-11-24
Bump version to 1.1.1

### LINKED ISSUE #267: Release the secrets provider version that includes the authn-k8s client v0.19.0
To include v0.19.0 in the next OSS suite release, we'd like to bump the downstream project versions so that they include the same version of the authn-k8s client library.

AC:
- [ ] There is a new Secrets Provider release that includes the authn-k8s client version bump to v0.19.0

izgeri:
### What does this PR do?
Prepares for a 1.1.1 release.

Resolves #267 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/deploy/test/test_cases/TEST_ID_22_helm_rbac_defaults_taken_successfully.sh
+++ b/deploy/test/test_cases/TEST_ID_22_helm_rbac_defaults_taken_successfully.sh
@@ -26,4 +26,4 @@ $cli_with_timeout "get RoleBinding secrets-provider-role-binding"
 $cli_with_timeout "get ConfigMap cert-config-map"
 
 # Validate that the Secrets Provider took the default image configurations if not supplied and was deployed successfully
-$cli_with_timeout "describe job secrets-provider | grep 'cyberark/secrets-provider-for-k8s:1.1.0'" | awk '{print $2}' && $cli_with_timeout "get job secrets-provider -o jsonpath={.status.succeeded}"
+$cli_with_timeout "describe job secrets-provider | grep 'cyberark/secrets-provider-for-k8s:1.1.1'" | awk '{print $2}' && $cli_with_timeout "get job secrets-provider -o jsonpath={.status.succeeded}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,20 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [1.1.1] - 2020-11-24
 ### Added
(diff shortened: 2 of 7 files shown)


alexkalish (approved):

orenbm (approved):

orenbm:
thanks for getting this @izgeri !

izgeri:
@cyberark/palmtree to get this green, I had to remove a hardcoded version in one of your test scripts. so, two things:
- the approval wasn't dismissed when I pushed that additional change. do you want your repo config to be updated to dismiss stale reviews when new changes are pushed?
- can you please re-review given this additional change? does that look ok to make?

thanks!

deploy/test/test_cases/TEST_ID_22_helm_rbac_defaults_taken_successfully.sh
```diff
@@ -26,4 +26,4 @@ $cli_with_timeout "get RoleBinding secrets-provider-role-binding"
 $cli_with_timeout "get ConfigMap cert-config-map"
 
 # Validate that the Secrets Provider took the default image configurations if not supplied and was deployed successfully
-$cli_with_timeout "describe job secrets-provider | grep 'cyberark/secrets-provider-for-k8s:1.1.0'" | awk '{print $2}' && $cli_with_timeout "get job secrets-provider -o jsonpath={.status.succeeded}"
+$cli_with_timeout "describe job secrets-provider | grep 'cyberark/secrets-provider-for-k8s'" | awk '{print $2}' && $cli_with_timeout "get job secrets-provider -o jsonpath={.status.succeeded}"
```
sigalsax, line 29: @izgeri there needs to be a version here because we need to validate that the SP container took the default which is defined in the `values.yaml`. So in our case now, we want to verify that 1.1.1 is the version deployed and not 1.1.0. We planned on making a function to fetch the values.yaml but never got around to it. 

Anywho, we don't want to leave it without a version because if 1.1.0 by accident was taken then the tests would pass

sigalsax: if you don't have the bandwidth then I think it would be best to add back the 1.1.1 and open an issue to fetch the tag from value.yaml that way, the test still has meaning. WDYT?

sigalsax:
@izgeri 
>  do you want your repo config to be updated to dismiss stale reviews when new changes are pushed?

Is there a way to control if its code related changes or MD related. Nonetheless, for me, it isn't critical and most of the times it can be cumbersome to catch the person who did the review if you just changed a MD or added a comment in the code.

deploy/test/test_cases/TEST_ID_22_helm_rbac_defaults_taken_successfully.sh
izgeri: Filed #270 - I'll fix this now and also add a reference to this file in the CONTRIBUTING release notes now.

PR #269, opened 2020-11-30
Always run pipeline on tags and nightly tests

### LINKED ISSUE #264: Jenkins pipeline validation is running although only .md files were changed

abrahamko:
This PR merges the same changes introduced for `conjur` at these PRs:
https://github.com/cyberark/conjur/pull/1946
https://github.com/cyberark/conjur/pull/1951

Recently, we introduced a change to skip running the CI tests when the
only changed files were markdown files, so that design doc and other
non-code PRs wouldn't block on the CI.

However, this had the unintended side-effect of skipping our nightly
builds, because the criteria for "only markdown files":

git diff origin/master --name-only | grep -v "^.*\\.md$" > /dev/null'

will also _always_ be true when a nightly build is auto-triggered.

This commit ensures that the nightly builds and tag release builds will always be run.

Resolves #264

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,8 +24,34 @@ pipeline {
 
     stage('Build and test Secrets Provider') {
       when {
-        expression {
-         sh(returnStatus: true, script: 'git diff origin/master --name-only | grep -v "^.*\\.md$" > /dev/null') == 0
+        // Run tests only when EITHER of the following is true:
+        // 1. A non-markdown file has changed.
+        // 2. It's the nightly build.
+        // 3. It's triggered by conjur's build
+        // 4. It's a tag-triggered build.
+        anyOf {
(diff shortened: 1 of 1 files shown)


eladkug (commented):
What about trigger from conjur , it will work as well?

abrahamko:
> What about trigger from conjur , it will work as well?

I guess not, since someone needs to add it explicitly.
@hughsaunders  - do you have a suggestion how to handle this?

hughsaunders:
@abrahamko I added 'upstreambuild' to `getTrigger()` so you can use that to determine if the build is caused by an upstream. https://github.com/conjurinc/jenkins-pipeline-library/blob/master/vars/getTrigger.groovy

```
expression { getTrigger() == "upstreambuild" }
```

eladkug (approved):

PR #272, opened 2020-12-01
Always run pipeline on master branch

### LINKED ISSUE #264: Jenkins pipeline validation is running although only .md files were changed

abrahamko:
Recently, we introduced a change to skip running the CI tests when the
only changed files were markdown files, so that design doc and other
non-code PRs wouldn't block on the CI.

However, this had the unintended side-effect of skipping our builds
on the master branch as well.

This commit ensures that the master branch builds will always be run.

Resolves #264

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,9 +26,7 @@ pipeline {
       when {
         // Run tests only when EITHER of the following is true:
         // 1. A non-markdown file has changed.
-        // 2. It's the nightly build.
-        // 3. It's triggered by conjur's build
-        // 4. It's a tag-triggered build.
+        // 2. It's the master branch.
         anyOf {
           // Note: You cannot use "when"'s changeset condition here because it's
           // not powerful enough to express "_only_ md files have changed".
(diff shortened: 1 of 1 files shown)


eladkug (approved):
Looks much better. By that we will know that master will run everything all the time.
@hughsaunders WDYT?

Jenkinsfile
```diff
@@ -44,14 +42,8 @@ pipeline {
...
           }
 
-          // Always run the full pipeline on nightly builds
-          expression { params.NIGHTLY }
-
-          // Always run the full pipeline when triggered by conjur build
-          expression { getTrigger() == "upstreambuild" }
-
-          // Always run the full pipeline on tags of the form v*
-          tag "v*"
```
izgeri, line 54: we run tag pipelines now too - you may consider moving this project to tag-triggered builds. the benefit is that you run `bin/publish` on the condition `tag "v*"` instead of using a condition like this:
https://github.com/cyberark/secrets-provider-for-k8s/blob/5c2a51375eb6c1a39f91ee7169feaca7c9e4f30a/bin/publish#L22

the problem with your current publishing strategy is that if a new commit is published to master before the tag is pushed, your master build may not meet this condition and your tagged artifacts may not be published. using the tag-triggered pipelines gets around this risk, but it will mean you need to add back this condition to the Jenkinsfile.

PR #282, opened 2021-01-20
Do not set log level in main

orenbm:
We now use the authn-client logger and its log level
is set when we create the authnConfig. Thus, setting the
log level in the SP will cause the logger to include two
log messages of `CAKC052 Debug mode is enabled`.


abrahamko (approved):
Thanks for this PR.
It's my bad that I haven't cleaned it up after moving this logic to authn client.

PR #284, opened 2021-01-21
Add util to get pod name

orenbm:
### What does this PR do?
We get the pod name in a lot of our tests and each one does it differently. It is best to have one util that performs this
action and use it in out tests.

### DIFF
--- a/deploy/test/test_cases/TEST_ID_2_multiple_pods_changing_pwd_inbetween.sh
+++ b/deploy/test/test_cases/TEST_ID_2_multiple_pods_changing_pwd_inbetween.sh
@@ -9,9 +9,7 @@ wait_for_it 600 "$CONFIG_DIR/secrets-access-role-binding.sh.yml | $cli_without_t
 
 deploy_init_env
 
-$cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers | wc -l | tr -d ' ' | grep '^1$'"
-pod_name1=$($cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers" | awk '{print $1}')
-
+pod_name1="$(get_pod_name ${APP_NAMESPACE_NAME} 'test-env')"
 
 echo "Verify pod $pod_name1 has environment variable 'TEST_SECRET' with value 'supersecret'"
 verify_secret_value_in_pod $pod_name1 TEST_SECRET supersecret
@@ -35,8 +33,7 @@ elif [ $PLATFORM = "openshift" ]; then
     $cli_with_timeout "delete pod $pod_name1"
 fi
 
-$cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers | wc -l | tr -d ' ' | grep '^1$'"
-pod_name2=$($cli_with_timeout "get pods --namespace=$APP_NAMESPACE_NAME --selector app=test-env --no-headers" | awk '{print $1}')
+pod_name2="$(get_pod_name ${APP_NAMESPACE_NAME} 'test-env')"
 
 echo "Verify pod $pod_name2 has environment variable 'TEST_SECRET' with value 'secret2'"
 verify_secret_value_in_pod $pod_name2 TEST_SECRET secret2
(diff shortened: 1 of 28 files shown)


deploy/utils.sh
```diff
@@ -477,3 +478,35 @@ get_logs() {
...
+  local cli=${3-$cli_with_timeout}
+
+  pod_name=$(
+    $cli get pods \
+      --namespace=${namespace} \
+      --selector ${selector} \
+      -o jsonpath='{.items[].metadata.name}'
+  )
+
+  if [[ -z ${pod_name:-""} ]]; then
```
jonahx, line 505:
```sh
  if [[ -z $pod_name ]]; then
```
should work here since `pod_name` will always be defined (line 498).

Nicely formatted bash, btw!

deploy/utils.sh
```diff
@@ -477,3 +478,35 @@ get_logs() {
...
+    --selector app=test-env \
+    --no-headers
+}
+
+# Return the pod name of a given namespace and app name
+# For example: 'secret-provider-0'
+get_pod_name() {
+  local namespace=$1
+  local selector=$2
+  local cli=${3-$cli_with_timeout}
```
jonahx, line 496: Use:
```sh
  local cli=${3:-$cli_with_timeout}
```
What you have works but isn't documented in man bash.  `:-` is the documented syntax.

deploy/utils.sh
```diff
@@ -424,7 +425,7 @@ get_app_logs_container() {
     $cli_without_timeout get pods
 
     if [[ -z "$helm" ]]; then
-      pod_name=$($cli_without_timeout get pods --selector app=test-env --no-headers  | awk '{print $1}' )
+      pod_name="$(get_pod_name "${APP_NAMESPACE_NAME}" app=test-env $cli_without_timeout)"
```
jonahx, line 428: `$cli_without_timeout` should be double-quoted.

deploy/utils.sh
orenbm: this method was written by @izgeri :)

deploy/utils.sh
```diff
@@ -166,7 +172,7 @@ fetch_ssl_from_conjur() {
     export cert_location="/root/conjur-${CONJUR_ACCOUNT}.pem"
   fi
 
-  export conjur_pod_name=$($cli_with_timeout get pods --selector=$selector --namespace $CONJUR_NAMESPACE_NAME --no-headers | awk '{ print $1 }' | head -1)
+  export conjur_pod_name="$(get_pod_name "${CONJUR_NAMESPACE_NAME}" $selector)"
```
jonahx, line 175: `$selector` should be double quoted.  Also for consistency, `"$CONJUR_NAMESPACE_NAME"` is fine here -- no need for the braces.

deploy/utils.sh
orenbm: oh right, missed that. I usually remember this by knowing that the syntax is a smily face when the default value is empty (`:-}`)

deploy/utils.sh
```diff
@@ -13,22 +16,25 @@ wait_for_it() {
...
   local spacer=5
   shift
-  if ! [ $timeout = '-1' ]; then
+  if ! [ "${timeout}" = '-1' ]; then
     local times_to_run=$((timeout / spacer))
 
-    #echo "Waiting for '$@' up to $timeout s"
     for i in $(seq $times_to_run); do
-      eval $*  && return 0
+      if cmd_output=$(eval "$@") ;
```
jonahx, line 23: Avoid `eval`.  Instead, you can de-structure this function's arguments into `cmd` and `args`, and then do:
```sh
      if cmd_output=$("$cmd" "${args[@]}") ; then
```

orenbm: it's not that simple. please look for usages of `wait_for_it`.

deploy/test/test_cases/TEST_ID_2_multiple_pods_changing_pwd_inbetween.sh
```diff
@@ -55,9 +52,10 @@ echo "Waiting for 3 running pod test-env"
 $cli_with_timeout "get pods | grep test-env | grep Running | wc -l | tr -d ' ' | grep '^3$'"
 
 echo "Iterate over new pods and verify their secret was updated"
-pod_names=$(cli_get_pods_test_env | awk '{print $1}' | grep -v $pod_name2)
+pods_info=$(get_pods_info "$APP_NAMESPACE_NAME" test-env)
```
abrahamko: why the `pods_info` necessary?

deploy/utils.sh
```diff
@@ -479,3 +475,43 @@ get_logs() {
...
+  pod_name=$(
+    $cli_with_timeout get pods \
+      --namespace=${namespace} \
+      --selector app=${app_name} \
+      -o jsonpath='{.items[].metadata.name}'
+  )
+
+  if [[ -z ${pod_name:-""} ]]; then
+    echo "Unable to find ${app_name} pod in namespace ${namespace} - aborting."
+    eval "${cli_with_timeout}" describe pods --namespace="${namespace}"
```
abrahamko: I think there's no need to use `eval` here.
call `$cli_with_tiomeout`, like above.

deploy/test/test_cases/TEST_ID_2_multiple_pods_changing_pwd_inbetween.sh
orenbm: this is an old code

deploy/utils.sh
jonahx: Fair enough.  Should probably be changed everywhere forcing code to be wrapped in functions, but I won't ask you to do that :)

jonahx (approved):

PR #291, opened 2021-01-28
Fix insertion of helm namespace parameter

### LINKED ISSUE #290: Helm namespace is ignored
## Summary
The secrets provider’s helm chart uses `{{ .Release.namespace }}`  and it won’t work! It’s supposed to be 
`{{ .Release.Namespace}}`
## Steps to Reproduce
`helm template secrets-provider  -f=values.yaml  --namespace=main | grep namespace`

## Expected Results
I would like to set my own namespace for the secrets provider chart

## Actual Results (including error logs, if applicable)
namepace was null
## Reproducible
   * [x] Always 
   * [ ] Sometimes
   * [ ] Non-Reproducible
   
## Version/Tag number
Always
Can you describe the environment in which this product is running? Is it running on a VM / in a container / in a cloud? 
EKS
## Additional Information
https://stackoverflow.com/questions/51760772/why-is-release-namespace-rendered-empty

liavyona:
### What does this PR do?
Fix the namespace parameter

Resolves #290 
#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Support for OpenShift 4.5.
   [cyberark/secrets-provider-for-k8s#265](https://github.com/cyberark/secrets-provider-for-k8s/issues/265)
 
+### Fixed
+- The Secrets Provider helm templates are updated to correctly refer to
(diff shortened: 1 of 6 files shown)


izgeri:
@liavyona this should have a changelog entry, right? if you hit this issue, then others using the helm chart may also have hit it and it would be good to note in the changelog that it is "Fixed" in this next release

izgeri (approved):

PR #294, opened 2021-02-01
Use built-in 'wait' command to wait for job completion

orenbm:
We have some instability with the way that we wait for the job to be ready. Hopefully using the built-in command for this in kubernetes will help us.


eladkug (approved):
Looks like a great solution

PR #300, opened 2021-02-10
Updated project to use new go client feature

### LINKED ISSUE #288: Bump Go client version
## Is your feature request related to a problem? Please describe.

The Go client was updated to retrieve batch secrets using the `Accept: base64` header in order to avoid encoding errors in Conjur. This secrets provider should be updated to use the new version of the client which includes this update.

## Additional context

[Here](https://github.com/cyberark/conjur/issues/1962) is the original issue referencing the use of secrets provider for k8s.

telday:
### What does this PR do?
Secrets provider now uses the `RetrieveBatchSecretsSafe` go client method which ensures that binary secrets are properly handled by Conjur.

Resolves #288 

#### Change log

#### Test coverage

#### Documentation


izgeri:
@telday if there is a way to add a binary secret to one of the test flows to ensure that is tested here too, that would be great

deploy/test/test_cases/TEST_ID_18_helm_multiple_provider_multiple_secrets.sh
```diff
@@ -4,9 +4,10 @@ set -euxo pipefail
 # This test verifies that two Secrets Provider Jobs deploy successfully in the same namespace
 setup_helm_environment
 
+secret_value=$(openssl rand 10)
```
orenbm, line 7: i am guessing this is a binary secret? can you add a comment if this is correct?
also - if the SP succeeds to provide a binary secret, is it granted that it can a plaintext one?

Dockerfile.test
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12.0-alpine3.9
+FROM golang:1.13-alpine
```
izgeri, line 1: I think we'll want this at 1.15 - but I've just pinged @BradleyBoutcher, and he may fix it elsewhere

izgeri (commented):
Noting that this will need a CHANGELOG entry and a NOTICES update

PR #301, opened 2021-02-15
Login to the registry url which we pull images from

orenbm:
We had some issues where we sometimes failed to pull images. This may
be caused by the fact that we log into one registry and pull the SP
image from another.

In OCP 4.5 we changed the registry that we pull images from to our internal registry. We now also login to that registry.

### DIFF
--- a/deploy/platform_login.sh
+++ b/deploy/platform_login.sh
@@ -12,7 +12,7 @@ main() {
       $GCLOUD_CLUSTER_NAME \
       --zone $GCLOUD_ZONE \
       --project $GCLOUD_PROJECT_NAME
-    docker login $DOCKER_REGISTRY_PATH \
+    docker login $PULL_DOCKER_REGISTRY_URL \
       -u oauth2accesstoken \
       -p "$(gcloud auth print-access-token)"
   elif [[ "$PLATFORM" = "openshift" ]]; then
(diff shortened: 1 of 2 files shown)


deploy/platform_login.sh
```diff
@@ -12,7 +12,7 @@ main() {
       $GCLOUD_CLUSTER_NAME \
       --zone $GCLOUD_ZONE \
       --project $GCLOUD_PROJECT_NAME
-    docker login $DOCKER_REGISTRY_PATH \
+    docker login $PULL_DOCKER_REGISTRY_URL \
```
izgeri, line 15: I don't think this is correct, actually; when you login here, it is the registry you will use when you `docker push` here: https://github.com/cyberark/secrets-provider-for-k8s/blob/74d4eaa8bf4088d6735a08dedbfda6f9d30910ab/deploy/run_with_summon.sh#L48

you want your apps to `docker pull` from the internal registry path (`PULL_DOCKER_REGISTRY_URL`), but that's unrelated to your local client login. since you are running `docker push` from a remote machine (ie not from the K8s cluster), you want to `docker login` to the external facing Docker registry URL.

This logic ^ appears to be consistent with how it's implemented in k8s-conjur-demo at current.

Can @cyberark/conjur-infra-team weigh in on the difference between `ci/openshift/current/registry-url` and `ci/openshift/current/internal-registry-url` for OCP 4.5+?

orenbm: and why we need to push into one and pull from another

orenbm:
Closing as this approach is not correct. we need to login to he external registry.

PR #309, opened 2021-03-09
Add support for newer openshift labels in test infra

### LINKED ISSUE #302: Pipelines run on OCP 4.6
## Description
As an end user who runs the Secrets Provider in OpenShift,
I want the Secrets Provider to run on OCP 4.6
So that I can use the entire DAP product on my 4.6 cluster

## Details
To implement this, we want to add a pipeline flow that runs the `next` cluster on a pipeline trigger.
- [x] Update deploy/summon/secrets.yml to reference `oldest`, `current`, and `next` with no hardcoded versions (version can be taken from the conjurops var `ACCOUNT/openshift/CLUSTER/version`
- [x] Update existing flows to remove hardcoded OCP versions and instead reference `oldest` and `current` / use the OCP version value from ConjurOps (stored in the `OPENSHIFT_VERSION` env var). Note that Secrets Provider test flows should NOT run against `oldest` in this update - the `oldest` flow (OCP 4.3) is currently not supported.
- [x] Add a flow to the Jenkinsfile for `next`:
       - Add a boolean parameter for whether to run the `next` pipeline (this should only be run when testing newer OCP versions):
         ```
          parameters {
            booleanParam(name: 'TEST_OCP_NEXT', description: 'Whether or not to run the pipeline against the next OCP version', defaultValue: false)
          }
          ```
        - Add a stage for testing against `next` that is gated on `when { expression { params.TEST_OCP_NEXT } }`

Once these changes are in place, we can test flows against the `next` cluster by:
- [x] Run the pipeline against the `next` cluster by setting the `TEST_OCP_NEXT` value to `true` in the Jenkins UI

BradleyBoutcher:
### What does this PR do?
We now support running tests against the 'next', 'current', and 'oldest' versions of Openshift, as defined in Conjurops.

Currently, only 'current' and 'next' are supported

Resolves #302 

#### Change log
- [X] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage

#### Documentation
- [X] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,12 @@ pipeline {
     timeout(time: 3, unit: 'HOURS')
   }
 
+  parameters {
+    booleanParam(name: 'TEST_OCP_NEXT', defaultValue: false, description: 'Run DAP tests against our running "next version" of Openshift')
+
+    booleanParam(name: 'TEST_OCP_OLDEST', defaultValue: false, description: 'Run DAP tests against our running "oldest version" of Openshift')
+  }
+
   stages {
     stage('Validate') {
       parallel {
@@ -105,15 +111,25 @@ pipeline {
           steps {
             script {
               def tasks = [:]
-                tasks["Kubernetes GKE, DAP"] = {
-                  sh "./bin/start --docker --dap --gke"
-                }
-                tasks["Openshift v3.11, DAP"] = {
+              tasks["Kubernetes GKE, DAP"] = {
+                sh "./bin/start --docker --dap --gke"
+              }
+              tasks["Openshift v3.11, DAP"] = {
                   sh "./bin/start --docker --dap --oc311"
+              }
+              if ( params.TEST_OCP_OLDEST ) {
+                tasks["Openshift (Oldest), DAP"] = {
+                  sh "./bin/start --docker --dap --oldest"
                 }
-                tasks["Openshift v4.5, DAP"] = {
-                  sh "./bin/start --docker --dap --oc45"
+              }
+              tasks["Openshift (Current), DAP"] = {
+                sh "./bin/start --docker --dap --current"
+              }
+              if ( params.TEST_OCP_NEXT ) {
+                tasks["Openshift (Next), DAP"] = {
+                  sh "./bin/start --docker --dap --next"
                 }
+              }
               parallel tasks
             }
           }
--- a/bin/start
+++ b/bin/start
@@ -20,9 +20,13 @@ Usage: ./bin/start [options]
 
     --gke         Run the integration tests on GKE. By default the tests are run on GKE.
 
-    --oc311       Run the integration tests on Openshift v3.11. By default the tests are run on GKE.
+    --oc311 Run the integration tests on Openshift v3.11. By default the tests are run on GKE.
 
-    --oc45       Run the integration tests on Openshift v4.5. By default the tests are run on GKE.
+    --oldest      Run the integration tests on the oldest running Openshift platform. By default the tests are run on GKE.
+
+    --current     Run the integration tests on the currently supported Openshift platform. By default the tests are run on GKE.
+
+    --next        Run the integration tests on the latest Openshift platform. By default the tests are run on GKE.
 
     -h, --help    Shows this help message.
 EOF
@@ -30,7 +34,7 @@ EOF
 }
 
 runScriptWithSummon() {
-    summon --environment $SUMMON_ENV -f ./summon/secrets.yml $1
+    summon --provider summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
 }
 
 RUN_IN_DOCKER=false
@@ -49,7 +53,9 @@ while true ; do
     --dap ) CONJUR_DEPLOYMENT=dap ; shift ;;
     --gke ) SUMMON_ENV=gke ; shift ;;
     --oc311 ) SUMMON_ENV=oc311 ; shift ;;
-    --oc45 ) SUMMON_ENV=oc45 ; shift ;;
+    --oldest ) SUMMON_ENV=oldest ; shift ;;
+    --current ) SUMMON_ENV=current ; shift ;;
+    --next ) SUMMON_ENV=next ; shift ;;
     -h | --help ) print_help ; shift ;;
      * ) if [ -z "$1" ]; then break; else echo "$1 is not a valid option"; exit 1; fi;;
   esac
--- a/deploy/run_with_summon.sh
+++ b/deploy/run_with_summon.sh
@@ -20,6 +20,10 @@ finish() {
 }
 trap finish EXIT
 
+# Will print platform regardless of GKE or Openshift
+# Will only print version if Openshift
+announce "Running tests on: ${PLATFORM} ${OPENSHIFT_VERSION}"
+
 if [ "${DEV}" = "false" ]; then
   ./platform_login.sh
 fi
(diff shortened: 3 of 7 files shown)


Jenkinsfile
```diff
@@ -89,31 +93,33 @@ pipeline {
...
-                tasks["Openshift v3.11, DAP"] = {
-                  sh "./bin/start --docker --dap --oc311"
-                }
-                tasks["Openshift v4.5, DAP"] = {
-                  sh "./bin/start --docker --dap --oc45"
+              tasks["Kubernetes GKE, DAP"] = {
+                sh "./bin/start --docker --dap --gke"
+              }
+              tasks["Openshift (Current), DAP"] = {
+                sh "./bin/start --docker --dap --current"
```
orenbm, line 116: thanks @BradleyBoutcher for getting this!
Can we verify that the OCP version is logged in a clear place so we know which versions are tested in each Jenkins run?

orenbm (dismissed):
Thanks @BradleyBoutcher !

Just blocking this PR from being merged just so we don't merge the OSS tests removal by accident :)

Jenkinsfile
```diff
@@ -105,15 +109,17 @@ pipeline {
           steps {
             script {
               def tasks = [:]
-                tasks["Kubernetes GKE, DAP"] = {
-                  sh "./bin/start --docker --dap --gke"
-                }
-                tasks["Openshift v3.11, DAP"] = {
-                  sh "./bin/start --docker --dap --oc311"
```
izgeri, line 112: we will need to keep the 3.11 tests in here for now

Jenkinsfile
```diff
@@ -105,15 +109,17 @@ pipeline {
...
-                }
-                tasks["Openshift v4.5, DAP"] = {
-                  sh "./bin/start --docker --dap --oc45"
+              tasks["Kubernetes GKE, DAP"] = {
+                sh "./bin/start --docker --dap --gke"
+              }
+              tasks["Openshift (Current), DAP"] = {
```
izgeri, line 115: maybe we should add a test flow against `oldest` but comment it out for now? oldest is currently 4.3, and SP is not certified for 4.3

bin/start
```diff
@@ -20,17 +20,19 @@ Usage: ./bin/start [options]
 
     --gke         Run the integration tests on GKE. By default the tests are run on GKE.
 
-    --oc311       Run the integration tests on Openshift v3.11. By default the tests are run on GKE.
+    --oldest     Run the integration tests on the oldest running Openshift platform. By default the tests are run on GKE.
```
diverdane, line 23: Silly nit: The spacing before the word `Run` is off by 1 for `--oldest` and `--next` options.
That's all I've got, everything else looks good to me (beyond comments already made).

bin/start
```diff
@@ -20,17 +20,21 @@ Usage: ./bin/start [options]
...
+    --next        Run the integration tests on the latest Openshift platform. By default the tests are run on GKE.
 
     -h, --help    Shows this help message.
 EOF
   exit
 }
 
 runScriptWithSummon() {
-    summon --environment $SUMMON_ENV -f ./summon/secrets.yml $1
+    summon -p summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
```
izgeri, line 37: nit, but if we add this here let's use the long form of the flag `--provider`

deploy/run_with_summon.sh
```diff
@@ -20,6 +20,10 @@ finish() {
 }
 trap finish EXIT
 
+# Will print platform regardless of GKE or Openshift
+# Will only print version if Openshift
+echo "Running tests on: ${PLATFORM} ${OPENSHIFT_VERSION}"
```
izgeri, line 25: can we use `announce` instead of echo? it will def get lost in the logs with just echo.

deploy/summon/secrets.yml
```diff
@@ -41,17 +41,45 @@ oc311:
...
-oc45:
-  OPENSHIFT_VERSION: '4.5'
+oldest:
+  OPENSHIFT_VERSION: !var ci/openshift/oldest/version
+  OPENSHIFT_URL: !var ci/openshift/oldest/hostname
+  OPENSHIFT_USERNAME: !var ci/openshift/oldest/username
+  OPENSHIFT_PASSWORD: !var ci/openshift/oldest/password
+
+  PLATFORM: openshift
+  TEST_PLATFORM: !var ci/openshift/oldest/version
```
izgeri, line 51: it looks like kubernetes-conjur-deploy expects this to be `openshift*` (though I'm not sure why it's working now)
https://github.com/cyberark/kubernetes-conjur-deploy/blob/fbe14df8a20e451ffdebca0d4b73c85df52d2bcc/test.sh#L43-L51

you can probably just make them all the literal strings `openshift` instead
```suggestion
  TEST_PLATFORM: openshift
```

deploy/summon/secrets.yml
```diff
@@ -41,17 +41,45 @@ oc311:
...
+  OPENSHIFT_VERSION: !var ci/openshift/current/version
   OPENSHIFT_URL: !var ci/openshift/current/api-url
   OPENSHIFT_USERNAME: !var ci/openshift/current/username
   OPENSHIFT_PASSWORD: !var ci/openshift/current/password
   OSHIFT_CLUSTER_ADMIN_USERNAME: !var ci/openshift/current/username
   OSHIFT_CONJUR_ADMIN_USERNAME: !var ci/openshift/current/username
 
   PLATFORM: openshift
-  TEST_PLATFORM: openshift45
+  TEST_PLATFORM: !var ci/openshift/current/version
```
izgeri, line 66:
```suggestion
  TEST_PLATFORM: openshift
```

deploy/summon/secrets.yml
```diff
@@ -41,17 +41,45 @@ oc311:
...
+next:
+  OPENSHIFT_VERSION: !var ci/openshift/next/version
+  OPENSHIFT_URL: !var ci/openshift/next/api-url
+  OPENSHIFT_USERNAME: !var ci/openshift/next/username
+  OPENSHIFT_PASSWORD: !var ci/openshift/next/password
+  OSHIFT_CLUSTER_ADMIN_USERNAME: !var ci/openshift/next/username
+  OSHIFT_CONJUR_ADMIN_USERNAME: !var ci/openshift/next/username
+
+  PLATFORM: openshift
+  TEST_PLATFORM: !var ci/openshift/next/version
```
izgeri, line 81:
```suggestion
  TEST_PLATFORM: openshift
```

izgeri (changes requested):
A few minor comments. Also, we should be adding a CHANGELOG entry about validing the secrets provider on OpenShift 4.6.

deploy/summon/secrets.yml
izgeri: btw, this seems silly and redundant now. makes me think we should just change the logic in k8s-deploy to use `PLATFORM`, remove `TEST_PLATFORM` from that project, and then remove it from here.

izgeri (commented):
LGTM, but if we can verify that `next` is passing we should update the [supported platforms](https://github.com/cyberark/secrets-provider-for-k8s/#supported-platforms) in the README with this PR to include 4.6

izgeri (approved):
LGTM pending green build, including with `next` pipeline

izgeri:
It's all green, including on OCP 4.6! See [Jenkins](https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--secrets-provider-for-k8s/detail/add-openshift-envs/27/pipeline/) for confirmation. Thanks @BradleyBoutcher !!

PR #312, opened 2021-03-16
Updating the README.md with the new name for DAP.

### LINKED ISSUE #311: Need clarification using secrets-provider-for-k8s
Hi,

Ques1: In Readme mentioned that for Openshift (dap only - Is it enterprise one?) is it supported only for DAP and we have a conjur oss setup shall we use it?
Ques2:
 while using conjur secret provider in Openshift. In conjur-map spec in secrets.yaml we don't want mention the values for uname and password with base64, instead we're specifying the the secret path for conjur server to retrieve the secret value.
So, when init container for secret-provider-k8s  running it will fetch the secret values from conjur-server-oss and update to k8s secret. 
In this case this secret-provider-k8s container get the original value and update on the keys on k8s secret with base64 encoded format?
If its updated in original k8s secret with those values again in k8s secret it will store the values in ETCD as plaintext right ? 
then there is no use of k8s-secret-povider with conjur for security purpose right?
Please clarify the above.

rpothier:
### What does this PR do?
References to DAP in README.md are replaced with Conjur Enterprise.
The is also a comment that it was formerly know as DAP

Resolves #311 

#### Change log

#### Test coverage

#### Documentation
- [X] This PR does not require updating any documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,15 +14,17 @@
 
 # CyberArk Secrets Provider for Kubernetes
 
-The CyberArk Secrets Provider for Kubernetes enables DAP to retrieve secrets stored and managed in the CyberArk Vault. The
- secrets can be consumed by your Kubernetes or Openshift application containers. To retrieve the secrets from Conjur or DAP, 
- the CyberArk Secrets Provider for Kubernetes runs as an init container or application container and fetches the 
- secrets that the pods require.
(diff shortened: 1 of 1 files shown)


izgeri (approved):
LGTMM

izgeri:
@rpothier it looks like you need a rebase - if you rebase, lmk and I will reapprove

PR #320, opened 2021-05-24
ConjurOps V5 Compatibility

hughsaunders:
CONJUR_MAJOR_VERSION was set to v4 for compatibility with v4
conjurops. Now that Jenkins agents are transitioning to conjurops v5,
This variable is no longer required.

Related: conjurinc/ops#782

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -34,7 +34,9 @@ EOF
 }
 
 runScriptWithSummon() {
-    summon --provider summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
+    # CONJUR_ACCOUNT is set in bootstrap.env to a test value, which isn't valid for
+    # conjurops.
+    CONJUR_ACCOUNT="conjur" summon --provider summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
 }
 
 RUN_IN_DOCKER=false
@@ -72,7 +74,7 @@ export DEV
 export SUMMON_ENV
 
 # summon environment variable
-export CONJUR_MAJOR_VERSION=4
+export CONJUR_MAJOR_VERSION=5
 
 source bootstrap.env
 


bin/start
```diff
@@ -72,7 +74,7 @@ export DEV
 export SUMMON_ENV
 
 # summon environment variable
-export CONJUR_MAJOR_VERSION=4
+export CONJUR_MAJOR_VERSION=5
```
sigalsax, line 77: We don't use this so I think this can be removed

bin/start
```diff
@@ -34,7 +34,9 @@ EOF
 }
 
 runScriptWithSummon() {
-    summon --provider summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
+    # CONJUR_ACCOUNT is set in bootstrap.env to a test value, which isn't valid for
```
sigalsax, line 37: so should we remove it from the boostrap.env file?

bin/start
```diff
@@ -34,7 +34,9 @@ EOF
...
+    # conjurops.
+    CONJUR_ACCOUNT="conjur" summon --provider summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
```
sigalsax, line 39: should these be two lines?
```
CONJUR_ACCOUNT="conjur"
summon --provider summon-conjur --environment $SUMMON_ENV -f ./summon/secrets.yml $1
```
?
Otherwise how does that summon line get executed if it is saved in a var?

hughsaunders: Bash allows setting an environment variable for just one command by prepending it to the line:
```
11:15 $ date
Tue 25 May 2021 11:15:24 BST
11:15 $ TZ="America/New_York" date
Tue 25 May 2021 06:15:27 EDT
```

bin/start
hughsaunders: Looks like it might be needed for connecting to a temporary test conjur instance?

bin/start
sigalsax: so it sets the command and runs it?

sigalsax (approved):
LGTM!

bin/start
hughsaunders: Summon will use it, and currently its useful for it to be explicitly set to 5.

bin/start
hughsaunders: Discussed in slack

PR #321, opened 2021-05-28
Adds solution design spec for M1 Push to File

### LINKED ISSUE #316: There is a solution design for adding the ability to push secret values to a file in a shared volume
The initial design will not include:
- Test plan

diverdane:
### What does this PR do?
Adds a solution design specification for M1 Push to File for Secrets Provider.

Resolves #316

#### Change log

#### Test coverage

#### Documentation


design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+<td><p><a href="https://cyberark.aha.io/epics/SCR-E-76">link</a> (private)</p>
+<p><em>Note: This design document covers work for “Milestone 1: Push to File” as defined in this Aha Card.</em></p></td>
+</tr>
+<tr class="odd">
+<td>Feature Doc</td>
+<td><a href="https://cyberark365.sharepoint.com/:w:/r/sites/Conjur/_layouts/15/Doc.aspx?sourcedoc=%7BB782E509-693F-4086-85A6-5D477A0F4ABD%7D&file=Feature%20Doc%20-%20Kubernetes%20Developer%20Experience%20M1.docx&action=default&mobileredirect=true&cid=048c1606-6533-443c-ba09-f91590c68095">link</a> (private)</td>
+</tr>
+<tr class="even">
+<td>Sample Manifests and Policies From Feature Spec</td>
+<td><a href="https://gist.github.com/alexkalish/8a514defa9e741800c095dded9837582">link</a> (private)</td>
```
izgeri, line 48: I think this is not private - anyone can access with the link

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+### User Experience
+
+#### Annotations for Secrets Provider Container/Application Configuration
+
+The Secrets Provider container/application configuration can be set using
+either Pod annotations or by using Pod environment variable settings
+([legacy configuration](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/11.2/en/Content/Integrations/Kubernetes_deployApplicationsConjur-k8s-Secrets.htm#Step4AddtheCyberArkSecretsProviderforKubernetesimagetotheapplicationdeploymentmanifest)).
+If both a Pod Annotation and a Pod environment variable is configured for
+any particular Secrets Provider setting, then the Pod Annotation configuration
+takes precedence.
```
izgeri, line 116: I don't think we're planning on adding an env-var mode for "file"-based flows, so this would only be supported for the "k8s"-based flows

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+
+_**QUESTION: Should we deprecate the current configuration method that uses
+   environment variable settings?**_
```
izgeri, line 119: @alexkalish this one is for you

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+The following annotations are supported for Secrets Provider configuration.
+Please refer to the
+[Secrets Provider documentation](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/11.2/en/Content/Integrations/Kubernetes_deployApplicationsConjur-k8s-Secrets.htm#Step4AddtheCyberArkSecretsProviderforKubernetesimagetotheapplicationdeploymentmanifest)
+for a description of each setting:
+
+| Annotation | Equivalent<br /> Environment Variable | Description, Notes |
+|--------------------------------------|---------------------|----------------------------------|
+| conjur.org/<br />authn-identity      | CONJUR_AUTHN_LOGIN  | Required value. Example: `host/conjur/authn-k8s/cluster/apps/inventory-api` |
+| conjur.org/<br />container-mode      | CONTAINER_MODE      | Allowed values: <ul><li>`init`</li><li>`application`</li><li>`sidecar`</li></ul>Defaults to `sidecar` |
+| conjur.org/<br />secrets-destination | SECRETS_DESTINATION | Required value. Allowed values: <ul><li>`file`</li><li>`k8s_secret`</li></ul> |
```
izgeri, line 135: ~maybe highlight that "file" is to be added here? at first, this chart read like a list of _current_ supported config options (ie it doesn't include the new file-based annotations), so I was surprised to see this option here~

~alternatively, we could add all of the new options and just highlight the ones that are new / to be added in this effort~

editing to just say instead: maybe we could highlight the new options in these tables? e.g. "New configuration settings or options are highlighted in **bold**" and then make "file" and `conjur.org/conjur-configmap` bold

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+| conjur.org/<br />k8s-secrets         | K8S_SECRETS         | This list is ignored when `conjur.org/secrets-destination` Annotation is set to `file` |
+| conjur.org/<br />retry-count-limit   | RETRY_COUNT_LIMIT   | Defaults to 5
+| conjur.org/<br />retry-interval-sec  | RETRY_INTERVAL_SEC  | Defaults to 1 (sec)              |
+| conjur.org/<br />debug-logging       | DEBUG               | Defaults to `false`              |
+| conjur.org/<br />conjur-configmap    | CONJUR_AUTHN_URL<br />CONJUR_URL<br />CONJUR_ACCOUNT<br />CONJUR_SSL_CERTIFICATE | ConfigMap containing Conjur connection information. |
```
izgeri, line 140: will we have a default value for the configmap name? I think that'd be nice to have - then people could skip it if they're using the same name.

also, do we assume the configmap is in the same namespace as the pod? can we be clear about that here?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+| conjur.org/conjur-secrets.{secret-group}       | List of secrets to be retrieved from Conjur. Each entry can be either:<ul><li>A Conjur variable path</li><li> A key/value pairs of the form `<alias>:<Conjur variable path>` where the `alias` represents the name of the secret to be written to the secrets file |
+| conjur.org/conjur-secrets-path.{secret-group}  | Prefix to use for all Conjur variable paths for a secrets group. Defaults to root policy path. |
+| conjur.org/secret-file-path.{secret-group}     | File path (including file name) for secrets file to be written. Default is `{secret-group}.{secret-file-type}`, e.g. `database.yaml`. |
+| conjur.org/secret-file-type.{secret-group}     | Allowed values:<ul><li>yaml (default)</li><li>json</li><li>dotenv</li><li>bash</li></ul>This setting is ignored when `conjur.org/secret-file-template.{secret-group}` is configured. |
+| conjur.org/secret-file-perms.{secret-group}    | File permissions for secrets file to be written, in octal format. Defaults to `660` |
+| conjur.org/secret-file-template.{secret-group} | Customized secret file template using the Golang templating language. |
```
izgeri, line 166:
```suggestion
| conjur.org/secret-file-template.{secret-group} | Custom template for secret file contents in [Golang text template](https://golang.org/pkg/text/template/) format. |
```

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+      - prod/backend/port 
+      - prod/backend/password
+      - prod/backend/username
+```
+
+For `conjur-secrets` entries without aliases, the last word in the path is
+used as the application secret name. So for the example above, if
+`conjur.org/secret-file-type.database` is set to `yaml`, then the resulting
+secrets file that is written to the shared secrets volume might look similar
+to the following:
```
izgeri, line 186:
```suggestion
no other annotations are set for the `database` secret group, then Secrets
Provider will write to a file `database.yaml` in the shared volume `/path/to/volume`
with contents that look like the following:
```

btw, what is the default mount path that the SP container will assume? I presume that we should have a default (it would correspond to default `secrets.mountPath` value in the helm chart section you have for SP volumes) and the `database.yaml` file would be written there (to the default place), unless the `conjur.org/conjur-secrets-path.` annotation is set

also, database.yaml or database.yml?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+Annotation that defines a list of Conjur variable paths, including some that
+use an alias, for a hypothetical secrets group named `cache`:
+
+```
+    conjur.org/conjur-secrets.cache: |
+      - url
+      - admin-password: password
+      - admin-username: username
+
+    conjur.org/conjur-secrets-path.cache: dev/memcached/
```
izgeri, line 207: does it matter if it's `/dev/memcached` vs `dev/memcached`?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+- dev/memcached/url
+- dev/memcached/password
+- dev/memcached/username
+
+and the names of the secrets that are written to the secrets file are:
+
+- url
+- admin-password
+- admin-username
+
```
izgeri, line 222: we could add here:

---

so that the `cache.yaml` file that Secrets Provider writes includes:
```
url: https://database.example.com
admin-password: my-secret-p@$$w0rd
admin-username: admin
```

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+###### Example `conjur.org/secret-file-template.{secret-group}`
+
+The following is an example of a custom secrets file template that uses the
+Golang templating language. Variables can be referenced via their name
```
izgeri, line 226:
```suggestion
Golang `text/template` syntax. Variables can be referenced via their name
```

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+      {
+        "cache": {
+          "url": {{ .url }},
+          "password": {{ .admin-password }},
+          "username": {{ .admin-username }},
+          "port": 123456
+        }
+      }
+```
+
```
izgeri, line 241: maybe include the sample output file here?

also, if no output file name is given this will write to a file named `cache.yaml`. I'd recommend we _require_ output file name if custom templates are used. what do you think?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+-   Platforms:
+
+    -   Kubernetes (this will be either Kubernetes-in-Docker, or GKE).
+    -   OpenShift 4.6
+
+#### Out of Scope
+
+- For this release, configuration will only be read upon startup.
+- For this release, annotations are required to have a group.
```
izgeri, line 339: Maybe be clear that you are only talking about the "push to file" annotations, and even link to the "Annotations for Push-to-File Secrets Mappings" section?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+      `command` entrypoint includes sourcing of a bash-formatted secrets file.
+
+### Project Scope and Limitations
+
+The initial implementation and testing will be limited to:
+
+-   Authentication containers to be tested:
+
+    -   Secrets Provider init container
+    -   Secrets Provider standalone Pod
```
izgeri, line 329: It might be worth digging into this a bit more. Unless we plan to fully deprecate env var config for "k8s" mode (an increase in scope which I'd not recommend now), we won't change existing tests for the init container / standalone pod modes. Rather, we will add tests for testing each mode when configured with annotations, and we will add init container tests with "push to file" mode.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+of the number of secrets groups and the number of secrets per secrets group.
+Details are TBD, but a representative configuration might look something like
+this:
+
+- 5 secrets groups
+- 5 secrets / secrets group
+- Mix of secret value lengths (10 to 20 characters, plus some base64-encoded
+  TLS certificates)
+
+## Logs
```
izgeri, line 535: Can we reference the "Error Handling / Recovery / Supportability tests" for the error modes, or should we combine these two tables?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1163 @@
...
+  `conjur.org` but an unexpected annotation type will be logged at `Info` level.
+- *Note 3:* Care must be taken to prevent secrets file content from leaking.
+
+## Documentation
+
+Documentation for the M1 Secrets Provider Push to File feature will include:
+
+- Update the
+  [Secrets Provider init container configuration documentation](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/k8s-ocp/cjr-k8s-secrets-provider-ic.htm).
+- Update the [Secrets Provider application container configuration documentation](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/k8s-ocp/cjr-k8s-secrets-provider-ac.htm).
```
izgeri, line 557: (to document using annotations for config, right?)

should we have an error flow for mixing env var and annotation config?

design/m1_push_to_file_design.md
izgeri: maybe use `path` instead of `dir` for consistency? sounds good in any case; comment about defaults still applies

design/m1_push_to_file_design.md
izgeri: @diverdane I would move forward assuming we will continue supporting this method for now

design/m1_push_to_file_design.md
izgeri: Do we have an error case for when the template is given but no file path is specified? If not, we should add one.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+      - prod/backend/port
+      - prod/backend/password
+      - prod/backend/username
+```
+
+For `conjur-secrets` entries without aliases, the last word in the path is
+used as the application secret name. So for the example above, if
+no other annotations are set for the `database` secret group, then Secrets
+Provider will write to a file `database.yaml` in the `/conjur/secrets/`
+directory (the default value for the conjur.org/secret-file-path.{secret-group}
```
izgeri, line 196:
```suggestion
directory (the default value for the `conjur.org/secret-file-path.{secret-group}`
```

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+| conjur.org/conjur-secrets-path.{secret-group}  | Prefix to use for all Conjur variable paths for a secrets group. An example is `prod/backend/`. The path is assumed to be relative to root policy path. A leading `/` is ignored. A trailing `/` is optional. Defaults to the root policy path. |
+| conjur.org/secret-file-path.{secret-group}     | File path (including file name) for secrets file to be written. This is a required value if the `conjur.org/secret-file-template.{secret-group} annotation is configured for the secret group. The default file name is `{secret-group}.{secret-file-type}`, e.g. `database.yaml`, and the default directory is `/conjur/secrets/`. The directory must match the `mountPath` that is defined for the shared Volume used by the Secrets Provider to write secrets files (refer to [example VolumeMounts below](example-helm-partial-named-template-for-secrets-provider-init-container)) |
```
izgeri, line 173: do we want to support providing a path only here (no filename) and the default filename is used? 

can we add an error case for `secret-file-path` supplied but conflicts with `mountpath`? maybe this is already covered

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+| 10 | File permissions error | You are using SP in "file" mode | Secrets Provider does not have permission to write a secrets file to the specified volume and mount path. | Error: Secrets Provider does not have permission to write the secrets file "{value}". Please check the volume mount permissions on the Secrets Provider container and try again.<br><br>SP fails | | |
+| 11 | Invalid file path provided | You are using SP in "file" mode and have provided a value for the conjur.org/secret-file-path annotation | The provided file path is not a valid file path. | Error: You have provided an invalid file path "{value}" for the "{value}" secret group.<br><br>SP fails to start | | |
+| 12 | File conflict between two secret groups | You are using SP in "file" mode and have defined multiple secret groups | One or more secret groups have he same ouput file. | Error: You have provided conflicting file paths; secret groups "{value}", "{value}", ... all have output file "{value}".<br><br>SP fails to start | | |
+
```
izgeri, line 581:
```suggestion
```suggestion
| 13 | Template provided but no output file defined | You are using SP in "file" mode and have provided a file template definition | You have not also provided an output file path. | Error: The output file path is missing for the "{value}" secret group. File path is required when using a custom template.<br><br>SP fails to start | | |

```

design/m1_push_to_file_design.md
izgeri: added a suggestion.

design/m1_push_to_file_design.md
alexkalish: Agreed.  I assume there are tests for all the error codes?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+container image:
+
+- File permissions: 660         (`rw-rw----`)
+- File owner: secrets-provider  (UID 777)
+- File group: secrets-provider  (GID 777)
+
+If the application container is built to run with a non-root user that has
+different a UID and GID than the default values shown above, then the
+application will not be able to access secrets files that have been
+created using the default attributes.
```
alexkalish, line 675: This worries me a bit.  I thought that containers were provided user IDs in OpenShift?  How do we...
1. Reduce the chance that the app cannot read the file.
2. Support the developer in the rare case where their app cannot read the file.
Reading through the sections below... I'm not sure I can quickly grok our options.  Maybe best we chat separately.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+
+```
+    securityContext:
+      fsGroup: 2000
+```
+
+Note that defining an `fsGroup` will cause the file permissions bits to be
+OR'd with `rw-rw----`.
+
+(For OpenShift deployments, the `fsGroup` may need to be configured as `0`.)
```
alexkalish, line 714: Does OpenShift even support this?  The reason we created a separate enterprise Conjur Follower was to avoid running as root, or specific user IDs.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+| conjur.org/<br />secrets-destination    | SECRETS_DESTINATION | Required value. Allowed values: <ul><li>`file`</li><li>`k8s_secret`</li></ul> |
+| conjur.org/<br />k8s-secrets            | K8S_SECRETS         | This list is ignored when `conjur.org/secrets-destination` annotation is set to **`file`** |
+| conjur.org/<br />retry-count-limit      | RETRY_COUNT_LIMIT   | Defaults to 5
+| conjur.org/<br />retry-interval-sec     | RETRY_INTERVAL_SEC  | Defaults to 1 (sec)              |
+| conjur.org/<br />debug-logging          | DEBUG               | Defaults to `false`              |
+| **conjur.org/<br />connect-config-dir** | CONJUR_AUTHN_URL<br />CONJUR_URL<br />CONJUR_ACCOUNT<br />CONJUR_SSL_CERTIFICATE | Directory containing Conjur connection information. Connection parameters are written to files with names `CONJUR_AUTHN_URL`, `CONJUR_URL`, `CONJUR_ACCOUNT`, and `CONJUR_SSL_CERTIFICATE`. The directory must match the `mountPath` used for the Conjur Connect ConfigMap as shown in the [example Kubernetes manifests](example-helm-partial-named-template-for-secrets-provider-init-container) below. The default value is `/conjur/connect`. |
```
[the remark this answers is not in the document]
izgeri: **Decision from the sign off call**: Secrets Provider will use default values for the “secrets”, “connect”, and “podinfo” volume mounts. We do not need to support configuring non-default values for the volume mount paths in this phase of the project; if there are use cases that require this, we may add this functionality later.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1302 @@
...
+| conjur.org/secret-file-format.{secret-group}     | Allowed values:<ul><li>yaml (default)</li><li>json</li><li>dotenv</li><li>bash</li></ul>This setting is ignored when `conjur.org/secret-file-template.{secret-group}` is configured. |
+| conjur.org/secret-file-perms.{secret-group}    | File permissions for secrets file to be written, in octal format. Defaults to `660` |
```
izgeri, line 169: We'd like to note here that this is future work, and that it defaults to 644

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1302 @@
...
+- [ ] There is a happy path e2e test in the Secrets Provider test cases to validate
+  writing of Bash formatted file
+- [ ] There is a happy path e2e test in the Secrets Provider test cases to validate
+  writing of dotenv formatted file
+- [ ] There is a happy path e2e test to validate Helm named templates for init
+  container def, volume mounts, and volumes
+- [ ] There is a blog post describing how to transition from an existing application to using the new Secrets Provider "push to file" flow
+- [ ] There is a recorded demo of the new SP "file" functionality
+- [ ] The new SP "file" functionality has been shared at a training session
+
```
izgeri, line 925: Future work: 
- [ ] Add ability to configure secret file permissions via `conjur.org/secret-file-perms.{secret-group}` annotation

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1302 @@
...
+</table>
+
+## Background
+
+### Secrets Provider Push to File Support
+
+The goal of the Milestone 1 (M1) Push-to-File initiative is to enhance the
+Secrets Provider init container integration with an option for writing
+secrets values that have been retrieved from Conjur into files that can be
+directly accessed by an application that is running in the same Pod.
```
sigalsax, line 76: In SP we also have the ability to deploy it as an application container so that it exists in the user's same namespace but isn't coupled with their actual application container as with the init container solution. How easy can Push to File be applied to the application container flow? Is this something we see value in?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+```
+url: https://database.example.com
+admin-password: my-secret-p@$$w0rd
+admin-username: admin
+```
+
+###### Example `conjur.org/secret-file-template.{secret-group}`
+
+The following is an example of a custom secrets file template that uses the
+Golang [`text/template`](https://golang.org/pkg/text/template/) syntax.
```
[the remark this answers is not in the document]
diverdane: I've added a couple of examples, a simple YAML output example, and an example that iterates through secrets (without explicit naming), and uses one of the secrets as a variable suffix for the secrets names generated in the output secrets file.

I also added a note in the table to say that predefined Golang template functions should be okay. It won't be practical for us to add tests for all possible Golang template functions, but we should test a representative sample of functions.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+#### Security testing
+
+Security testing will include:
+
+- Automated Vulnerability scans
+
+####  Performance testing
+
+As described in the [Performance](#performance) section above, the performance
+of the Secrets Provider for Push-to-File mode will be manually measured using
```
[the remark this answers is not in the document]
diverdane: I can add the instrumentation using Prometheus metrics as a nice-to-have. @doodlesbykumbi might have a better idea of the cost/scope of adding this, but I would guess that it's about 1 person-sprint to develop the instrumentation, and then another 1 person-sprint to document and add unit/integration tests.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+- Latency for writing retrieved secrets to files
+
+_**QUESTION: Do we need to measure CPU and memory footprints?**_
+
+The measurements will be made using a representative configuration in terms
+of the number of secrets groups and the number of secrets per secrets group.
+Details are TBD, but a representative configuration might look something like
+this:
+
+- 5 secrets groups
```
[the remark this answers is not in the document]
diverdane: @alexkalish This is good information. I added this to the `Project Scope and Limitations` section, and mentioned that the number of Conjur secrets used in performance testing:
```
      5 Conjur secrets x 5 secrets groups = 25 Conjur secrets total
```
represents about half of the Secrets Provider's scaling limit.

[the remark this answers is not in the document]
diverdane: @andytinkham , We'll only have one Secrets Provider init container per application Pod, so technically speaking there won't be multiple Secrets Providers writing to the secrets files.

However, your point is valid that we need to account for the potential for multiple "file writers" (with one file writer per configured secrets group) writing to the same secrets file.

For the short term, I'm assuming that we can consider a configuration that uses the same destination secrets file name for multiple groups to be a (fatal) misconfiguration. We'll add a validation check for this after all annotations have been parsed.

I've also added a note that **in the future**, we may consider adding concurrency protection mechanisms that would allow for multiple file writers (i.e. multiple secrets groups) to append secrets settings to a common secrets file. This is technically possible to implement, but I don't know if there are a lot of use cases where this would be required.

design/m1_push_to_file_design.md
diverdane: I reworded this entire section, based on our discussion. Secrets Provider will set file permissions to `664` or `rw-rw-r--` for the secrets files to avoid mismatches when a deployment is configured to run containers in a Pod using an arbitrary UID and GID at run time. The security picture for using these permissions should be analogous to what is done for OpenShift. (I've added all of these details to the Security section.)

design/m1_push_to_file_design.md
diverdane: Reworded per our discussion to describe why this isn't a concern for OpenShift.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+| conjur.org/secret-file-path.{secret-group}     | File path for secrets file to be written. This path can be a directory (any value ending in `/`, Service Provider will use the default file name of `{secret-group}.{secret-file-type}` for the secret group) or a full path including a file name. The default value is `/conjur/secrets/`. The directory portion of the path must match the `mountPath` that is defined for the shared Volume used by the Secrets Provider to write secrets files (refer to [example VolumeMounts below](#example-helm-partial-named-template-for-secrets-provider-init-container)). Note that a full path including a file name is required value if the `conjur.org/secret-file-template.{secret-group}` annotation is configured for the secret group. |
+| conjur.org/secret-file-type.{secret-group}     | Allowed values:<ul><li>yaml (default)</li><li>json</li><li>dotenv</li><li>bash</li></ul>This setting is ignored when `conjur.org/secret-file-template.{secret-group}` is configured. |
+| conjur.org/secret-file-perms.{secret-group}    | File permissions for secrets file to be written, in octal format. Defaults to `660` |
```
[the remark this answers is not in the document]
diverdane: Per our discussion, I've deleted this permissions knob. Secrets files will be created with permissions `664` octal (or `rw-rw-r--`).  I've added a description of this approach to the Security section in the document.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+the [Performance testing](#performance-testing) section below. The performance
+measurements will include the following metrics:
+
+- Latency (processing time) for parsing Pod annotations
+- Latency for generating secrets requests and processing responses
+- Latency for writing retrieved secrets to files
+
+The measurements will be made using a representative configuration (details
+TBD) in terms of the number of secrets groups and the number of secrets per
+secrets group.
```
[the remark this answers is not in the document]
diverdane: I think that's a broader concern that applies to any of our current integrations. I don't think the push-to-file feature makes a significant change in potential request rates.

design/m1_push_to_file_design.md
diverdane: Unfortunately, this wouldn't be easy. The thing that makes Push to File viable when running the Secrets Provider as a container within the same Pod as the application is that they can share a Volume (SP can write a file, the application can directly read it).
Adding this functionality to the Secrets Provider running as a standalone application container would require a remote file-writing functionality, something akin to creating a Kubernetes CSI driver (see https://kubernetes.io/blog/2019/01/15/container-storage-interface-ga/). This would be possible, but one tricky part would be figuring how to allow the Secrets Provider (running as a CSI agent) to sort of proxy the machine identity of the application for which its retrieving from Conjur (to then write to a CSI volume).

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+| 8 | Unable to retrieve Conjur variables | You are using SP in "file" mode | | Error: Failed to provide DAP/Conjur secrets<br><br>SP fails | CSPFK016E | Unit | |
+| 9 | Missing volume mounts | You are using SP in "file" mode | One or more volumes cannot be found by Secrets Provider. | Error: Unable to access volume "{value}". Error: "{value}". Ensure that the correct volumes are defined in the deployment manifest and attached to the Secrets Provider container.<br><br>SP fails to start | | Unit | US-07 |
+| 10 | File permissions error | You are using SP in "file" mode | Secrets Provider does not have permission to write a secrets file to the specified volume and mount path. | Error: Secrets Provider does not have permission to write the secrets file "{value}". Please check the volume mount permissions on the Secrets Provider container and try again.<br><br>SP fails | | Unit | US-07 |
```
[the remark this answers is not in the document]
diverdane: I'll add a test case.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+
+#### Integration / E2E tests
+
+E2E tests are expensive. We should try to limit their numbers. There are cheaper ways to validate behavior. For the push to file functionality it will likely be possible to validate the happy path in one fell swoop by using configuration tht captures all the supported annotations and template types.
+
+- [ ] Happy path, Secret Provider runs as an init container in file mode and successfully injects files. This single test can cover all supported annotations and template types.
+
+The sad path is where things tend to become difficult, because there's a myriad of ways in which failure can occur. A possible way to reduce the need to E2E test failures is to establish a nexus point for all failures in the code so that an error can be validated at the unit/functional testing level.
+
+- [ ] Sad path, Secret Provider runs as an init container in file mode and fails, thereby preventing the deployment from proceeding. TBC it's possible we don't need this at all.
```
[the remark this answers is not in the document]
diverdane: These steps:

- Secrets file rendering (retrieving Conjur secrets and generating file content)
- Writing rendered content to disk

will be done in parallel for each secrets group (as parallel Go functions). So it's entirely possible that a secrets file could be written, and then a rendering/writing error is encountered for another secrets group.

However:
- Any secrets files that might have gotten written will be in ephemeral files (`emptyDir` Volume)
- Rendering/Writing errors are considered fatal, so SP will abort and restart.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+
+#### Security testing
+
+Security testing will include:
+
+- Automated Vulnerability scans
```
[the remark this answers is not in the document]
diverdane: This probably depends heavily on how the secrets files are consumed by the application.

For `yaml` or `json` file formats, I don't know if there are best practices that we can recommend e.g. to escape secrets values or the file content to avoid code injection or XSS?

For `bash` or `dotenv`, we can probably recommend that the application does not consume these using `eval`.

For custom templates, customers will be providing their own templates and will be rendering the templates using their own Conjur secrets values. In this case, there isn't much control over what they implement... unless we implemented some highly complex mechanism to look for potential security vulnerability patterns in the templates that are provided. We can at least document some best practices e.g. for escaping HTML text before writing secrets file content to disk.

diverdane:
@alexkalish , @andytinkham, @sigalsax:

I believe I've addressed all of the review comments, and it's ready for re-review.
The biggest differences since you last reviewed:
- Added more example custom templates
- Added details on validation that will be performed
- Security section (near the end):
   - Reworded the discussion on running containers with arbitrary users
   - Added discussion on how we'll avoid leaking secrets values while rendering secrets files
- Added section for the boilerplate reduction
- Added many more test cases
- Reworded the `Logs` section

The section that probably deserves the most scrutiny is the `Security` section near the end of the document.

I think everything is in good shape, although there may be a few tasks to add to the development task list.

Thank you,
Dane

design/m1_push_to_file_design.md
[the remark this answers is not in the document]
alexkalish: Agreed.  I like being this explicit and failing early if the path isn't right.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1247 @@
...
+   "admin-password": "p@$$w0rd",
+   "admin-username": "zappa"
+}
+```
+
+##### Example `bash` Secrets File
+
+```
+   export URL="https://database.example.com"
+   export ADMIN_PASSWORD="p@$$w0rd"
```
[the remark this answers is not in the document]
alexkalish: Oh, I like @andytinkham idea about validating the resulting file to avoid any issues.  Doesn't feel required for MVP  though.  Maybe just take note of it, @diverdane?

design/m1_push_to_file_design.md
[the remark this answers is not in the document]
alexkalish: Great.  Lets go that direction and only dial it back to manual if we don't have another choice.

design/m1_push_to_file_design.md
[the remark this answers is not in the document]
alexkalish: My table of errors in the Feature Doc calls out this case as an error.  We should fail early if two secret groups specify the same file path.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1769 @@
...
+| 10 | Parsing of relative secrets file path | "secret-file-path" annotation set to subdirectory + filename | Parsing is carried out | The secret file is written to `/conjur/secrets/{secret-file-path}` | US-07 |
+| 11 | Parsing of relative secrets file path | "secret-file-path" annotation set to subdirectory + filename with extension not matching file format | Parsing is carried out | The secret file is written to `/conjur/secrets/{secret-file-path}` | US-07 |
+| 12 | Default value for secrets file type | "secret-file-format" annotation is not set | Parsing is carried out | The default secrets file type set to `yaml` | US-08 |
+| 13 | Parsing of secrets file type | Valid "secret-file-format" annotation (`yaml`, `json`, `dotenv`, `bash`) | Parsing is carried out | The internal representation matches expectations  | US-08 |
+| 14 | Parsing of secrets file type | Valid "secret-file-format" annotation (`yaml`, `json`, `dotenv`, `bash`) | Parsing is carried out | The internal representation matches expectations  | US-08 |
```
andytinkham, line 952: Are these identical?

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+| 2 | Invalid input file type | You are using SP in "file" mode | You have provided an unknown or unsupported type in the secret-file-type annotation | Error: Invalid output file type: "{value}" for secret group "{value}". Acceptable values are "json", "yaml", and "dotenv".<br><br>SP fails to start | | Unit | |
+| 3 | Conjur variable names collision | You are using SP in "file" mode | Two or more secrets in the same Secret Group share the same name or alias. | Error: Multiple variables in the "{value}" secret group are called "{value}". Provide a unique alias for each variable in the "{value}" secret group.<br><br>SP fails to start | | Unit | |
+| 4 | Unparseable Conjur secrets list | You are using SP in "file" mode | The "conjur-secrets" annotation is not parseable for one or more secret groups. | Error: The list of secrets for the "{value}" secret group is not formatted correctly. Error: "{value}". Verify that the annotation value is a YAML list with optional keys for each list item.<br><br>SP fails to start | | Unit | |
```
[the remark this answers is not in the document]
andytinkham: If that's a possible scenario, do we need an empty secrets list test in the unit tests on the positive side then?

andytinkham (approved):
I added a few questions, but I don't think any of them are blockers. Kumbi, Dane & I have a discussion tomorrow about templates leaking secrets, but if that leads to needing a change, I'll do another review. With what's in the doc, I'm OK with this plan.

alexkalish (approved):
Solution approved!  @diverdane: who else needs to approved?  Can we try to get this done by EOW?

design/m1_push_to_file_design.md
diverdane: Yes, deleted duplicate.

design/m1_push_to_file_design.md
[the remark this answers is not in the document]
diverdane: Added a UT case for empty secrets list.

design/m1_push_to_file_design.md
```diff
@@ -0,0 +1,1280 @@
...
+| 11 | Invalid file path provided | You are using SP in "file" mode and have provided a value for the conjur.org/secret-file-path annotation | The provided file path is not a valid file path. | Error: You have provided an invalid file path "{value}" for the "{value}" secret group.<br><br>SP fails to start | | Unit | US-07 |
+| 12 | File conflict between two secret groups | You are using SP in "file" mode and have defined multiple secret groups | One or more secret groups have he same ouput file. | Error: You have provided conflicting file paths; secret groups "{value}", "{value}", ... all have output file "{value}".<br><br>SP fails to start | | Unit | US-07 |
+
+#### Integration / E2E tests
+
+E2E tests are expensive. We should try to limit their numbers. There are cheaper ways to validate behavior. For the push to file functionality it will likely be possible to validate the happy path in one fell swoop by using configuration tht captures all the supported annotations and template types.
+
+- [ ] Happy path, Secret Provider runs as an init container in file mode and successfully injects files. This single test can cover all supported annotations and template types.
```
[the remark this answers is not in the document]
diverdane: Hmmm... if I understand what your asking for here, that's going to be tricky. One theme in our push-to-file design is that we fail early and fail "hard" when errors occur. Changing that behavior just for the E2E happy-path testing would be difficult.

As a counter argument, the E2E happy-path testing will be run after all of the UT and IT tests have passed, so failures should be rare, except for situations where external dependencies have changed their behavior, or external dependencies aren't behaving the way we expected.

diverdane:
@alexkalish @andytinkham , 
Thank you for the approvals! I just wanted to mention that post-approval, I made a few changes based on discussions yesterday and some recent comments:

- **Preventing leaking of sensitive info during custom template execution**: Reworded this section to say that we'll include a "test run" of `template.Execute()` using fake (i.e. safe) values (as suggested by @doodlesbykumbi ), and following that, we'll do a real run of `template.Execute()` using real data with raw output redirected to `/dev/null`. I also included alternatives to this approach.
- **Performance metrics**: Included some details about supporting an HTTP metrics endpoint for scraping with Prometheus. I included a note saying that this is a "best effort" featurette for the early GA phase, and that as a contingent we'd do some manual instrumentation to measure latencies.
- **Custom template function support**: Added comment that initially we'd only support the ["built-in" Go template functions](https://golang.org/pkg/text/template/#hdr-Functions), and that in the future, we'd add support of functions e.g. in the `strings` standard Go module or in the [sprig Go template module](http://masterminds.github.io/sprig/). These would be added to a `funcMap` on a function-by-function basis (essentially an access list of supported functions).

PR #324, opened 2021-05-31
fix openshift testing

eladkug:
### What does this PR do?
Docker changes the way it works with mount and we cant mount with empty string.
I separate between kubernetes and openshift docker command so we will not have empty string as part of mounting. 

#### Change log

#### Test coverage

#### Documentation


eladkug:
I had some problem. Will open new PR instead.

PR #325, opened 2021-05-31
Fix openshift testing

eladkug:
### What does this PR do?
Docker changes the way it works with mount and we cant mount with empty string.
I separate between kubernetes and openshift docker command so we will not have empty string as part of mounting.

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/deploy/utils.sh
+++ b/deploy/utils.sh
@@ -88,51 +88,94 @@ get_conjur_cli_pod_name() {
 }
 
 runDockerCommand() {
-  docker run --rm \
-    -i \
-    -e UNIQUE_TEST_ID \
-    -e CONJUR_APPLIANCE_IMAGE \
-    -e CONJUR_LOG_LEVEL \
-    -e CONJUR_FOLLOWER_COUNT \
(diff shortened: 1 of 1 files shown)


deploy/utils.sh
```diff
@@ -88,51 +88,94 @@ get_conjur_cli_pod_name() {
...
+      -e OPENSHIFT_USERNAME \
+      -e OPENSHIFT_PASSWORD \
+      -e DOCKER_REGISTRY_PATH \
+      -e DOCKER_REGISTRY_URL \
+      -e PULL_DOCKER_REGISTRY_PATH \
+      -e PULL_DOCKER_REGISTRY_URL \
+      -e GCLOUD_CLUSTER_NAME \
+      -e GCLOUD_ZONE \
+      -e GCLOUD_PROJECT_NAME \
+      -e GCLOUD_SERVICE_KEY=/tmp$GCLOUD_SERVICE_KEY \
```
orenbm, line 118: i see that the difference between k8s and ocp are these vars. can you put them at the bottom?

orenbm (approved):

PR #352, opened 2021-08-11
Adds Helm chart option to set Secrets Provider Job name

diverdane:
### What does this PR do?

Currently, the Secrets Provider Helm chart deploys a Secrets Provider
Kubernetes Job with a name that is (unconditionally) set to the Helm Release
name. This should work fine in most cases.

In the case where the Secrets Provider Helm chart is included as a
[Helm dependency](https://helm.sh/docs/helm/helm_dependency) to another
(parent) Helm chart, however, the Secrets Provider Helm chart inherits its Release
name from its parent chart. This currently results in the Secrets Provider
Job name being set to the Release name of the **parent** Helm chart, which
is a bit confusing. This is a scenario we're seeing in adding a workflow
to test Secrets Provider as a standalone job with the
[Conjur authn-k8s client E2E workflow Helm charts](https://github.com/cyberark/conjur-authn-k8s-client/tree/master/helm/conjur-app-deploy). 

This change adds a chart value to allow the Secrets Provider Kubernetes Job
to be explicitly set, if desired. If this chart value is not set explicitly,
then the Secrets Provider Kubernetes Job name will default to the chart's
Release name.

This change also includes a couple of Helm unittest test cases to test
the precedence ordering for the SP Job name.

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/helm/secrets-provider/templates/secrets-provider.yaml
+++ b/helm/secrets-provider/templates/secrets-provider.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ .Release.Name }}
+  name: {{ .Values.secretsProvider.jobName | default .Release.Name }}
   namespace: {{ .Release.Namespace }}
 {{- with .Values.labels }}
   labels:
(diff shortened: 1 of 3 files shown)


imheresamir (approved):
👍

PR #355, opened 2021-08-19
Avoid dockerhub pull limits in Openshift for sleeping container

doodlesbykumbi:
### What does this PR do?

Change sleeping container image from 'debian' to the image of the secrets provider. This avoids having to push another container and also avoids worrying about dockerhub pull limits. The test app container really just need 
1. to have `sleep` so it can initially sleep
2. to have `cut`, `grep` and `printenv` for making test assertions

Resolves -

Sometimes the build is failing due to errrors of the form
```
Warning  Failed          1m (x3 over 4m)   kubelet, ip-10-0-158-137.ec2.internal  Failed to pull image "debian": rpc error: code = Unknown desc = Error reading manifest latest in docker.io/library/debian: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/deploy/config/k8s/test-env.sh.yml
+++ b/deploy/config/k8s/test-env.sh.yml
@@ -31,7 +31,7 @@ spec:
     spec:
       serviceAccountName: ${APP_NAMESPACE_NAME}-sa
       containers:
-      - image: debian
+      - image: '${PULL_DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/debian:latest'
         name: test-app
         command: ["sleep"]
         args: ["infinity"]
--- a/deploy/config/openshift/test-env.sh.yml
+++ b/deploy/config/openshift/test-env.sh.yml
@@ -30,7 +30,7 @@ spec:
     spec:
       serviceAccountName: ${APP_NAMESPACE_NAME}-sa
       containers:
-      - image: debian
+      - image: '${PULL_DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/debian:latest'
         name: test-app
         command: ["sleep"]
         args: ["infinity"]
--- a/deploy/run_with_summon.sh
+++ b/deploy/run_with_summon.sh
@@ -50,6 +50,11 @@ if [ "${DEV}" = "false"  ]; then
     "${DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/secrets-provider"
 
   docker push "${DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/secrets-provider"
+
+  # Make sure debian is pushed to the internal registry to avoid Dockerhub pull restrictions
+  docker pull debian:latest
+  docker tag debian:latest "${DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/debian:latest"
+  docker push "${DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/debian:latest"
 else
   docker tag "secrets-provider-for-k8s:dev" \
     "${APP_NAMESPACE_NAME}/secrets-provider"


deploy/config/openshift/test-env.sh.yml
```diff
@@ -30,7 +30,8 @@ spec:
     spec:
       serviceAccountName: ${APP_NAMESPACE_NAME}-sa
       containers:
-      - image: debian
+      - image: '${PULL_DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/secrets-provider:latest'
```
sigalsax, line 33: It may not really matter because this is just a test, dummy container but was thinking that because these are testing, should we be using the one we built as part of the run `secrets-provider-for-k8s:latest` (see `build`)?
Also this image appears in k8s/test-env.sh.yml and also secrets-provider-init-container.sh.yml so which ever you choose, shouldn't we also be percing the change to them as well?

doodlesbykumbi: My understanding is that this is purely just a test container. It only needs to satisfy the following:
1. to have `sleep` so it can initially sleep
2. to have `cut`, `grep` and `printenv` for making test assertions
Debian was chosen because it meets these requirements. Incidentally the secrets-provider container image meets these requirements too.

I opted to use the secrets-provider container image because it is already being pushed to the OpenShift registry as part of the pipeline. The alternative is to push debian, which honestly might be the more honest approach.

> Also this image appears in k8s/test-env.sh.yml and also secrets-provider-init-container.sh.yml 
The reason why those are being left out is because the dockerhub pull limits don't seem to be an issue in k8s, only in OpenShift.

diverdane: Another option might be to use `cyberark/conjur-cli:5-latest` for the test container, since that will be getting pulled already when `cyberark/kubernetes-conjur-deploy` scripts are cloned and run, and `conjur-cli` has the `cut`, `grep`, and `printenv`.  It might be more palatable using `conjur-cli` to test rather than secretless-provider?

Or, if we get a team DockerHub account for a bigger download rate? It may come to that eventually.

doodlesbykumbi: @diverdane

It's true that `cyberark/conjur-cli:5-latest` is already available but the issue there is that we have to inspect the internals of `cyberark/kubernetes-conjur-deploy` to determine what the image name is, and in doing so we couple the 2 repos together. If that image name changes in the deploy scripts then that impacts this one.

Infra suggest pushing images into the repository as the go to vs using a Dockerhub account.

To be honest, pushing `debian` isn't all that involved. I think I'll just go in that direction

deploy/config/k8s/test-env.sh.yml
```diff
@@ -31,7 +31,7 @@ spec:
     spec:
       serviceAccountName: ${APP_NAMESPACE_NAME}-sa
       containers:
-      - image: debian
+      - image: '${PULL_DOCKER_REGISTRY_PATH}/${APP_NAMESPACE_NAME}/debian:latest'
```
sigalsax, line 34: I think we also need this in `secrets-provider-init-container.sh.yml`

sigalsax (approved):
One comment but other than that, LGTM!

deploy/config/k8s/test-env.sh.yml
doodlesbykumbi: I thought so but turns out that manifest is entirely for development purposes so we can't make assumptions about the image registry. You'll see that `secrets-provider-init-container.sh.yml` is pulling in `debian` directly as well as `secrets-provider-for-k8s:latest`, see https://github.com/cyberark/secrets-provider-for-k8s/blob/main/deploy/dev/config/k8s/secrets-provider-init-container.sh.yml#L50. So we can leave that one out.

PR #358, opened 2021-09-02
Adding initial directories for push to file

### LINKED ISSUE #342: There is a defined project structure for M1 push to file
As a M1 developer,
I want a defines project structure,
So that development can be coordinated between developers.

For the M1 push to file, the general project layout should be defined.

- There should be a directory(s) where the components should be places.
   Verify if placing in pkg is the best location.
- Subdirectories for each component should be defined?

## Acceptance Criteria

- [ ] Subdirectories are created and the project layout is defined.

rpothier:
### What does this PR do?

This lays out the initial file structure for push to file.
 

Resolves #342 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/pkg/secrets/annotations/README.md
+++ b/pkg/secrets/annotations/README.md
@@ -1,0 +1,4 @@
+This directory will be used to define a package that provides generic parsing
+of Annotations files that have been created by Kubernetes via the Kubernetes
+Downward API. This package will read in files that contain entries of the form
+`<key>`=`<value>`, and will create a Golang map of the form map[string]string.
--- a/pkg/secrets/push_to_file/README.md
+++ b/pkg/secrets/push_to_file/README.md
@@ -1,0 +1,9 @@
+This directory will be used to define a package for the Secrets Provider
+"Push to File" functionality. It will contain source code to do the following:
+
+- Parse the keys used in push-to-file annotations (e.g. sort annotations based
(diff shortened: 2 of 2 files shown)


diverdane (commented):
Thanks for getting the ball rolling on this project structure!
I think what we can do to get this merged as a placeholder is:
- Create two new directories for packages:
   - pkg/secrets/push_to_file
   - pkg/secrets/annotations
- Since GitHub won't allow us to merge empty directories, either temporarily add an empty `.keep` file into each directory (see https://stackoverflow.com/questions/115983/how-can-i-add-a-blank-directory-to-a-git-repository), or add temporary `README.md` files in each, saying e.g.:

_This directory will be used to define a package for the Secrets Provider "Push to File" functionality. It will contain source code to do the following:_
- Parse the keys used in push-to-file annotations (e.g. sort annotations based on secrets group)
- Retrieve Conjur secrets for each secrets group and write retrieved values to a file info data struct
- Write secrets files to a shared volume

...and...

_This directory will be used to define a package that provides generic parsing of Annotations files that have been created by Kubernetes via the Kubernetes Downward API. This package will read in files that contain entries of the form `<key>=<value>`, and will create a Golang map of the form `map[string]string`._

diverdane (approved):
LGTM!

PR #362, opened 2021-09-23
Parse Secrets Groups

imheresamir:
### What does this PR do?
Parses Push-to-File annotations into `SecretsGroups` object ready for secrets reading and file writing.

ONYX-12725

#### Change log

#### Test coverage

#### Documentation


pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+
+	t, err := template.New("FileTemplate").Parse(fileTemplate)
+	if err != nil {
+		return nil, log.RecordedError(messages.CSPFK054E, fmt.Sprintf("%s%s", SECRETS_GROUP_FILE_TEMPLATE_PREFIX, groupName), err.Error())
+	}
+
+	return t, nil
+}
+
+func validateGroupInfo(groupName string, groupInfo SecretsGroupInfo) error {
```
diverdane, line 227: A comment would help here describing what's being detected, e.g. something along the lines that if a template is specified, then a secrets file name is required, not just a directory.

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+	if groupInfo.FileTemplate != nil {
+		if strings.HasSuffix(groupInfo.FilePath, "/") {
```
diverdane, line 229: I think we need a check for the `FilePath` being unspecified (i.e. a null string) (in addition to the check for trailing `/`)?

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+		return FILE_TYPE_NONE, log.RecordedError(messages.CSPFK053E, fmt.Sprintf("%s%s", SECRETS_GROUP_FILE_TYPE_PREFIX, groupName), fileType)
+	}
+}
+
+func parseFileTemplate(groupName string, fileTemplate string) (*template.Template, error) {
+	if fileTemplate == "" {
+		return nil, nil
+	}
+
+	t, err := template.New("FileTemplate").Parse(fileTemplate)
```
diverdane, line 219: The parsing of the raw template string should probably done separately, e.g. in a separate file that does the business of Golang parsing and execution.

pkg/secrets/push_to_file/secrets_groups_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+type extractSecretsGroupsTestCase struct {
+	description string
+	contents    map[string]string
+	assert      func(t *testing.T, result SecretsGroups, err error)
+}
+
+func assertValidSecretsGroups(expected SecretsGroups) func(*testing.T, SecretsGroups, error) {
```
diverdane, line 15: Function is named `assertValidSecretsGroup`, but maybe it should be something like `assertExpectedSecretsGroup`? i.e. it's checking that the results are expected, not just that they're valid?

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	}
+
+	return secretsGroups, nil
+}
+
+func parseSecretsPaths(groupName string, secretsPaths string, secretsPathsPrefix string) (SecretsPaths, error) {
+	secrets := make(SecretsPaths)
+
+	unpacked := []interface{}{}
```
diverdane, line 117: How does this implementation compare to what @doodlesbykumbi had in his PoC?

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+		secrets[name] = secretsPathsPrefix + secretPath
+
+		return nil
+	}
+
+	for _, secret := range unpacked {
+		switch val := secret.(type) {
+		case string:
+			{
+				name := val[strings.LastIndex(val, "/")+1:]
```
diverdane, line 136: Maybe using `path.Base()` would be simpler for extracting the last word of the variable path?

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	if len(secrets) == 0 {
+		return nil, log.RecordedError(messages.CSPFK051E, "No valid secrets found", groupName)
+	}
+
+	return secrets, nil
+}
+
+func parseSecretsPathPrefix(groupName string, secretsPathPrefix string) (string, error) {
```
diverdane, line 168: A comment describing what's being done here would help, e.g. it defaults to the root policy path `/`, otherwise adds a trailing `/` if necessary, or something like that.

diverdane, line 168: Can we put `Conjur` somewhere in the function name? Initially I was confusing the `SecretsPathPrefix` portion of the function name to mean it's a path prefix for the destination secrets file.

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+
+	return secretsPathPrefix + "/", nil
+}
+
+func parseFilePath(groupName string, filePath string) (string, error) {
+	if strings.HasPrefix(filePath, "/") {
+		return "", log.RecordedError(messages.CSPFK052E, fmt.Sprintf("%s%s", SECRETS_GROUP_FILE_PATH_PREFIX, groupName))
+	}
+
+	if strings.HasSuffix(filePath, "/") {
```
diverdane, line 182: Maybe add a comment saying the file path can be a directory (ending in `/`) or a file path including a file name (not ending in `/`)?

diverdane (commented):
Here are my comments so far. Still going through the UT.

pkg/secrets/push_to_file/secrets_groups_test.go
```diff
@@ -0,0 +1,178 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type extractSecretsGroupsTestCase struct {
+	description string
+	contents    map[string]string
+	assert      func(t *testing.T, result SecretsGroups, err error)
```
diverdane, line 12: Your call, but you could simplify a tiny bit by declaring an `assertFunc` type:
```
 type assertFunc func(*testing.T, SecretsGroups, error)         
 ```                                                                             
 And then the `extractSecretsGroupsTestCase` struct would be:
 ```                                                                           
 type extractSecretsGroupsTestCase struct {                                      
     description string                                                          
     contents    map[string]string                                               
     assert      assertFunc                                                      
 }
 ```
 and the definition of the `assertValidSecretsGroups` function signature would be simplified:
 ```
 func assertValidSecretsGroups(expected SecretsGroups) assertFunc {
 ```
 and similar for assertProperError:
 ```
func assertProperError(expectedErr string) assertFunc {
```

imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
imheresamir: Done

imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups_test.go
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
imheresamir: Done

pkg/log/messages/error_messages.go
```diff
@@ -56,3 +56,10 @@ const CSPFK037E string = "CSPFK037E Failed to parse DAP/Conjur variable IDs"
 // General
 const CSPFK038E string = "CSPFK038E Retransmission backoff exhausted"
 const CSPFK039E string = "CSPFK039E Secrets Provider for Kubernetes failed to update Kubernetes Secrets"
+
+// Push to File
+const CSPFK051E string = "CSPFK051E Failed to unmarshal Push-to-File secrets. Reason: %s in group '%s'"
+const CSPFK052E string = "CSPFK052E Annotation '%s' must contain relative path without leading '/'"
+const CSPFK053E string = "CSPFK053E Annotation '%s' specifies unknown file format '%s'"
+const CSPFK054E string = "CSPFK054E Failed to parse template in annotation '%s'. Reason: %s"
+const CSPFK055E string = "CSPFK055E Push-to-File template specified but directory found in annotation '%s'"
```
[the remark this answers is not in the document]
imheresamir: Done, let me know if you have verbiage suggestions

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+				FilePerms:    DEFAULT_FILE_PERMS,
+				FileTemplate: fileTemplate,
+			}
+
+			err = validateGroupInfo(groupName, groupInfo)
+			if err != nil {
+				return nil, err
+			}
+
+			/*
```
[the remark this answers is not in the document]
imheresamir: Removed - don't worry just a manual testing remnant

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+const SECRETS_GROUP_FILE_TYPE_PREFIX = "conjur.org/secret-file-format."
+const SECRETS_GROUP_FILE_TEMPLATE_PREFIX = "conjur.org/secret-file-template."
+
+const DEFAULT_FILE_PERMS os.FileMode = 0777
+const DEFAULT_FILE_PATH string = "/conjur/secrets/"
+
+type SecretsFileType int
+
+const (
+	FILE_TYPE_NONE SecretsFileType = iota
```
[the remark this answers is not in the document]
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+	"path"
+	"strings"
+	"text/template"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
+	"gopkg.in/yaml.v3"
+)
+
+const SECRETS_GROUP_PREFIX = "conjur.org/conjur-secrets."
```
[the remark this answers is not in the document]
imheresamir: Done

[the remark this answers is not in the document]
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
+package secrets_groups
```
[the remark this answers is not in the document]
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
imheresamir: Removed

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+type SecretsPaths map[string]string
+
+// SecretsGroupInfo comprises secrets mapping information for a given secrets
+// group.
+type SecretsGroupInfo struct {
+	Secrets      SecretsPaths
+	FilePath     string
+	FileType     SecretsFileType
+	FilePerms    os.FileMode
+	FileTemplate *template.Template
```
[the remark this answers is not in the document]
imheresamir: Changed

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+	FilePath     string
+	FileFormat   SecretsFileFormat
+	FilePerms    os.FileMode
+	FileTemplate string
+}
+
+// SecretsGroups comprises secrets mapping info for all secret groups
+type SecretsGroups map[string]SecretsGroupInfo
+
+func ExtractSecretsGroupsFromAnnotations(annotations map[string]string) (SecretsGroups, error) {
```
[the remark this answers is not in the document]
imheresamir: Looks straightforward to me...

diverdane (commented):
Okay, I'm finally done reviewing everything.

Very nice work with the table-driven testing and the use of closure for the assertion functions!

I'm ready to approve, but I'd like to hold off for now and let @doodlesbykumbi get a chance to review.

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+	FILE_FORMAT_YAML
+	FILE_FORMAT_JSON
+	FILE_FORMAT_DOTENV
+	FILE_FORMAT_BASH
+	FILE_FORMAT_PLAINTEXT
+)
+
+// SecretsPaths comprises Conjur variable paths for all secrets in a secrets
+// group, indexed by secret name.
+type SecretsPaths map[string]string
```
rpothier, line 38: So is this a map of the secret alias to the full path? Would it be better to have this as a struct
with the ID and alias 
`type SecretSpec struct {
	Id    string
	Alias string
}`

pkg/log/messages/error_messages.go
```diff
@@ -56,3 +56,9 @@ const CSPFK037E string = "CSPFK037E Failed to parse DAP/Conjur variable IDs"
 // General
 const CSPFK038E string = "CSPFK038E Retransmission backoff exhausted"
 const CSPFK039E string = "CSPFK039E Secrets Provider for Kubernetes failed to update Kubernetes Secrets"
+
+// Push to File
+const CSPFK051E string = "CSPFK051E Failed to unmarshal Push-to-File secrets. Reason: %s in group '%s'"
+const CSPFK052E string = "CSPFK052E Annotation '%s' must contain relative path without leading '/'"
+const CSPFK053E string = "CSPFK053E Annotation '%s' specifies unknown file format '%s'"
+const CSPFK054E string = "CSPFK055E Annotation '%s' contains directory instead of file. Destination file name required when template specified."
```
rpothier, line 64: Should this string be "CSPFK054E Annotation ..."

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+type SecretsGroupInfo struct {
+	Secrets      SecretsPaths
+	FilePath     string
+	FileFormat   SecretsFileFormat
+	FilePerms    os.FileMode
+	FileTemplate string
+}
+
+// SecretsGroups comprises secrets mapping info for all secret groups
+type SecretsGroups map[string]SecretsGroupInfo
```
rpothier, line 51: So here the key of the map is the label? Looks good.

pkg/secrets/pushtofile/secrets_groups_test.go
```diff
@@ -0,0 +1,170 @@
...
+		{input: "invalid", output: SecretsFileFormat(FILE_FORMAT_NONE), valid: false, hasTemplate: false},
+		{input: "dotenv", output: SecretsFileFormat(FILE_FORMAT_NONE), valid: true, hasTemplate: true},
+		{input: "invalid", output: SecretsFileFormat(FILE_FORMAT_NONE), valid: false, hasTemplate: true},
+	}
+
+	secretsGroupTestCase := extractSecretsGroupsTestCase{
+		description: "file type annotation",
+		contents: map[string]string{
+			"conjur.org/conjur-secrets.cache":     someValidSecrets,
+			"conjur.org/secret-file-format.cache": "",
```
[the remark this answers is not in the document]
imheresamir: Refactored the file format tests

pkg/secrets/pushtofile/secrets_groups_test.go
```diff
@@ -0,0 +1,170 @@
...
+			"conjur.org/secret-file-path.cache":   "this/relative/file",
```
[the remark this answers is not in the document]
imheresamir: A file needs to be specified here since some of the file format tests test with a valid template, which requires a file, not a directory (which is the default)

pkg/secrets/pushtofile/secrets_groups_test.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+	}
+
+	resultSecretsGroups := SecretsGroups{
```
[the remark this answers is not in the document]
imheresamir: Refactored the file format tests

pkg/secrets/pushtofile/secrets_groups_test.go
```diff
@@ -0,0 +1,170 @@
...
+		if tc.hasTemplate {
+			secretsGroupTestCase.contents["conjur.org/secret-file-template.cache"] = someValidTemplate
+			group.FileTemplate = someValidTemplate
+		} else {
+			secretsGroupTestCase.contents["conjur.org/secret-file-template.cache"] = ""
+			group.FileTemplate = ""
+		}
+
+		if tc.valid {
+			secretsGroupTestCase.assert = assertExpectedSecretsGroups(resultSecretsGroups)
```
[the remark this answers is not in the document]
imheresamir: Refactored the file format tests

pkg/log/messages/error_messages.go
imheresamir: Fixed, thanks

pkg/secrets/pushtofile/secrets_groups.go
imheresamir: I just followed the spec - and I think it makes more sense for this to be a map.

pkg/secrets/pushtofile/secrets_groups.go
imheresamir: Yep, the name of the secrets group

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// SecretsGroups comprises secrets mapping info for all secret groups
+type SecretsGroups map[string]SecretsGroupInfo
+
+func ExtractSecretsGroupsFromAnnotations(annotations map[string]string) (SecretsGroups, error) {
```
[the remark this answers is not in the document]
imheresamir: I'm open to suggestions! My thinking with the name was: input = annotations map, output = validated `SecretsGroups`

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+		secrets[name] = secretsPathsPrefix + secretPath
+
+		return nil
+	}
+
+	for _, secret := range unpacked {
+		switch val := secret.(type) {
+		case string:
+			{
+				name := path.Base(val)
```
doodlesbykumbi, line 128: This is another case where we use the path package which assumes file paths, and we get some funky behaviour because policy paths are not file paths. Beware edge cases! For example when `val` is an empty string then name becomes `.` which is not what we want.

We could roll our own with:
```
name := val[strings.LastIndex(val, "/")+1:]
```

pkg/secrets/pushtofile/secrets_groups_test.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, tc := range extractSecretsGroupsTestCases {
+		t.Run(tc.description, func(t *testing.T) {
+			secretGroups, err := ExtractSecretsGroupsFromAnnotations(tc.contents)
+
+			tc.assert(t, secretGroups, err)
+		})
+	}
+}
+
+func TestExtractFileFormatFromAnnotations(t *testing.T) {
```
doodlesbykumbi, line 97: I'm not sure the type SecretsFileFormat needs to be public. If not then we can avoid unit testing it. It is indirectly tested as part of the tests for `ExtractSecretsGroupsFromAnnotations`. 

Supposing that we keep `TestExtractFileFormatFromAnnotations`, it's currently being tested by exercising `ExtractSecretsGroupsFromAnnotations`. I think `parseFileFormat` might be the more appropriate unit of logic to test here.

doodlesbykumbi (changes requested):
Thanks for the PR @imheresamir. I've left some feedback and highlighted some concerns around using the path package. There are some suggestions for refactor to aid in readability where some logic has a high line count. I also highlighted a feature which seems to be missing from the PR
> ... The file path can be either a directory (any value ending in /) or a path that includes a file name. For directory paths, a default file name of the form {secret-group}.{secret-file-format} is assumed...

pkg/secrets/pushtofile/secrets_groups_test.go
```diff
@@ -0,0 +1,148 @@
...
+		{input: "PLAINTEXT", output: SecretsFileFormat(FILE_FORMAT_PLAINTEXT), valid: true, hasTemplate: false},
+		{input: "plaintext", output: SecretsFileFormat(FILE_FORMAT_PLAINTEXT), valid: true, hasTemplate: false},
+		{input: "invalid", output: SecretsFileFormat(FILE_FORMAT_NONE), valid: false, hasTemplate: false},
+		{input: "dotenv", output: SecretsFileFormat(FILE_FORMAT_NONE), valid: true, hasTemplate: true},
+		{input: "invalid", output: SecretsFileFormat(FILE_FORMAT_NONE), valid: false, hasTemplate: true},
+	}
+
+	someValidAnnotationsMap := map[string]string{
+		"conjur.org/conjur-secrets.cache":   someValidSecrets,
+		"conjur.org/secret-file-path.cache": "some/file/name",
```
diverdane, line 124: One enhancement to the test cases that you might consider is, rather than setting the `conjur.org/secret-file-path.cache` annotation unconditionally for every test case,  instead add a `hasFileName` boolean flag (or a more general `secretFilePath` field) to the `fileFormatTestCase` struct.

That way, you could test more combinations, e.g. test with `hasTemplate` set to `true`, but with `secretFilePath` set to a directory instead of a full file path.

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+
+				if err := insertSecret(name, val); err != nil {
+					return nil, err
+				}
+			}
+
+		case map[string]interface{}:
+			{
+				for alias, secretPath := range val {
+					secretPath, _ := secretPath.(string)
```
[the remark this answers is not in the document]
doodlesbykumbi: The case `map[string]interface{}` would match for values of any type (not only strings). Right now we just ignore if the type casting fails. This can happen If the user somehow provides values that are not strings. We would ideally want to inform them through a validation error.

pkg/secrets/pushtofile/secret_groups.go
```diff
@@ -0,0 +1,90 @@
...
+	case "json":
+		fallthrough
+	case "dotenv":
+		fallthrough
+	case "bash":
+		fallthrough
+	case "plaintext":
+		return fileFormat, nil
+	case "":
+		return "yaml", nil
```
jonahx, line 86: You can use a comma separate list for a case, that will make this shorter:

https://gobyexample.com/switch

pkg/secrets/pushtofile/secret_spec.go
```diff
@@ -0,0 +1,61 @@
...
+		for k, v := range mapValue {
+			t.Id = v
+			t.Alias = k
+		}
+
+		return nil
+	}
+
+	// ELSE
+	return log.RecordedError(messages.CSPFK051E, "unrecognized format for secret spec")
```
jonahx, line 50: Line of sight will mitigate large chunks of indentation and is consistent with our style guide.

Here that must means return early if there's an error, then handle the happy path under that.

imheresamir: I agree, thanks for the suggestion

pkg/secrets/pushtofile/secret_groups.go
imheresamir: Done

pkg/secrets/push_to_file/secrets_groups.go
```diff
@@ -0,0 +1,235 @@
...
+	secretsGroups := make(SecretsGroups)
+
+	for annotation := range annotations {
+		if strings.HasPrefix(annotation, SECRETS_GROUP_PREFIX) {
+			groupName := strings.TrimPrefix(annotation, SECRETS_GROUP_PREFIX)
+
+			secretsPathPrefix, err := parseSecretsPathPrefix(groupName, annotations[SECRETS_GROUP_POLICY_PATH_PREFIX+groupName])
```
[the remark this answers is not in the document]
imheresamir: Reverted the error checking...

pkg/secrets/push_to_file/secrets_groups.go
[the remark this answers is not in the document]
imheresamir: Went with `New SecretGroupsFromAnnotations`

pkg/secrets/push_to_file/secrets_groups.go
imheresamir: There's no null string just `""`

pkg/secrets/pushtofile/secrets_groups.go
imheresamir: I disagree and think a map is better suited here. 1) Disagree on added cognitive load (for whom and how?) 2) It maps (key = secret name, val = conjur secret path) so `SecretPaths` seems like an appropriate name. I changed it to avoid prolonged pointless bikeshedding

pkg/secrets/pushtofile/secrets_groups.go
imheresamir: Disagree but changed to avoid prolonged pointless bikeshedding

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+	"path"
+	"strings"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	CONJUR_SECRETS_PREFIX             = "conjur.org/conjur-secrets."
```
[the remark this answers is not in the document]
imheresamir: Strongly dislike this convention by Golang. Really confusing how a `conjurSecretsPrefix` looks like either a const or a local variable.

pkg/secrets/pushtofile/secrets_groups.go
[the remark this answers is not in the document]
imheresamir: The error case was handled but I dropped all this code in the interest of time. IMO: custom unmarshaller is unnecessary here - I prefer standard unmarshalling to `interface{}` type + type switch. Custom unmarshaller requires 2 calls to `unmarshal()` in the worst case. This approach seems more straightforward. Again, changed since I don't want to spend more time on a 0 value add discussion

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	if filePath == "." {
+		return DEFAULT_FILE_PATH, nil
+	}
+
+	return DEFAULT_FILE_PATH + filePath, nil
+}
+
+func parseFileFormat(groupName string, fileFormat string) (SecretsFileFormat, error) {
```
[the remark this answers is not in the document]
imheresamir: 1) Why pass around strings? That's the point of enums
2) IMO better to have a function which takes in the enum value as input and returns the file extension as output instead of attaching it to a struct
3) Changed anyways because who cares

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, log.RecordedError(messages.CSPFK050E, "No valid secrets found", groupName)
+	}
+
+	return secrets, nil
+}
+
+func parseConjurSecretsPathPrefix(groupName string, secretsPathPrefix string) (string, error) {
+	// By default returns the root policy path '/'
+	secretsPathPrefix = path.Clean("/" + secretsPathPrefix)
+	if secretsPathPrefix == "." || secretsPathPrefix == "/" {
```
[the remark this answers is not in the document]
imheresamir: The special cases were handled here but I went ahead and removed all of this sanitization and error checking

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			secrets, err := parseConjurSecretsPaths(groupName, annotations[CONJUR_SECRETS_PREFIX+groupName], secretsPathPrefix)
+			if err != nil {
+				return nil, err
+			}
+
+			filePath, err := parseFilePath(groupName, annotations[SECRET_FILE_PATH_PREFIX+groupName])
```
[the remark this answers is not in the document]
imheresamir: I'm no longer parsing this annotation as part of this PR, just dumping it into the object

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+
+func parseConjurSecretsPaths(groupName string, secretsPaths string, secretsPathsPrefix string) (SecretsPaths, error) {
+	secrets := make(SecretsPaths)
+
+	unpacked := []interface{}{}
+	if err := yaml.Unmarshal([]byte(secretsPaths), &unpacked); err != nil {
+		return nil, log.RecordedError(messages.CSPFK050E, err.Error(), groupName)
+	}
+
+	insertSecret := func(name string, secretPath string) error {
```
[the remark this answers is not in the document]
imheresamir: Disagree for reasons explained above. The closure is just to DRY out the code. IMO the function was straightforward and easy to follow. Changed anyway.

pkg/secrets/pushtofile/secrets_groups.go
```diff
@@ -0,0 +1,220 @@
...
+				FilePerms:    DEFAULT_FILE_PERMS,
+				FileTemplate: fileTemplate,
+			}
+
+			err = validateGroupInfo(groupName, groupInfo)
+			if err != nil {
+				return nil, err
+			}
+
+			secretsGroups[groupName] = groupInfo
```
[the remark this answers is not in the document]
imheresamir: Not sure I'm following the vision here but this function seems straightforward to me. It already delegates field parsing and validation to other functions. Changed anyway in later commits.

pkg/secrets/pushtofile/secrets_groups.go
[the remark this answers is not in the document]
doodlesbykumbi: I agree that the implementation of unmarshalling that avoids the 2 calls to `unmarshal` is more straightforward. My suggestion was to move it out into the custom unmarshaller method for the dual benefit of unmarshalling via a "native" API and collocating the unmarshalling logic with the entity that needs to be unmarshalled.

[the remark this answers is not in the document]
imheresamir: I think a custom unmarshaller is entirely unnecessary in this case. There's no need to create an entity just to unmarshal to, especially a struct with 2 fields representing key and value. It should be unmarshalled into a temporary object and then directly put into the appropriate structure, which in this case is a map.

pkg/secrets/pushtofile/secrets_groups.go
imheresamir: I believe we're overthinking this. This is not a complex problem. Just use a map and add a parameter for the key to functions processing the group object

doodlesbykumbi (approved):

PR #363, opened 2021-09-28
retrieve secrets

### LINKED ISSUE #343: Given populated data structure, get secrets and write to YAML file
As a developer,
I need a defined data structure,
So that developers can develop in parallel.

The write secrets file module will need to have a data structure passed in with the
following data to be able to write the file.

- The retrieved conjur variable map
- The parsed template
- Custom Template if provided
- Destination filepath

The secrets can be written to different types of files, yaml, json, bash and dotenv.
The file writing methods will need to 

- Validate the output names for to be valid for the format output
- Parese the template
- Test rendering the template to check for errors before using secrets
- Render the template
- Write out the fike.
 
This could be implemented with an go interface as shown below.

```
type fileWriter interface {
    validateVarNames() error
    parseTemplate( .... ) // Called BEFORE Conjur secrets are retrieved
    testRender() error   // Could be called before Conjur secrets are retrieved
    render(varMap map[string]string) ([]byte, error)
    write(content []byte, filePath string) error
}
```

# Acceptance Criteria

- [ ] The data structures exist for writing the secrets to file.

rpothier:
### What does this PR do?

Parse the secrets group structure and retrieve the secrets from Conjur

Resolves #343 

#### Change log

#### Test coverage

#### Documentation


john-odonnell (commented):
Just a couple small comments. Looks good otherwise!

pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -0,0 +1,51 @@
...
+
+		if err != nil {
+			return log.RecordedError(messages.CSPFK052E, err.Error())
+		}
+
+	}
+	return err
+}
+
+func RetrieveConjurSecretGroup(
```
diverdane, line 42: I think this function doesn't need to be exported?

pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -0,0 +1,51 @@
...
+	accessToken access_token.AccessToken,
+	variableIDs []string) (
+	map[string][]byte, error) {
+	accessTokenData, err := accessToken.Read()
+	if err != nil {
+		return nil, log.RecordedError(messages.CSPFK053E, err.Error())
```
diverdane, line 48: Can we simply reuse error `CSPFK003E` here, rather then creating a new error code?

pkg/secrets/pushtofile/secret_groups.go
```diff
@@ -28,6 +28,7 @@ type SecretGroup struct {
 	SecretSpecs            []SecretSpec
 	FileFormat             string
 	FilePerms              os.FileMode
+	SecretsMap			   map[string][]byte
```
diverdane, line 31: Might be simpler to store the values as strings, rather than slices of bytes (unless I'm missing something?).

Also, the field name `SecretsValues` might be more unambiguous than `SecretsMap`?

diverdane (commented):
Looks good so far! I'm still digging through the UT but wanted to post a partial review.

pkg/secrets/pushtofile/retrieve_secrets.go
rpothier: OK, will do. Maybe 2E fits better.

pkg/secrets/pushtofile/secret_groups.go
rpothier: Changed the name, I will change the []bytes to strings also.

pkg/secrets/pushtofile/retrieve_secrets_test.go
```diff
@@ -0,0 +1,152 @@
...
+		)
+	}
+}
+
+var retrieveSecretsTestCases = []retrieveSecretsTestCase{
+	{
+		description: "Happy Case",
+		contents: SecretGroups{
+			SecretGroup{
+				Label:                  "cache",
```
diverdane, line 35: There is a lot of repetition in the test case definitions that can be eliminated, making the test code more DRY.

For each test case, the fields such as `FilePath`, `FileTemplate`, and `ConjurSecretPathPrefix` don't really matter for the purposes of the test. The test case input that really matters (and is distinct) for each case can be boiled down to the `SecretSpecs` for each group.

Additionally, what we really care about in the resulting `SecretGroups` can be boiled down to the `SecretsValues` maps for each group.

So each test case can be minimally defined by providing:
- Description
- Test case "input": Map of secret group labels to the group's `SecretSpec`s
- Test case expected "output": Map of secret group labels to the groups expected `SecretsValues` map.

This minimizes the amount of "boiler plate" that we have to define for each test case. On the other hand, it does require some helper functions to:
- Create a `SecretGroups` slice based on the SecretSpec for each group
- Compare the resulting SecretsValues to the expected SecretsValues for each group.

Here is some sample code that does this:
```
package pushtofile

import (
	"fmt"
	"reflect"
	"testing"

	"github.com/cyberark/secrets-provider-for-k8s/pkg/secrets/k8s_secrets_storage/mocks"
	"github.com/stretchr/testify/assert"
)

type retrieveSecretsTestCase struct {
	description string
	secretSpecs map[string][]SecretSpec
	assert      func(t *testing.T, result SecretGroups, err error)
}

func createSecretGroups(groupSpecs map[string][]SecretSpec) SecretGroups {
	secretGroups := SecretGroups{}
	for label, secretSpecs := range groupSpecs {
		secretGroup := SecretGroup{
			Label:         label,
			SecretSpecs:   secretSpecs,
			SecretsValues: map[string]string{},
		}
		secretGroups = append(secretGroups, secretGroup)
	}
	return secretGroups
}

func findGroupValues(secretGroups SecretGroups, label string) *map[string]string {
	for _, secretGroup := range secretGroups {
		if secretGroup.Label == label {
			return &secretGroup.SecretsValues
		}
	}
	return nil
}

func valuesAreEqual(values1, values2 map[string]string) bool {
	return reflect.DeepEqual(values1, values2)
}

func assertGoodResults(expectedGroupValues map[string]map[string]string) func(*testing.T, SecretGroups, error) {
	return func(t *testing.T, result SecretGroups, err error) {

		if !assert.NoError(t, err) {
			return
		}
		for groupLabel, expValues := range expectedGroupValues {
			actualValues := findGroupValues(result, groupLabel)
			assert.NotNil(t, actualValues)
			assert.True(t, valuesAreEqual(*actualValues, expValues))
		}
	}
}

var retrieveSecretsTestCases = []retrieveSecretsTestCase{
	{
		description: "Happy Case",
		secretSpecs: map[string][]SecretSpec{
			"cache": []SecretSpec{
				{Alias: "api-url", Id: "dev/openshift/api-url"},
				{Alias: "username", Id: "dev/openshift/username"},
				{Alias: "password", Id: "dev/openshift/password"},
			},
			"db": []SecretSpec{
				{Alias: "api-url", Id: "ci/openshift/api-url"},
				{Alias: "username", Id: "ci/openshift/username"},
				{Alias: "password", Id: "ci/openshift/password"},
			},
		},
		assert: assertGoodResults(map[string]map[string]string{
			"cache": map[string]string{
				"api-url":  string("https://postgres.example.com"),
				"username": string("admin"),
				"password": string("open-$e$ame"),
			},
			"db": map[string]string{
				"api-url":  string("https://ci.postgres.example.com"),
				"username": string("administrator"),
				"password": string("open-$e$ame"),
			},
		}),
	},
	{
		description: "Bad ID",
		secretSpecs: map[string][]SecretSpec{
			"cache": []SecretSpec{
				{Alias: "api-url", Id: "foo/openshift/bar"},
				{Alias: "username", Id: "dev/openshift/username"},
				{Alias: "password", Id: "dev/openshift/password"},
			},
			"db": []SecretSpec{
				{Alias: "api-url", Id: "ci/openshift/api-url"},
				{Alias: "username", Id: "ci/openshift/username"},
				{Alias: "password", Id: "ci/openshift/password"},
			},
		},
		assert: func(t *testing.T, result SecretGroups, err error) {
			assert.Contains(t, err.Error(), "Failed to retrieve secrets")
		},
	},
}

func TestRetrieveSecrets(t *testing.T) {
	var mockAccessToken mocks.MockAccessToken
	prepareMockDBs()
	retrieveConjurSecretsFunc = mocks.RetrieveConjurSecrets
	for _, tc := range retrieveSecretsTestCases {
		t.Run(tc.description, func(t *testing.T) {
			s := createSecretGroups(tc.secretSpecs)
			fmt.Printf("TestRetrieveSecrets %t\n", mocks.CanExecuteConjurVar)
			err := IterateOverGroups(mockAccessToken, &s)
			tc.assert(t, s, err)
		})
	}
}

func prepareMockDBs() {
	mocks.CanExecuteConjurVar = true
	mocks.MockConjurDB = make(map[string][]byte)
	mocks.MockConjurDB["dev/openshift/api-url"] = []byte("https://postgres.example.com")
	mocks.MockConjurDB["dev/openshift/username"] = []byte("admin")
	mocks.MockConjurDB["dev/openshift/password"] = []byte("open-$e$ame")
	mocks.MockConjurDB["ci/openshift/api-url"] = []byte("https://ci.postgres.example.com")
	mocks.MockConjurDB["ci/openshift/username"] = []byte("administrator")
	mocks.MockConjurDB["ci/openshift/password"] = []byte("open-$e$ame")
}
```

diverdane: The boiler plate can probably be minimized further by defining the default input test cases (i.e. the "happy path" inputs), and the default output for test cases (i.e. the "happy path" results). For the negative test case, you would then simply need to specify the incremental change to the default inputs to turn it into a bad case.

pkg/secrets/pushtofile/retrieve_secrets_test.go
```diff
@@ -0,0 +1,152 @@
...
+			s = tc.contents
+			fmt.Printf("TestRetrieveSecrets %t\n", mocks.CanExecuteConjurVar)
+			err := IterateOverGroups(mockAccessToken, &s)
+			tc.assert(t, s, err)
+		})
+	}
+}
+
+func prepareMockDBs() {
+	mocks.CanExecuteConjurVar = true
```
diverdane, line 144: Here's another example where the existing Secrets Provider code uses global variables for test mocking. (I chatted with @doodlesbykumbi  about this earlier today.) This means that tests probably can't be run in parallel, since parallel tests may be writing conflicting values to the global vars.

I don't think we should hold up your merge for this, but we'll probably need to file a ticket to eventually fix this.

pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/access_token"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/secrets/clients/conjur"
+	"strings"
+)
+
+var retrieveConjurSecretsFunc = conjur.RetrieveConjurSecrets
```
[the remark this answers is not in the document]
rpothier: Looks good, I actually had this in my first iteration, then changed it. I will restore it.

pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -0,0 +1,50 @@
+package pushtofile
+
+import (
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/access_token"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/secrets/clients/conjur"
+	"strings"
```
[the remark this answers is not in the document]
rpothier: yes, thanks, updating. Without the space gofmt was putting it at the bottom.

pkg/secrets/pushtofile/retrieve_secrets_test.go
rpothier: Rebased to [Refactor mocks packages](https://github.com/cyberark/secrets-provider-for-k8s/pull/368)

pkg/secrets/pushtofile/retrieve_secrets_test.go
rpothier: updated with the suggestions in the first part of this comment.

pkg/secrets/pushtofile/retrieve_secrets_test.go
```diff
@@ -0,0 +1,122 @@
...
+				{Alias: "username", Id: "dev/openshift/username"},
+				{Alias: "password", Id: "dev/openshift/password"},
+			},
+			"db": []SecretSpec{
+				{Alias: "api-url", Id: "ci/openshift/api-url"},
+				{Alias: "username", Id: "ci/openshift/username"},
+				{Alias: "password", Id: "ci/openshift/password"},
+			},
+		},
+		assert: assertGoodResults(map[string]map[string]string {
```
doodlesbykumbi, line 72: Nice! I like the format of the test cases.  Assertions on input-output pairs for the happy path, followed by deviation from the happy path.

pkg/secrets/pushtofile/retrieve_secrets_test.go
```diff
@@ -0,0 +1,122 @@
...
+    conjurMockClient.AddSecret("ci/openshift/api-url", "https://ci.postgres.example.com")
+    conjurMockClient.AddSecret("ci/openshift/username", "administrator")
+    conjurMockClient.AddSecret("ci/openshift/password", "open-$e$ame")
+
+	for _, tc := range retrieveSecretsTestCases {
+		t.Run(tc.description, func(t *testing.T) {
+			s := createSecretGroups(tc.secretSpecs)
+			err := iterateOverGroups(mockAccessToken, &s, conjurMockClient.RetrieveSecrets)
+			tc.assert(t, s, err)
+		})
```
doodlesbykumbi, line 120: I have recently taken to defining a `Run` method on the test spec so that 
1. The loop is a lot simpler and in its body we just call `tc.Run(t, ...deps)`
2. Each instance of the test case becomes easier to run in isolation because it defines how it needs to be run

pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+		retrieved, err = retrieveConjurSecretGroup(accessToken, requiredSecrets,
+			retrieveConjurSecretsFunc)
+		if err != nil {
+			return log.RecordedError(messages.CSPFK052E, err.Error())
+		}
+		for _, spec := range group.SecretSpecs {
+			for id, secret := range retrieved {
+				if strings.Contains(id, spec.Id) {
+					(*secretGroups)[groupkey].SecretsValues[spec.Alias] = string(secret)
```
doodlesbykumbi, line 44: `secretGroups` should work in place of `(*secretGroups)`

pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func iterateOverGroups(accessToken access_token.AccessToken,
+	secretGroups *SecretGroups,
+	retrieveConjurSecretsFunc conjur.RetrieveConjurSecretsFunc) error {
+
+	var err error
+	var retrieved map[string][]byte
+
+	for groupkey, group := range *secretGroups {
```
doodlesbykumbi, line 31: Some thoughts on
>`retrieveConjurSecretGroup` is being called for each group separately and in sequence.

1. We could make this code concurrent by using `sync.Waitgroup`, see https://gobyexample.com/waitgroups, and running each fetch in a separate Go routine. For capturing results and errors you can use a channel, see this for errors https://stackoverflow.com/a/24715755/5458056.
2. We could also leverage the fact that secrets across all groups can be fetched in one fell swoop, and only have a single call to the `retrieveConjurSecretGroup` function.

PR #364, opened 2021-09-28
Adds UT for FromFile in annotations package

diverdane:
### What does this PR do?

This change adds UT for the FromFile function in the Secrets Provider `annotations` package.

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/pkg/secrets/annotations/annotation_parser.go
+++ b/pkg/secrets/annotations/annotation_parser.go
@@ -11,32 +11,53 @@ import (
 	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
 )
 
-// FromFile reads and parses and annotations file that has been created
-// by Kubernetes via the Downward API, based on Pod annotations that are defined
-// in a deployment manifest.
-func FromFile(path string) (map[string]string, error) {
-	annotationsFile, err := os.OpenFile(path, os.O_RDONLY, os.ModePerm)
+// fileOpener is a function type that captures dependency injection for
+// filesystem operations. It returns an instantiation of the 'io.ReadCloser'
+// interface, which incorporates the only two filesystem operations that we
+// need for parsing an annotations file:
+//   - File closer
+//   - IO reader
+type fileOpener func(name string, flag int, perm os.FileMode) (io.ReadCloser, error)
+
+// osFileOpener is a 'fileOpener' that uses standard OS.
+func osFileOpener(name string, flag int, perm os.FileMode) (io.ReadCloser, error) {
+	return os.OpenFile(name, flag, perm)
+}
+
+// NewAnnotationsFromFile reads and parses an annotations file that has been
+// created by Kubernetes via the Downward API, based on Pod annotations that
+// are defined in a deployment manifest.
+func NewAnnotationsFromFile(path string) (map[string]string, error) {
+	// Use standard OS
+	return newAnnotationsFromFile(osFileOpener, path)
+}
+
+// newAnnotationsFromFile performs the work of NewAnnotationsFromFile(), and
+// provides a function entrypoint that allows filesystem mocking for test
+// purposes.
+func newAnnotationsFromFile(fo fileOpener, path string) (map[string]string, error) {
+	annotationsFile, err := fo(path, os.O_RDONLY, os.ModePerm)
 	if err != nil {
 		return nil, log.RecordedError(messages.CSPFK041E, path, err.Error())
 	}
 	defer annotationsFile.Close()
-	return ParseReader(annotationsFile)
+	return newAnnotationsFromReader(annotationsFile)
 }
 
-// ParseReader parses an input stream representing an annotations file that
-// had been created by Kubernetes via the Downward API, returning a string-to-string
-// map of annotations key-value pairs.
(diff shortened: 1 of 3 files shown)


pkg/secrets/annotations/annotation_parser.go
```diff
@@ -15,7 +16,14 @@ import (
...
 // in a deployment manifest.
 func FromFile(path string) (map[string]string, error) {
-	annotationsFile, err := os.OpenFile(path, os.O_RDONLY, os.ModePerm)
+	// Use the standard OS
+	return fromFile(vfs.OS(), path)
+}
+
+// fromFile performs the work of FromFile(), and provides a function
+// entrypoint that allows filesystem mocking for test purposes.
+func fromFile(fs vfs.Filesystem, path string) (map[string]string, error) {
```
doodlesbykumbi, line 25: I like the general approach, and happy to approve but I wanted to run by you a suggestion in the spirit of avoiding adding a dependency. The suggestion works because the function that actually does the parsing, `ParseReader`, just needs a reader. Files just need closing, so the lowest common denominator in terms of interfaces is the `io.ReadCloser`.

```
package main

import (
	"fmt"
	"io"
	"io/ioutil"
	"os"
	"strings"
)

// fileOpener is the interface that captures the dependency injection, much like vfs.Filesystem in your example
type fileOpener func(name string, flag int, perm os.FileMode) (io.ReadCloser, error)

// some utilities for mocking
func mockReadCloser(contents string) io.ReadCloser {
	return ioutil.NopCloser(strings.NewReader(contents))
}
func mockFileOpenerGenerator(store map[string]io.ReadCloser) fileOpener {
	return func(name string, flag int, perm os.FileMode) (io.ReadCloser, error) {
		rc, ok := store[name]
		if ok {
			return rc, nil
		}
		return nil, fmt.Errorf("file not found")
	}
}

func main() {
	// mock fileOpener
	fo := mockFileOpenerGenerator(map[string]io.ReadCloser{
		"file": mockReadCloser("file contents"),
	})

	// real fileOpener
	// fo := func(name string, flag int, perm os.FileMode) (io.ReadCloser, error) {
	//	return os.OpenFile(name, flag, perm)
	// }

	rc, err := fo("file", os.O_RDONLY, os.ModePerm)
	if err != nil {
		panic(err)
	}
	defer rc.Close()

	out, err := ioutil.ReadAll(rc)
	fmt.Println(string(out), err)
}
```

pkg/secrets/annotations/annotation_parser.go
```diff
@@ -15,7 +16,14 @@ import (
 // by Kubernetes via the Downward API, based on Pod annotations that are defined
 // in a deployment manifest.
 func FromFile(path string) (map[string]string, error) {
-	annotationsFile, err := os.OpenFile(path, os.O_RDONLY, os.ModePerm)
+	// Use the standard OS
```
doodlesbykumbi, line 19: @diverdane would you mind making these changes as part of this PR ? To maintain the convention in Go of explicitly calling out what the function is constructing e.g. `strings.NewReader`. 

FromFile -> NewAnnotationsFromFile
ParseReader -> NewAnnotationsFromReader

doodlesbykumbi (commented):
Some suggestions, otherwise happy to merge and follow up

pkg/secrets/annotations/annotation_parser.go
diverdane: Yeah, I think it's worth the effort to avoid a new dependency.

pkg/secrets/annotations/annotation_parser.go
diverdane: Fixed.

diverdane: @doodlesbykumbi I actually changed this up a little bit.
Instead of using `NewAnnotationsFromFile`, I used `NewFromFile`. The reason I used this shortened form for the function name is that when it gets invoked from outside of this package, it will get invoked as:
```
    annotations.NewFromFile( . . . )
```
Whereas using the `NewAnnotationsFromFile` would stutter:
```
    annotations.NewAnnotationsFromFile( . . . )
```

I also used `newFromReader` instead of `NewAnnotationsFromReader` to be consistent with the `NewFromFile()`. I also made this a local (unexported) function, since it's only used internally.

doodlesbykumbi (changes requested):
Thanks for addressing those earlier comments @diverdane. We just need to change all newFrom to newAnnotationsFrom. I think the convention of being explicit about the thing new is constructing is important since  it tells you all you need to know about what you’re making from the name, I think it’s an anti pattern to give that responsibility to the package name since a package could house multiple entities and could be aliased.

diverdane:
> We just need to change all newFrom to newAnnotationsFrom.

I'll switch this back. I had intentionally switched to the `newFrom...` (instead of `newAnnotations From...`) to avoid stuttering, e.g. where we invoke `annotations.NewAnnotationsFromFile( . . . )` from outside of this package. I just checked whether this would be flagged as stuttering with `revive` (a replacement for `golint`), and it's not flagged, so I'm good with sticking with the Go convention.

doodlesbykumbi (approved):
LGTM! Thanks for addressing feedback

PR #365, opened 2021-09-29
Disable Cobertura autoUpdateStability

diverdane:
### What does this PR do?

This change disables the `autoUpdateStability` and authUpdateHealth features in Cobertura runs in Jenkins. These features apparently use moving average statistics (not sure of the algorithm used since documentation for these Cobertura plugin features is scarce to nonexistent) over a sliding window of past code coverage results. There's also a "ratchet" mechanism that keeps "raising the bar" for acceptible overall coverage (e.g. starts at 80%, but then can move to 90-95% after several runs).

This means that if you've had a history of good coverage, the coverage threshold can be ratcheted up to 92%. So if you add new code that falls below this stringent threshold, Cobertura runs will continue to fail. And because of the sliding window of past code coverage results, the moving average keeps going down with each run, yet the "ratcheted"
threshold is never lowered.

At some point, when we have more Golang unit tests implemented for Secrets Provider, the `autoUpdateStability` feature may be less sensitive to new code that's being added (i.e when the number of lines of code being introduced is small compared to number of lines already being tested). When we get to that point, it may be more appropriate to use the
`autoUpdateStability` feature, and we can consider re-enabling this feature.

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -98,7 +98,7 @@ pipeline {
             sh './bin/test_unit'
 
             junit 'junit.xml'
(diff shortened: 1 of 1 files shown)


imheresamir (approved):
TY

PR #369, opened 2021-10-05
Run Conjur DAP E2E CI first before Conjur OSS E2E

diverdane:
### What does this PR do?

This change incorporates some enhancements for the E2E tests:

1. The order for running E2E tests on GKE for Conjur OSS vs. running
   parallel E2E tests for Conjur Dap on GKE and OpenShift.

   The issue is that the E2E tests running on OpenShift ("current" version)
   is flakey, so it would be better to run this test (and parallel tests
   that are run concurrently with that test) before the Conjur OSS E2E tests,
   so that flakey failures happen faster, and we can therefor restart
   failed tests failure.

2. The TEST_ID_3_SECRETS_DESTINATION_with_incorrect_value.sh E2E
   test case is eliminated. This is already tested in UT here:
   https://github.com/cyberark/secrets-provider-for-k8s/blob/main/pkg/secrets/config/config_test.go#L286-L293

3. The TEST_ID_7_K8S_SECRETS_env_var_not_exist.sh E2E test case is
   eliminated. This test case is already coverd in UT here:
   https://github.com/cyberark/secrets-provider-for-k8s/blob/main/pkg/secrets/config/config_test.go#L279-L285

4. The TEST_ID_8_K8S_SECRETS_env_var_empty.sh E2E test case is
   eliminated. A new UT test case is being added to cover this test
   case.

5. The TEST_ID_21_helm_service_account_does_not_exist.sh E2E test case is
   eliminated. This test case is currently failing. This test is really
   only testing Kubernetes functionality: It's testing the case where:

   - The SP Helm chart is being used
   - The user does not enable creation of a ServiceAccount (or other
     RBAC objects)
   - The user does not provide the name of an existing ServiceAccount
     in the Namespace where SP is being deployed.

   In this scenario, the SP Job is OF COURSE going to fail to start
   because the Job manifest specifies this ServiceAccount is to be
   used by the SP Job, on this line:
   https://github.com/cyberark/secrets-provider-for-k8s/blob/main/helm/secrets-provider/templates/secrets-provider.yaml#L26
   This is a function of how Kubernetes works; it's not testing
   Secrets Provider functionality directly.

#### Change log

#### Test coverage

#### Documentation


doodlesbykumbi (approved):
Works for me. I think a small additional win would be to make the OpenShift stuff run nightly. It's the main offender when it comes to flakiness

doodlesbykumbi (commented):
@diverdane Can we restrict OpenShift E2E to only run on main. I think it's a nice compromise for now.
```
when {
  branch 'main'
}
```

rpothier (approved):
LGTM! Thanks for fixing this.

diverdane:
> @diverdane Can we restrict OpenShift E2E to only run on main. I think it's a nice compromise for now.
> 
> ```
> when {
>   branch 'main'
> }
> ```

@doodlesbykumbi , I'll make this change separately. For now, the OpenShift runs on `current` are completely broken because of `TEST_ID_21_helm_service_account_does_not_exist.sh` (this test is very fragile: It's looking for a specific error string from Kubernetes). As part of this change, I've deleted this test because the test is merely testing that Kubernetes fails in some expected way... it's not really testing Secrets Provider functionality.

I've also deleted several other E2E test cases that are already covered in UT, and added a simple UT test case so that we can delete yet another E2E test case.

PR #383, opened 2021-11-01
Review unit testing for templated secret file output

john-odonnell:
### Desired Outcome

Review the design doc and confirm that we have sufficient test coverage for the template file output feature.

From the design doc:
| Given | Then |
|-|-|
| "secret-file-template" annotation referencing secrets by name (alias) | The internal representation matches expectation |
| "secret-file-template" annotation iterative over all secrets | The internal representation matches expectation |
| "secret-file-template" annotation using built-in template function (e.g. `html`) | The internal representation matches expectation |
| "secret-file-template" annotation using custom functions with nested templates | The internal representation matches expectations |

### Implemented Changes

The following were already tested:
- template referencing secrets by name (alias)

Adds unit testing for the following cases:
- iterative custom template
- use of built-in template funcs (html)
- nested templates

Includes implementing a "double-pass" method for template rendering.
Now, a malformed template will cause SP to fail fast, before fetching secrets.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-13144]()

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/pkg/secrets/pushtofile/secret_group.go
+++ b/pkg/secrets/pushtofile/secret_group.go
@@ -89,7 +89,7 @@ func (sg *SecretGroup) pushToFileWithDeps(
 	defer func() {
 		_ = wc.Close()
 		if r := recover(); r != nil {
-			err = fmt.Errorf("panic recovered while pushing to writer")
+			err = fmt.Errorf("panic recovered while executing template with secret values")
 		}
 	}()
 
--- a/pkg/secrets/pushtofile/secret_group.go
+++ b/pkg/secrets/pushtofile/secret_group.go
@@ -316,11 +316,7 @@ func newSecretGroup(groupName string, secretsBasePath string, annotations map[st
 
 	secretSpecs, err := NewSecretSpecs([]byte(groupSecrets))
 	if err != nil {
-		err = fmt.Errorf(
-			`cannot create secret specs from annotation "%s": %s`,
-			secretGroupPrefix+groupName,
-			err,
-		)
+		err = fmt.Errorf(`cannot create secret specs from annotation "%s": %s`, secretGroupPrefix+groupName, err)
 		return nil, []error{err}
 	}
 
--- a/pkg/secrets/pushtofile/secret_group.go
+++ b/pkg/secrets/pushtofile/secret_group.go
@@ -2,6 +2,7 @@ package pushtofile
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 	"sort"
@@ -87,12 +88,29 @@ func (sg *SecretGroup) pushToFileWithDeps(
 		_ = wc.Close()
 	}()
 
-	return depPushToWriter(
+	// Calls to pushToWriter include a second-pass at parsing and executing the provided Go template
+	// First-pass parsing and rendering is validated with dummy secret values in newSecretGroup
+	// Output from this step should be redirected to /dev/null to mitigate secret leaks in log output
+	// This may result in obscured details of an execution failure here, which would require further experimentation in a local environment
+	stdout := os.Stdout
+	stderr := os.Stderr
+	defer func() {
+		os.Stdout = stdout
+		os.Stderr = stderr
+	}()
+	os.Stdout = os.NewFile(0, os.DevNull)
+	os.Stderr = os.NewFile(0, os.DevNull)
+
+	err = depPushToWriter(
 		wc,
 		sg.Name,
 		fileTemplate,
 		secrets,
 	)
+	if err != nil {
+		return fmt.Errorf("template for secret group %q failed to render with provided secret values", sg.Name)
+	}
+	return nil
 }
 
 func (sg *SecretGroup) absoluteFilePath(secretsBasePath string) (string, error) {
@@ -156,6 +174,7 @@ func (sg *SecretGroup) absoluteFilePath(secretsBasePath string) (string, error) 
 func (sg *SecretGroup) validate() []error {
 	groupName := sg.Name
 	fileFormat := sg.FileFormat
+	fileTemplate := sg.FileTemplate
 	secretSpecs := sg.SecretSpecs
 
 	if errors := validateSecretPaths(secretSpecs, groupName); len(errors) > 0 {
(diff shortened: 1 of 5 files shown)


pkg/secrets/pushtofile/secret_group.go
```diff
@@ -217,6 +235,25 @@ func newSecretGroup(annotations map[string]string, groupName string) (*SecretGro
...
+	if len(fileTemplate) > 0 {
+		dummySecrets := []*Secret{}
+		for _, secretSpec := range secretSpecs {
+			dummySecrets = append(dummySecrets, &Secret{Alias: secretSpec.Alias, Value: "REDACTED"})
+		}
+
+		err = pushToWriter(ioutil.Discard, groupName, fileTemplate, dummySecrets)
+		if err != nil {
+			return nil, []error{fmt.Errorf(
+				`file template for secret group %q cannot be used as written: %s`,
```
doodlesbykumbi, line 250: I think let's change
```
file template for secret group %q cannot be used as written: %s
```
to
```
unable to use file template for secret group %q: %s
```

to
1. Front-load the verb
2. I think the `as written` might be redundant

pkg/secrets/pushtofile/secret_group.go
```diff
@@ -85,12 +86,29 @@ func (s *SecretGroup) pushToFileWithDeps(
...
+	// Output from this step should be redirected to /dev/null to mitigate secret leaks in log output
+	// This may result in obscured details of an execution failure here, which would require further experimentation in a local environment
+	stdout := os.Stdout
+	stderr := os.Stderr
+	defer func() {
+		os.Stdout = stdout
+		os.Stderr = stderr
+	}()
+	os.Stdout = os.NewFile(0, os.DevNull)
+	os.Stderr = os.NewFile(0, os.DevNull)
```
doodlesbykumbi, line 100: Some thoughts...

Redirecting to `/dev/null` in-process will be in effect while this function runs to completion. Code executing in other Go routines at the same time will be affected. I don't know if there's way to limit the effect to the function body.

---

I think there are other approaches to preventing secret leakage that don't need to rely on redirecting to `/dev/null`...

Here are the ways in which secrets could be leaked 
1.`depPushToWriter` panics, and prints to stdout/stderr
2. `depPushToWriter` returns an error, which is then printed to stdout/stderr
3. `depPushToWriter` uses the log module and leaks to stdout/stderr

(1) and (2) and be dealt with in a similar fashion. (1) requires recovering from the panic first. The approach is to hide the actual error and instead return a generic error that states that write to file failed due to issues executing the template.

For (3) we can statically find all the places in the code where `log.RecordedErr` etc. are invoked, and we can ensure the payload doesn't include a potential leak.

doodlesbykumbi (commented):
Thanks for working on this. I think the general approach to double-pass template execution works. I left some comments around the /dev/null redirection.

pkg/secrets/pushtofile/secret_group.go
john-odonnell: 1. Added panic recovery before call to `depPushToWriter`
2. Errors returned by `depPushToWriter` are received, and then exchanged for a hardcoded message
3. The standard `depPushToWriter`, `pushToWriter`, does not use `log.RecordedError`

pkg/secrets/pushtofile/secret_group.go
```diff
@@ -58,11 +59,12 @@ func (sg *SecretGroup) PushToFile(secrets []*Secret) error {
 func (sg *SecretGroup) pushToFileWithDeps(
 	depOpenWriteCloser openWriteCloserFunc,
 	depPushToWriter pushToWriterFunc,
-	secrets []*Secret) error {
+	secrets []*Secret,
+) (err error) {
```
doodlesbykumbi, line 63: I'm curious about the motivation behind switched to a named return argument ?
I think we tend to avoid them because they can be a bit challenging to follow

pkg/secrets/pushtofile/secret_group.go
```diff
@@ -75,24 +77,33 @@ func (sg *SecretGroup) pushToFileWithDeps(
...
-	return depPushToWriter(
+	err = depPushToWriter(
 		wc,
 		sg.Name,
 		fileTemplate,
 		secrets,
 	)
+	if err != nil {
+		err = fmt.Errorf("failed to render template for secret group %q with provided secret values", sg.Name)
+		return
```
doodlesbykumbi, line 104: This return can be removed, the code would still work as intended and code climate would be happy.

pkg/secrets/pushtofile/secret_group.go
```diff
@@ -75,24 +77,33 @@ func (sg *SecretGroup) pushToFileWithDeps(
...
 	//// Open and push to file
 	wc, err := depOpenWriteCloser(sg.FilePath, sg.FilePermissions)
 	if err != nil {
-		return err
+		return
 	}
+
 	defer func() {
 		_ = wc.Close()
+		if r := recover(); r != nil {
```
doodlesbykumbi, line 91: I think it might aid readability to have a separate defer for the panic recovery for push to write, just before calling push to writer.

pkg/secrets/pushtofile/secret_group.go
```diff
@@ -75,24 +77,33 @@ func (sg *SecretGroup) pushToFileWithDeps(
...
+			err = fmt.Errorf("panic recovered while executing template with secret values")
```
doodlesbykumbi, line 92: More context in the message e.g.
```
failed to execute template, with secret values, on push to file for secret group %q
```

pkg/secrets/pushtofile/secret_group.go
```diff
@@ -285,11 +316,7 @@ func newSecretGroup(groupName string, secretsBasePath string, annotations map[st
 
 	secretSpecs, err := NewSecretSpecs([]byte(groupSecrets))
 	if err != nil {
-		err = fmt.Errorf(
-			`cannot create secret specs from annotation "%s": %s`,
-			secretGroupPrefix+groupName,
-			err,
-		)
+		err = fmt.Errorf(`cannot create secret specs from annotation "%s": %s`, secretGroupPrefix+groupName, err)
```
doodlesbykumbi, line 319: I know this isn't related to this PR specifically but while we're here, I think we should be consistent with the `unable to create...` vs `cannot create`.

pkg/secrets/pushtofile/secret_group_test.go
```diff
@@ -89,6 +97,24 @@ func goodSecretSpecs() []SecretSpec {
...
+
+func panicPushToWriter(
+	writer io.Writer,
+	groupName string,
+	groupTemplate string,
+	groupSecrets []*Secret,
+) error {
+	panic("canned panic response - maybe containing secrets")
+}
+
```
doodlesbykumbi, line 117: Since each of these functions are being used only once I think it's fine to inline them... to avoid a step of indirection.

doodlesbykumbi (commented):
PR looking good. Left some comments

pkg/secrets/pushtofile/secret_group.go
john-odonnell: That's fair, they are kind of hard to follow.

I was having trouble getting the deferred function to override the returned `err` variable upon panic in `depPushToWriter`. I don't think it's capable of assigning variables that are declared in any other way ([playground example](https://play.golang.org/p/-MVgib3WVQx)).

doodlesbykumbi: Thanks, TIL! Interesting behaviour, not at all the mental model I had

pkg/secrets/pushtofile/secret_group.go
john-odonnell: I used this example as a general template execution failure message, whether recovering from panic or returning an error.

doodlesbykumbi (approved):
LGTM! Thanks for addressing the comments.

PR #404, opened 2021-12-08
List of paths to retrieve from Conjur should be a set

diverdane:
### Desired Outcome

In Push-to-File mode, the Secrets Provider should not fail (crash) when either:

- Duplicated secret paths: A secret path is used in more than one secret group
- Empty secret path list: When there are no secrets to actually retrieve from Conjur

Currently, Secrets Provider in Push-to-File mode is crashing for both of these cases,
since Conjur is apparently intolerant of either duplicate secret paths being retrieved
or an empty list of secret paths to retrieve.

### Implemented Changes

- When secret paths are gathered across all secret groups, the paths are compiled
  into a **set** of secret paths, not simply an aggregated slice.
- Secrets retrieval from Conjur is skipped if there are no secrets to retrieve.

### Definition of Done

- [x] Changes made to ensure list of secrets to retrieve from Conjur is guaranteed to contain unique entries
- [x] Changes are made to skip secret retrieval if there are no secrets to retrieve

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/pkg/secrets/clients/conjur/conjur_secrets_retriever.go
+++ b/pkg/secrets/clients/conjur/conjur_secrets_retriever.go
@@ -60,6 +60,11 @@ func (retriever conjurSecretRetriever) Retrieve(variableIDs []string) (map[strin
 func retrieveConjurSecrets(accessToken []byte, variableIDs []string) (map[string][]byte, error) {
 	log.Info(messages.CSPFK003I, variableIDs)
 
+	if len(variableIDs) == 0 {
+		log.Info(messages.CSPFK016I)
+		return nil, nil
+	}
+
 	conjurClient, err := NewConjurClient(accessToken)
 	if err != nil {
 		return nil, log.RecordedError(messages.CSPFK033E)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   if it finds any pre-existing secret files. This is helpful when the Secrets
   Provider is being run multiple times.
   [cyberark/secrets-provider-for-k8s#397](https://github.com/cyberark/secrets-provider-for-k8s/pull/397)
+- If the Secrets Provider is run in Push-to-File mode, it no longer errors out
+  if either (a) multiple secret groups use the same secret path, or (b) there
+  are no secrets that need to be retrieved.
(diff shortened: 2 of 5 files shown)


szh (approved):
Good catch! 🐞 
LGTM 👍

pkg/secrets/clients/conjur/conjur_secrets_retriever.go
```diff
@@ -60,6 +60,10 @@ func (retriever conjurSecretRetriever) Retrieve(variableIDs []string) (map[strin
 func retrieveConjurSecrets(accessToken []byte, variableIDs []string) (map[string][]byte, error) {
 	log.Info(messages.CSPFK003I, variableIDs)
 
+	if len(variableIDs) == 0 {
+		return nil, nil
+	}
```
john-odonnell, line 65: I might be missing some context, but wouldn't Secrets Provider not retrieving any secrets from Conjur be a condition where failure is expected and preferable?

If there is some context where this is an acceptable outcome, should we at least provide an Info or Warn level log when Secrets Provider isn't retrieving any secrets? I think the log messaging would be a bit unclear.

Example: if no error is returned at this point, then eventually the SP logs will display a success message like `Secrets [pushed to file/updated in K8s] successfully` when actually no secrets were pushed/updated.

diverdane: I've updated the code to produce an INFO level message when this occurs. I'm open for debate on whether this should be an error, but my preference is to leave this as info for future corner cases.

diverdane: Thinking about this some more, if we did want things to error out for this sort of misconfiguration, I think a better place for it would be to check within each group, maybe even making it conditional on secret file format for each group. Said another way, even if we did "error out" here when no Conjur secrets are retrieved for all groups, there could still be a case where one particular secret group has not specified any secrets to retrieve from Conjur, but other secret groups have.

john-odonnell: > I think the best bet is to add a warning here, but otherwise allow for zero secrets ... for "future proofing"

I agree! Thanks for context around reasonable no-secret scenarios.

> I think a better place for [errors re. zero secrets] would be to check within each group

Definitely a better idea to approach the no-secret scenarios on a per-group basis. That said, I'm all for the INFO level log ATM.

PR #426, opened 2022-01-18
Add Secrets Provider refresh interval annotations and timer

rpothier:
### Desired Outcome

Add annotations and a Go timer to periodically refresh secrets.

golang-gopher-wall-clock

### Implemented Changes

Two new annotations have been added for provider refresh.
`conjur.org/secrets-refresh-enabled`
`conjur.org/secrets-refresh-interval`
also, `side-car` has been added to  `conjur.org/container-mode`

Adds a go timer routine to call periodicProvider.
This is configured by adding a wrapper function to provideSecrets, if Secrets Provider
is configured in side-car mode a new goRouting is invoked that waits for a 
timer tick that was configured with secrets-refresh-interval, and will update
the secrets on each tick.

Also had to optimize ValidateSecretsProviderSettings due to CodeClimate complaining
there were too many lines in the function.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-15538](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-15538)
[ONYX-16504](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-16504)

### Definition of Done
- [x] Add support for new Annotations `conjur.org/secrets-refresh-interval` and `conjur.org/secrets-refresh-enabled`.
- [x] In Secrets Provider, call the `provideSecrets()` function periodically based on the new configuration.
- [x] Add UT for the parsing of the new Annotation.

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


pkg/secrets/provide_conjur_secrets_test.go
```diff
@@ -41,6 +51,119 @@ func eventualProvider(successAfterNtries int) func() error {
...
+		err := periodicProvider()
+
+		if err == nil && providerCount != tc.expectedCount {
+			err = fmt.Errorf("incorrect number of timer ticks, got %d expected %d",
+				providerCount,tc.expectedCount)
+		}
+
+		logMessages := logBuffer.String()
+		if tc.assertOn == "fail" {
+			assert.NotNil(t/**/, err)
```
szh, line 159: `/**/` seems to be a typo

cmd/secrets-provider/main.go
```diff
@@ -130,11 +130,12 @@ func secretRetriever(ctx context.Context,
 		log.Error(messages.CSPFK008E)
 		return nil, err
 	}
-	if err = validateContainerMode(authnConfig.GetContainerMode()); err != nil {
+	// Do we need to check this? Container mode is checked in ValidateAnnotations
```
szh, line 133: Does `ValidateAnnotations` catch if the container mode annotation is missing? Or is that OK since it'll just use the default?

szh (commented):
Looking great! Left a couple of questions

cmd/secrets-provider/main.go
rpothier: I have removed validateContainerMode as `ValidateAnnotations` checks this.

PUSH_TO_FILE.md
```diff
@@ -323,12 +323,14 @@ for a description of each environment variable setting:
...
 | `conjur.org/k8s-secrets`            | `K8S_SECRETS`         | This list is ignored when `conjur.org/secrets-destination` annotation is set to **`file`** |
 | `conjur.org/retry-count-limit`      | `RETRY_COUNT_LIMIT`   | Defaults to 5
 | `conjur.org/retry-interval-sec`     | `RETRY_INTERVAL_SEC`  | Defaults to 1 (sec)              |
 | `conjur.org/debug-logging`          | `DEBUG`               | Defaults to `false`              |
+| `conjur.org/secrets-refresh-enabled`| Note\* | Defaults to `false`. Secrets Provider will exit with error if this is explicitly set to `false` and `conjur.org/secrets-rotation-interval` is explicitly set. |
+| `conjur.org/secrets-refresh-interval`           | Note\* | Set to a valid duration string as defined [here](https://pkg.go.dev/time#ParseDuration). Valid time units are `s`, `m`, and `h` (for seconds, minutes, and hours, respectively). Some examples of valid duration strings:<ul><li>`5m`</li><li>`2h30m`</li><li>`48h`</li></ul>The minimum refresh interval is 1 second. A refresh interval of 0 seconds is treated as a fatal configuration error. The maximum refresh interval is approximately 290 years. |
```
szh, line 333:
```suggestion
| `conjur.org/secrets-refresh-interval`           | Note\* | Set to a valid duration string as defined [here](https://pkg.go.dev/time#ParseDuration). Valid time units are `s`, `m`, and `h` (for seconds, minutes, and hours, respectively). Some examples of valid duration strings:<ul><li>`5m`</li><li>`2h30m`</li><li>`48h`</li></ul>The minimum refresh interval is 1 second. A refresh interval of 0 seconds is treated as a fatal configuration error. The maximum refresh interval is 100 years. |
```

to match https://github.com/cyberark/secrets-provider-for-k8s/pull/426/files#diff-1a2f43d8162d071687f1a5864ce5b6b825e9f25d5a74a1df1c218bbbfe597a33R25

szh (commented):
Looks great! Just one comment

PUSH_TO_FILE.md
rpothier: Hmm, would it be better to leave this and change the default timeout?

szh: I don't think it makes much of a difference, 100 or 290.

PUSH_TO_FILE.md
```diff
@@ -323,12 +323,14 @@ for a description of each environment variable setting:
 | K8s Annotation  | Equivalent<br>Environment Variable | Description, Notes |
 |-----------------------------------------|---------------------|----------------------------------|
 | `conjur.org/authn-identity`         | `CONJUR_AUTHN_LOGIN`  | Required value. Example: `host/conjur/authn-k8s/cluster/apps/inventory-api` |
-| `conjur.org/container-mode`         | `CONTAINER_MODE`      | Allowed values: <ul><li>`init`</li><li>`application`</li></ul>Defaults to `init`.<br>Must be set (or default) to `init` for Push to File mode.|
+| `conjur.org/container-mode`         | `CONTAINER_MODE`      | Allowed values: <ul><li>`init`</li><li>`application`</li><li>`side-car`</li></ul>Defaults to `init`.<br>Must be set (or default) to `init` or `side-car`for Push to File mode.|
```
diverdane, line 326: Typo, `side-car` should be single word `sidecar`. Same throughout the rest of the changes.
Also, we should consider making the `sidecar` option the default, but probably should discuss with Alex.

rpothier: thanks for catching that, updated.

cmd/secrets-provider/main.go
```diff
@@ -186,6 +181,14 @@ func retryableSecretsProvider(
 		secretsConfig.RetryCountLimit,
 		provideSecrets,
 	)
+
+	provideSecrets = secrets.PeriodicSecretProvider(
+		secretsConfig.SecretsRefreshInterval,
+		getContainerMode(),
+		provideSecrets,
+		// Set a large timeout, basically forever
+		secretsConfigProvider.MaxRefreshTimeout,
```
diverdane, line 190: Is this MaxRefreshTimeout used for implementing a close proximity for "sleep forever" functionality? If so, I think we can do this differently, and have the SP block forever.

pkg/secrets/provide_conjur_secrets.go
```diff
@@ -97,3 +96,51 @@ func RetryableSecretProvider(
...
+		case mode == "sidecar" && secretRefreshInterval > 0:
+			ticker := time.NewTicker(secretRefreshInterval)
+			go periodicSecretProvider(provideSecrets, ticker)
+		default:
+			// fall through
+		}
+		// Wait here
+		select {
+		case <-ProviderDone: // Todo change to SIGTERM or SIGINTERRUPT
+		case <-time.After(timeout):
```
diverdane, line 122: Can we just delete this line?:
```
    case <-time.After(timeout):
```
the previous case `case <-ProviderDone:` should be enough to wait forever, right?

PUSH_TO_FILE.md
```diff
@@ -323,12 +323,14 @@ for a description of each environment variable setting:
...
+| `conjur.org/container-mode`         | `CONTAINER_MODE`      | Allowed values: <ul><li>`init`</li><li>`application`</li><li>`sidecar`</li></ul>Defaults to `init`.<br>Must be set (or default) to `init` or `sidecar`for Push to File mode.|
 | `conjur.org/secrets-destination`    | `SECRETS_DESTINATION` | Allowed values: <ul><li>`file`</li><li>`k8s_secrets`</li></ul> |
 | `conjur.org/k8s-secrets`            | `K8S_SECRETS`         | This list is ignored when `conjur.org/secrets-destination` annotation is set to **`file`** |
 | `conjur.org/retry-count-limit`      | `RETRY_COUNT_LIMIT`   | Defaults to 5
 | `conjur.org/retry-interval-sec`     | `RETRY_INTERVAL_SEC`  | Defaults to 1 (sec)              |
 | `conjur.org/debug-logging`          | `DEBUG`               | Defaults to `false`              |
+| `conjur.org/secrets-refresh-enabled`| Note\* | Defaults to `false`. Secrets Provider will exit with error if this is explicitly set to `false` and `conjur.org/secrets-rotation-interval` is explicitly set. |
```
diverdane, line 332: I prefer what you have here: having the `secrets-refresh-enabled` setting simply default to `false`, rather then defaulting to `true` or `false` depending upon whether `secrets-refresh-enabled` is explicitly set.  The latter way is more convenient in some ways, but it's MUCH more difficult to document. I think we need to get an okay from Alex on this, since I think his expectation is the latter.

PUSH_TO_FILE.md
```diff
@@ -323,12 +323,14 @@ for a description of each environment variable setting:
...
+| `conjur.org/secrets-refresh-interval`           | Note\* | Set to a valid duration string as defined [here](https://pkg.go.dev/time#ParseDuration). Valid time units are `s`, `m`, and `h` (for seconds, minutes, and hours, respectively). Some examples of valid duration strings:<ul><li>`5m`</li><li>`2h30m`</li><li>`48h`</li></ul>The minimum refresh interval is 1 second. A refresh interval of 0 seconds is treated as a fatal configuration error. The maximum refresh interval is approximately 290 years. |
```
diverdane, line 333: Before "The minimum refresh interval is 1 second", it should also say "The default refresh interval is 5 minutes."

PUSH_TO_FILE.md
rpothier: Based on other offline conversation, I'm removing the timeout so this is no longer an issue.

pkg/secrets/provide_conjur_secrets.go
rpothier: deleted.

PUSH_TO_FILE.md
rpothier: The implementation does not require a `true` if the ` secrets-refresh-interval` set. So this needs to be updated still.

pkg/secrets/config/config.go
```diff
@@ -46,17 +52,38 @@ type annotationRestraints struct {
 	allowedValues []string
 }
 
+// Annotations used to support Conjur secrets rotation.
```
diverdane, line 55: This comment needs to be updated. The Annotations listed aren't just for secrets rotation, but for all SP non-push-to-file Annotations.

pkg/secrets/config/config.go
```diff
@@ -138,51 +165,53 @@ func ValidateSecretsProviderSettings(envAndAnnots map[string]string) ([]error, [
 	}
 
 	envStoreType := envAndAnnots["SECRETS_DESTINATION"]
-	annotStoreType := envAndAnnots["conjur.org/secrets-destination"]
+	annotStoreType := envAndAnnots[SecretsDestinationKey]
 	storeType := ""
+	var err error
 
 	if annotStoreType == "" {
```
diverdane, line 172: Your call on this, but since you're rewriting some of this already.... Golang coding style recommends using `switch` block rather than `if`/`else if`/`else` block. So this section would have the form:
```
    switch {
        case annotStoreType == "":
            // Do things
        case validStoreType(annotStoreType):
            // Do other things
        default:
            // Log an error
    }
```

cmd/secrets-provider/main.go
```diff
@@ -87,10 +87,13 @@ func main() {
 	}
 
 	// Provide secrets
-	err = provideSecrets()
-	if err != nil {
-		errStr := fmt.Sprintf(messages.CSPFK039E, secretsConfig.StoreType, err.Error())
-		logError(errStr)
+	go provideSecrets()
+	select {
+	case err = <-secrets.SecretProviderError:
```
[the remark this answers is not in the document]
diverdane: Looking at this again, I think the logic of calling a goroutine and providing an error channel should be moved to the `pkg/secrets` package. Generally, we want the `main.go` to be as simple as possible, and I think that this implementation detail for periodic secrets provider function should be abstracted away from `main.go`.

We would still need to do exactly what you're doing here inside the test code, since that will need to have a goroutine that can either run once and return, run once and sleep, or run periodically... and be able to stop that goroutine from the test thread.

To see what I'm suggesting, take a look at these diffs that I created on a branch based on your PR branch:
https://github.com/cyberark/secrets-provider-for-k8s/commit/d61c96c091f171be0aa04be5c75a6407a3e5f092

cmd/secrets-provider/main.go
```diff
@@ -186,6 +184,12 @@ func retryableSecretsProvider(
 		secretsConfig.RetryCountLimit,
 		provideSecrets,
 	)
+
+	provideSecrets = secrets.PeriodicSecretProvider(
```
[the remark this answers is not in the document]
diverdane: Or maybe just rename it to something very generic e.g. "SecretProvider()"? Not sure what makes sense here.

pkg/secrets/provide_conjur_secrets.go
```diff
@@ -97,3 +96,55 @@ func RetryableSecretProvider(
...
+		default:
+			// fall through
+		}
+		// Wait here
+		select {
+		case <-ProviderDone: // Todo change to SIGTERM or SIGINTERRUPT
+		}
+		if mode == "sidecar" && secretRefreshInterval > 0{
+			TickerQuit <- true
+			time.Sleep(1 * time.Second) // give a second for the go routine to shut down
```
diverdane, line 130: Might be better to define this as a const at the top of the file so it's easy to find/change.

pkg/secrets/provide_conjur_secrets.go
```diff
@@ -97,3 +96,55 @@ func RetryableSecretProvider(
 		return err
 	}
 }
+
+var TickerQuit = make(chan bool)
```
diverdane, line 100: These three channels don't need to be defined/declared as global variables.

To make this testable, the "ProviderDone" channel should passed in from the caller to PeriodicSecretProvider (in main.go, we'll pass it a channel that's not used for now, but in the future we can add a signal handler that can use this channel to stop the SP).

See diffs here:
https://github.com/cyberark/secrets-provider-for-k8s/commit/d61c96c091f171be0aa04be5c75a6407a3e5f092#diff-c083eb56e1c331264db79fbb96d5fbb5ecfd8f082e9dde16e1a0e28f5f0f4230L100-R119

pkg/secrets/provide_conjur_secrets.go
```diff
@@ -97,3 +96,55 @@ func RetryableSecretProvider(
...
+		}
+		if mode == "sidecar" && secretRefreshInterval > 0{
+			TickerQuit <- true
+			time.Sleep(1 * time.Second) // give a second for the go routine to shut down
+		}
+		return nil
+	}
+}
+
+func periodicSecretProvider(provideSecrets ProviderFunc, ticker *time.Ticker) {
```
diverdane, line 136: You can pass in the error channel (for output) and the quit channel (for input) to this function, so that you won't have to use global variables.

See: https://github.com/cyberark/secrets-provider-for-k8s/commit/d61c96c091f171be0aa04be5c75a6407a3e5f092#diff-c083eb56e1c331264db79fbb96d5fbb5ecfd8f082e9dde16e1a0e28f5f0f4230L136-R164

pkg/secrets/provide_conjur_secrets_test.go
```diff
@@ -14,11 +14,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	providerDelayTime = 500
+)
+var providerCount = 0
```
diverdane, line 20: I think the reason that we're seeing some flaky test cases below (the ones that are commented out) is that we're using one common `providerCount` gobal variable for all test cases. It makes testing more complex when you're testing slow provider functionality because you have to wait for some grace period at the end of each test case for the previous test case's slow provider to finish.

Generally, we want each test case to have its own state. So this should be implemented as a structure that has a `provide()` method, and instantiate one of those for each test case.

To see what I mean, check out this mocking:
https://github.com/cyberark/secrets-provider-for-k8s/commit/d61c96c091f171be0aa04be5c75a6407a3e5f092#diff-c0314624500148cb0ccac2409ef0d4dc9a8cddf7603e8879bae326cc2d96d7d5R17-R52

This mocking works, except that I haven't reconciled it with the `eventualProvider()` function, which depends upon those base objects.

pkg/secrets/provide_conjur_secrets_test.go
```diff
@@ -14,11 +14,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	providerDelayTime = 500
```
diverdane, line 18: I think we can get by with 10s of milliseconds instead of 100s.
For example, this seems to work:
https://github.com/cyberark/secrets-provider-for-k8s/commit/d61c96c091f171be0aa04be5c75a6407a3e5f092#diff-c0314624500148cb0ccac2409ef0d4dc9a8cddf7603e8879bae326cc2d96d7d5L132-R176

PUSH_TO_FILE.md
rpothier: Added.

cmd/secrets-provider/main.go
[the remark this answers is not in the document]
rpothier: Patched in the diffs.

pkg/secrets/config/config.go
rpothier: done

pkg/secrets/provide_conjur_secrets.go
rpothier: made suggested changes.

pkg/secrets/provide_conjur_secrets_test.go
rpothier: made the change.

cmd/secrets-provider/main.go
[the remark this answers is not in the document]
rpothier: yea, as it's not always periodic, we could go with SecretProvider()

szh (approved):
🎉

PR #428, opened 2022-01-19
Add solution design for secrets rotation

diverdane:
### Desired Outcome

A solution design is available as a markdown document in the `design` folder.

### Implemented Changes

Solution doc added to `design` folder.

### Connected Issue/Story

Resolves Jira #15257

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] Docs (e.g. `README`s) were updated in this PR

#### Behavior

#### Security


design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+
+- Start of annotation parsing
+- Completion of annotation parsing
+- Start of content rendering for each Kubernetes Secret / secrets file
+- Start of each Secrets update / write of secrets file
+- Completion of each Secrets update / write of secrets file
+
+For error conditions, the Secrets Provider will include ERROR-level logs.
+Annotation values are treated as non-sensitive and will not be redacted
+in log messages.
```
andytinkham, line 819: Should check that if natefinch/atomic fails in writing out the file, that it doesn't echo what would have been the file contents in its error message.  Also related, but not logs, it looks like atomic tried to remove the temp file it uses if an error occurs, which is good, but simply ignores any error in removing that temp file. Do we need to have a final check to make sure we didn't leave a temp file with secrets values somewhere the k8s admin doesn't expect?

design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+| Implement atomic write of secret files | | | |
+| Implement detection of changes in K8s Secret content with UT | | 3 | |
+| Implement E2E tests for SP Sidecar (with/without rotation) | | 3 | |
+| Implement process restart with UT | | 3 | | 
+| Measure incremental Push-to-file latency using Secrets Provider tracing | | 1 | |
+| Measure incremental `k8s-secret` mode latency using Secrets Provider tracing | | 1 | |
+| Work with tech writers to create official documentation for secrets rotation | | 2 | |
+| Katacoda/Instruqt Demo ??? | | 3 | |
+| Document process restart in GitHub README | | 2 | |
+| Work with tech writers to create official documentation for process restart | | 1 | |
```
andytinkham, line 902: If we're only doing a community-level security review before, we'll need a deeper dive here (plus can cover all the additional implement tasks).

design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+<tr class="odd">
+<td>System Architect</td>
+<td>Rafi Schwarz</td>
+<td><ul>
+<p> </p>
+</ul></td>
+</tr>
+<tr class="even">
+<td>Security Architect</td>
+<td>Andy Tinkham</td>
```
andytinkham, line 938: Approved.

andytinkham (approved):
I added some comments but nothing blocking, I don't think. I'm on board with this plan.

design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+
+    **User Story:** US-03
+    
+  - **Use of SHA-256 checksums to Detect Changes:**<br />
+    For added security, the Secrets Provider will not cache any unencrypted,
+    retrieved secret values in order to detect when Kubernetes Secrets or
+    Push-to-File secret files need to be updated. Instead, the Secrets Provider
+    will store a SHA-256 checksum of the last rendered content that was written
+    to each target Kubernetes Secret or secret file for detecting when contents
+    have changed.
```
adamouamani, line 202: What happens if the file is deleted by the application?  Will SP restore the file @ next sync?  Will SP compare against the cached SHA-256 checksum so file will not be re-created if checksum matches and will only be re-created if checksum does not match?

design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+
+#### How to Configure Secret Rotation
+
+The Secrets Provider Secret Rotation feature is configured on an application
+as follows:
+
+1. Start by configuring the application Deployment as described in the
+   [Secrets Provider documentation](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/11.2/en/Content/Integrations/Kubernetes_deployApplicationsConjur-k8s-Secrets.htm?tocpath=Integrations%7COpenShift%252C%20Kubernetes%252C%20and%20GKE%7CDeploy%20Applications%7C_____3),
+
+1. If you would like to use the Secrets Provider as a sidecar container
```
adamouamani, line 252: If the SP is configured as a Sidecar container and the container fails/crashes - do we need readiness/liveness probes to restart container/pod?

design/secret_rotation_design.md
szh: > it looks like atomic tried to remove the temp file it uses if an error occurs, which is good, but simply ignores any error in removing that temp file. Do we need to have a final check to make sure we didn't leave a temp file with secrets values somewhere the k8s admin doesn't expect?

We can, or we can fork the project and change the behavior so it doesn't ignore the error. What would we do if we found that a temp file couldn't be deleted? Just print a message to the log?

andytinkham: I think we'd need to log it at least. I don't know if we'd have escalation paths built in (sort of like, "if kill doesn't work, try kill -9") to see if we could remove it on subsequent tries, but those might be possibilities to consider as well. I'm guessing not since it's atomic not handling it. Not sure if there's a time component to this either, but a simple retry maybe? Again probably not an issue most of the time and in the rare cases it might be, not clear there's much we can do. We should definitely dig in and make sure we do anything we can not to leave secret values hanging around in a temp file somewhere though.

design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+
+When the Secrets Provider is running as a sidecar container in `file`
+destination mode, it renders and writes secret files asynchronously with
+respect to the application container reading those files. In order to
+guarantee that the application container reads a consistent instance of
+file content, the Secrets Provider needs to write each secret file as an
+atomic operation.
+
+This atomicity can be achieved by using the following third-party package:
+https://github.com/natefinch/atomic.
```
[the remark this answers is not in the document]
szh: My understanding is that this shouldn't pose a problem. Before replacing the file, atomic copies the permissions from the existing destination file ([code here](https://github.com/natefinch/atomic/blob/master/atomic.go#L63-L67)). We can still create that file with the permissions we want, and then pass the path to the atomic writer. Only the temp file will have 0600 permissions.
I imagine we don't need to be concerned that the newly created file will be deleted before we have a chance to write to it (which would cause the destination file to be recreated with 0600 permissions). Is that a safe assumption?

design/secret_rotation_design.md
szh: Here's a draft that implements what I understand you to be saying.
PR https://github.com/cyberark/secrets-provider-for-k8s/pull/440, and specifically commit [9e0f689](https://github.com/cyberark/secrets-provider-for-k8s/pull/440/commits/9e0f6894a7fa30afaa914acb844c9aa1c1cf78dc) (this uses a forked version of the 'atomic' library to report whether the temp file was deleted)

design/secret_rotation_design.md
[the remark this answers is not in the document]
andytinkham: There's nothing we do that would lead to the file ending up with 0600 permissions, right?  If the file did get deleted before it's written to, the only result would be the file would be less accessible than if we'd created it with world readable permissions. Probably safe - I was wondering more if it would end up inaccessible to the end application than if a security hole would be present.

design/secret_rotation_design.md
andytinkham: I think that looks good. Thank you! Different security hat - with a forked version, do we have mechanisms in place to keep our fork uptodate as the base package evolves?

design/secret_rotation_design.md
```diff
@@ -0,0 +1,951 @@
...
+- `conjur.org/secrets-refresh-enabled`
+- `conjur.org/secrets-refresh-interval`
+
+The validation and defaulting for this configuration is implemented in the
+`ValidateSecretsProviderSettings()` in `pkg/secrets/config/config.go` as
+follows:
+
+- If `secrets-refresh-interval` is explicitly set:
+  - If the value cannot be parsed by time.ParseDuration, exit with error
+  - If `secrets-refresh-interval` is set to zero value (`0s`, `0m`, etc.),
```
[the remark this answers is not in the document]
andytinkham: Good. Thanks for the example code!

design/secret_rotation_design.md
diverdane: Added.

rpothier (approved):
LGTM!

PR #429, opened 2022-01-24
ONYX-16318: Add annotation for JWT TOKEN PATH

tzheleznyak:
Added annotation for JWT token path

Ticket : ONYX-16318

### DIFF
--- a/cmd/secrets-provider/main.go
+++ b/cmd/secrets-provider/main.go
@@ -42,6 +42,7 @@ var envAnnotationsConversion = map[string]string{
 	"DEBUG":                "conjur.org/debug-logging",
 	"JAEGER_COLLECTOR_URL": "conjur.org/jaeger-collector-url",
 	"LOG_TRACES":           "conjur.org/log-traces",
+	"JWT_TOKEN_PATH":       "conjur.org/jwt-token-path",
 }
 
 func main() {


cmd/secrets-provider/main.go
```diff
@@ -42,6 +42,7 @@ var envAnnotationsConversion = map[string]string{
 	"DEBUG":                "conjur.org/debug-logging",
 	"JAEGER_COLLECTOR_URL": "conjur.org/jaeger-collector-url",
 	"LOG_TRACES":           "conjur.org/log-traces",
+	"JWT_TOKEN_PATH":       "conjur.org/jwt-token-path",
```
sashaCher, line 45:
```suggestion
	"JWT_TOKEN_PATH":       "conjur.org/jwt-token-path"
```

tzheleznyak: In GOLANG in dicts you should put comma in the end of the final line too

diverdane, line 45: Since the Secrets Provider is the only container that needs to be aware of this path to the JWT token, would it be possible to simply hard-code this path in Secrets Provider code, and not require the user to set this Annotation?

That's the approach we've been taking with the short-lived authn-k8s token path. We've been trying to minimize the number of "configuration knobs" that a user needs to learn. It will also help simplify things when we add sidecar injector support for Secrets Provider (i.e. the Sidecar Injector has one less configurable field in the container that it's injecting).

Then again, maybe you're considering using this in unit testing, where you'll set this path to a `tmp` directory? For this, I think there's probably other ways to inject this path without Annotations, but I could be wrong.

sashaCher: TIL

tzheleznyak: Im not sure i understand it. Its not mandatory here but optional. Its just optinal support for "jwt-token-path" in annotation. 

Lets have a short call and talk about it today

diverdane: Sure, let's chat!

PR #431, opened 2022-01-30
ONYX-16576: Add Authn-JWT support to helm

tzheleznyak:
Changed helm chart to support token projection for JWT cases using this values.yaml 

```
environment:
  k8sSecrets:
  conjur: 
    account:
    applianceUrl:
    authnUrl:
    sslCertificate: 
      value:
    authnJWT:
      projectToken:
      projectedFilename:
      audience:
      expiration:
```

### DIFF
--- a/helm/secrets-provider/templates/secrets-provider.yaml
+++ b/helm/secrets-provider/templates/secrets-provider.yaml
@@ -27,8 +27,17 @@ spec:
       containers:
       - image: {{ .Values.secretsProvider.image }}:{{ .Values.secretsProvider.tag }}
         imagePullPolicy: {{ .Values.secretsProvider.imagePullPolicy }}
+        {{- if .Values.environment.conjur.authnJWT.projectToken}}
+        volumeMounts:
+          - name: jwt-token
+            mountPath: /var/run/secrets/tokens
+        {{- end }}
         name: {{ .Values.secretsProvider.name }}
         env:
+        {{- if .Values.environment.conjur.authnJWT.projectToken}}
+        - name: JWT_TOKEN_PATH
+          value: /var/run/secrets/tokens/{{ .Values.environment.conjur.authnJWT.projectedFilename }}
+        {{- end }}
         - name: MY_POD_NAME
           valueFrom:
             fieldRef:
@@ -91,5 +100,15 @@ spec:
         - configMapRef:
             name: {{ .Values.environment.conjur.conjurConnConfigMap }}
         {{- end }}
+      {{- if .Values.environment.conjur.authnJWT.projectToken}}
+      volumes:
+        - name: jwt-token
+          projected:
+            sources:
+              - serviceAccountToken:
+                  path: {{ .Values.environment.conjur.authnJWT.projectedFilename }}
+                  expirationSeconds: {{ .Values.environment.conjur.authnJWT.expiration }}
+                  audience: {{ .Values.environment.conjur.authnJWT.audience }}
+      {{- end }}
       restartPolicy: Never
   backoffLimit: 0
--- a/helm/secrets-provider/tests/test-schema
+++ b/helm/secrets-provider/tests/test-schema
@@ -138,6 +138,30 @@ function k8s_secrets_test() {
         --set "environment.k8sSecrets=$1"
 }
 
+function authn_jwt_test() {
+    helm lint . --strict \
+        --set "$DEFAULT_ACCOUNT_SETTING" \
+        --set "$DEFAULT_APPLIANCE_URL_SETTING" \
+        --set "$DEFAULT_AUTHN_URL_SETTING" \
+        --set "$DEFAULT_SSL_CERT_SETTING" \
+        --set "$DEFAULT_K8S_SECRETS_SETTING" \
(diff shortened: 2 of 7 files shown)


sashaCher (approved):
LGTM

helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -27,8 +27,17 @@ spec:
...
         imagePullPolicy: {{ .Values.secretsProvider.imagePullPolicy }}
+        {{- if .Values.environment.conjur.authnJWT.projectToken}}
+        volumeMounts:
+          - name: jwt-token
+            mountPath: /var/run/secrets/tokens
+        {{- end }}
         name: {{ .Values.secretsProvider.name }}
         env:
+        {{- if .Values.environment.conjur.authnJWT.projectToken}}
+        - name: JWT_TOKEN_PATH
```
diverdane, line 38: Using an environment variable here for the Secrets Provider "standalone" or "app" mode Helm chart is okay, but when the Secrets Provider is run in any other mode, this should be configured as a Pod Annotation. Is there a plan to add support to Secrets Provider so that this can be configured as a Pod Annotation?

helm/secrets-provider/tests/test-schema
```diff
@@ -216,10 +216,6 @@ function main() {
     conjur_authn_login_test ""
     update_results "$?" "$EXPECT_FAILURE"
 
-    announce "Missing Conjur authn login causes failure"
```
diverdane, line 219: For non-JWT functionality, the value `.Values.environment.conjur.authnLogin` is still required, and we want the user notified (as a "fail fast" mechanism) if this is missing, and it's not configured for authn-jwt.

We'll probably need to add an authnJWT.enabled boolean chart value to this Helm chart, and default it to "false". Then, either in the Helm chart `secrets-provider.yaml` template, or better yet, in the `values.schema.json`, we need to indicate that either `authnJWT.enabled` or `authnLogin` needs to be provided.

With the above changes, then the `missing_conjur_authn_login_test` should still be called here.

helm/secrets-provider/values.yaml
```diff
@@ -84,3 +84,8 @@ environment:
     # This setting is required.
```
diverdane, line 84: If you add an `authnJWT.enabled` setting, then this comment should say:
```
# This setting is required if `authnJWT.enabled` is set to false.
```
Or something like that.

helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -27,8 +27,17 @@ spec:
...
+          value: /var/run/secrets/tokens/{{ .Values.environment.conjur.authnJWT.projectedFilname }}
```
diverdane, line 39: Typo here and below: `projectedFilname` should be `projectedFilename`.

helm/secrets-provider/templates/secrets-provider.yaml
diverdane: Actually, I thought we wouldn't need this setting, and the token would be hard-coded in the Secrets Provider code?

diverdane (dismissed):
We should keep the check for missing `authnLogin` for the non-JWT case, if possible.
Also, I thought we would hard-code the token path in Secrets Provider code?

helm/secrets-provider/templates/secrets-provider.yaml
tzheleznyak: It will be hardcoded  to defaults in documentation and helm charts but if client for any reason want to change he can

tzheleznyak: We talked about not adding it as annotation not?

helm/secrets-provider/tests/test-schema
tzheleznyak: I thought about `authnJWT.enabled` but because token projection is mandatory for authn-jwt then
`authnJWT.projectToken`.

tzheleznyak: How can i do this test thing with values.schema.json?

diverdane:
@tzheleznyak , 

Please take a look at the changes that I made on this branch: https://github.com/cyberark/secrets-provider-for-k8s/commit/20cd2c6cf611662442a52a663fd4c8fc42310aee

On this branch, I've implemented the following:

- Helm chart schema to make `authnLogin` a required Helm chart value if `authn-jwt` authentication is not being used.
- Added back the schema test for missing `authnLogin`.
- Helm chart schema for the chart values that you're adding. This will cause Helm install to fail fast if users supply invalid values for `authn-jwt`.
- Helm chart `unittest` unit tests for the changes that you're adding to the `secrets-provider.yaml` template.
- Helm chart schema tests for the chart values that you're adding.
- Fixed the typo for `Filename`.

(I figured that it's probably quicker to just add the schema/unittest code, since I've done it before.)

I've set the minimum JWT token interval to 1 (second). I don't know if this is appropriate.
I haven't set a maximum on the JWT token interval... if there's an appropriate value, we can add that to the schema.

Please feel free to copy the changes from that branch/commit into your PR branch.

PR #432, opened 2022-02-02
Secrets Rotation: only write secret files if changed

szh:
### Overview:

For conjur secrets rotation functionality for push-to-file mode, the SP's pushtofile package needs to be able to detect changes in secret file content, and conditionally write secret files only if content has been changed since the last write of the file.
### Notes:
 * For security reasons, the previous content of the secret files should not be cached directly. Instead, we should keep a SHA-256 checksum of the file content to compare against.
 * The SHA-256 checksum should not be updated until the secret file has been successfully written (i.e. updated).
 * The processing of the pushToWriter() function can be modified to be:
   1. Parse the template for a secret group
   1. Render the secret file content to a bytes.Buffer (using tpl.Execute())
   1. Caclulate the SHA-256 checksum for the bytes.Buffer data
   1. If checksum differs from prefiously cached checksum, write the file
   1. If writing the file is successful, update the cached checksum with the new checksum

### Implemented Changes

- Modified push to file logic to store the hash of each group's rendered content
- Before writing to the files, checks if the new content matches the previous hash and only writes if it's different

### Connected Issue/Story

CyberArk internal issue link: [ONYX-15541](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-15541)

### Definition of Done

- [x] P2F does not perform a file write if the content of the group hasn't changed
- [x] P2F performs a file write if the rendered content changes, whether due to a secret or template change

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [insert issue ID]()

#### Behavior

#### Security


pkg/log/messages/info_messages.go
```diff
@@ -29,3 +29,4 @@ const CSPFK014I string = "CSPFK014I Authenticator setting %s provided by %s"
 const CSPFK015I string = "CSPFK015I DAP/Conjur Secrets pushed to shared volume successfully"
 const CSPFK016I string = "CSPFK016I There are no secrets to be retrieved from Conjur"
 const CSPFK017I string = "CSPFK017I Creating default file name for secret group '%s'"
+const CSPFK018I string = "CSPFK018I No change in secret files, no secret files written"
```
diverdane, line 32: I believe that this should be singular "secret file" (for both instances), i.e.
`CSPFK018I No change in secret file, no secret file written`

pkg/secrets/pushtofile/push_to_writer.go
```diff
@@ -85,8 +97,62 @@ func pushToWriter(
...
+
+	// Calculate a sha256 checksum on the content
+	checksum, _ := fileChecksum(fileContent)
+
+	// If file contents have changed, write the file and update checksum
+	if contentHasChanged(groupName, checksum) {
+		if _, err := writer.Write(fileContent.Bytes()); err != nil {
+			return err
+		}
+		prevFileChecksums[groupName] = checksum
```
diverdane, line 133: Do we need to log an INFO level message here that an individual file has been written?

Or do we assume that the user will deduce that the secret file has been written if we don't see the `No change in secret files, no secret files written` for this file, but we do see the `DAP/Conjur Secrets pushed to shared volume successfully` at the very end of the refresh cycle?

diverdane (commented):
Looks great, very nice job on the UT!!!

pkg/secrets/pushtofile/push_to_writer.go
szh: I don't think it's necessary to add a new log statement for this functionality. Nothing has changed in that scenario and the "Secrets pushed to shared volume" log at the end should be sufficient. I think it's more important to log if they _aren't_ written since this could be a source of problems if, say, the secrets file was deleted.

diverdane: Makes sense. The only weird aspect is that the logs will continually show "Secrets pushed to shared volume" when they really aren't (after the first write). But then again, we don't want to make the logs too chatty either.

diverdane (approved):
LGTM!

PR #433, opened 2022-02-04
ONYX-16318: Add annotation for JWT TOKEN PATH

tzheleznyak:
Added annotation for JWT token path

Ticket : ONYX-16318


szh:
This will still allow this to be configured via environment variable as well, correct?

szh (changes requested):
I believe we need to add this [here](https://github.com/cyberark/secrets-provider-for-k8s/blob/483275ff8837fef8ef21e112c98ce5c3a10985ae/pkg/secrets/config/config.go#L50) as well.

tzheleznyak:
> This will still allow this to be configured via environment variable as well, correct?

yes

PR #435, opened 2022-02-08
Upgrade to latest go modules

andytinkham:
Signed-off-by: Andy Tinkham <andy.tinkham@cyberark.com>

### Implemented Changes

Update packages in go.mod in an attempt to reduce the number of old versions of golang.org/x/net included.  Unfortunately, it seems to have not reduced them at all, but it still updates package versions.


szh (approved):
LGTM 👍

PR #440, opened 2022-02-16
Write secret files atomically

szh:
### Desired Outcome

This functionality is described here:

https://github.com/cyberark/secrets-provider-for-k8s/blob/secrets-rotation-design/design/secret_rotation_design.md#atomic-file-writes-for-push-to-file-mode

### Implemented Changes

Files are written to a temp file first. If that succeeds, the temp file is moved to the destination in an atomic operation.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-16710](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-16710)

### Definition of Done

- [x] Investigate if it's possible to test atomicity
- [x] Use atomic operation to write secrets files
- [x] Address security questions raised in #428 ([this](https://github.com/cyberark/secrets-provider-for-k8s/pull/428#discussion_r794972112) and [this](https://github.com/cyberark/secrets-provider-for-k8s/pull/428#discussion_r794975481) comment)

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/pkg/atomicwriter/atomic_writer.go
+++ b/pkg/atomicwriter/atomic_writer.go
@@ -1,0 +1,114 @@
+package atomicwriter
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/cyberark/secrets-provider-for-k8s/pkg/log/messages"
+)
+
+// OS Function table
+type osFuncs struct {
+	chmod    func(string, os.FileMode) error
+	rename   func(string, string) error
+	remove   func(string) error
+	truncate func(string, int64) error
+}
+
+// Instantiation of OS Function table using std OS
+var stdOSFuncs = osFuncs{
+	chmod:    os.Chmod,
+	rename:   os.Rename,
+	remove:   os.Remove,
+	truncate: os.Truncate,
+}
+
+type atomicWriter struct {
+	path        string
+	permissions os.FileMode
+	tempFile    *os.File
+	os          osFuncs
+}
+
+// This package provides a simple atomic file writer which implements the
+// io.WriteCloser interface. This allows us to use AtomicWriter the way we
+// would use any other Writer, such as a Buffer. Additonally, this struct
+// takes the file path during construction, so the code which calls
+// `Write()` doesn't need to be concerned with the destination, just like
+// any other writer.
+func NewAtomicWriter(path string, permissions os.FileMode) (io.WriteCloser, error) {
+	return newAtomicWriter(path, permissions, stdOSFuncs)
+}
+
+func newAtomicWriter(path string, permissions os.FileMode, osFuncs osFuncs) (io.WriteCloser, error) {
+	dir, file := filepath.Split(path)
+
+	f, err := ioutil.TempFile(dir, file)
+	if err != nil {
+		log.Error(messages.CSPFK055E, path)
+		return nil, err
+	}
+
+	return &atomicWriter{
+		path:        path,
+		tempFile:    f,
+		permissions: permissions,
+		os:          osFuncs,
+	}, nil
+}
+
+func (w *atomicWriter) Write(content []byte) (n int, err error) {
+	// Write to the temporary file
+	return w.tempFile.Write(content)
+}
+
+func (w *atomicWriter) Close() error {
+	defer w.Cleanup()
+
+	// Flush and close the temporary file
+	err := w.tempFile.Sync()
+	if err != nil {
+		log.Error(messages.CSPFK056E, w.tempFile.Name())
+		return err
+	}
+	w.tempFile.Close()
+
+	// Set the file permissions
+	err = w.os.chmod(w.tempFile.Name(), w.permissions)
+	if err != nil {
+		log.Error(messages.CSPFK057E, w.tempFile.Name())
+		// Try to rename the file anyway
+	}
+
+	// Rename the temporary file to the destination
+	err = w.os.rename(w.tempFile.Name(), w.path)
+	if err != nil {
+		log.Error(messages.CSPFK058E, w.tempFile.Name(), w.path)
+		return err
+	}
+
+	return nil
+}
+
+// Cleanup attempts to remove the temporary file. This function is called by
+// the `Close()` method, but can also be called manually in cases where `Close()`
+// is not called.
+func (w *atomicWriter) Cleanup() {
+	err := w.os.remove(w.tempFile.Name())
+	if err == nil {
+		return
+	}
+
+	// If we can't remove the temporary directory, truncate the file to remove all secret content
+	err = w.os.truncate(w.tempFile.Name(), 0)
+	if err == nil || os.IsNotExist(err) {
+		log.Error(messages.CSPFK059E, w.tempFile.Name(), w.path)
+		return
+	}
+
+	// If that failed as well, log the error
+	log.Error(messages.CSPFK060E, w.tempFile.Name(), w.path)
+}
--- a/pkg/atomicwriter/atomic_writer_test.go
+++ b/pkg/atomicwriter/atomic_writer_test.go
@@ -1,0 +1,255 @@
+package atomicwriter
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/stretchr/testify/assert"
+)
+
+type assertFunc func(path string, tempFilePath string, t *testing.T, err error)
+type errorAssertFunc func(buf *bytes.Buffer, wc io.WriteCloser, t *testing.T, err error)
+
+func TestWriteFile(t *testing.T) {
+	testCases := []struct {
+		name        string
+		path        string
+		permissions os.FileMode
+		content     string
+		assert      assertFunc
+	}{
+		{
+			name:        "happy path",
+			path:        "test_file.txt",
+			permissions: 0644,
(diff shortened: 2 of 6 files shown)


NOTICES.txt
```diff
@@ -34,6 +34,7 @@ Section 3: MIT
 >>> github.com/stretchr/testify - v1.7.0
 >>> gopkg.in/yaml.v3 - v3.0.0-20210107192922-496545a6307b
 >>> sigs.k8s.io/yaml - v1.3.0
+>>> github.com/natefinch/atomic - v1.0.1
```
doodlesbykumbi, line 37: I think you previously mentioned the idea of creating a fork of this package (though I forget the reason). I'm wondering if it might be worth just adding an atomic file writer to a package in this repo. The implementation is on the small side. The atomic write implementations I've seen elsewhere write to a temp file then use `os.Rename` to finish the job.

szh: The reason for the fork was because the original package doesn't report if the temp file couldn't be deleted. Here's what I'm looking to add: https://github.com/natefinch/atomic/pull/17
It is indeed a small feature to implement, just a couple of go files. The only complicated part of it is the [Windows](https://github.com/natefinch/atomic/blob/master/file_windows.go) implementation, but I'm not sure we even need that.

pkg/atomicwriter/atomic_writer.go
```diff
@@ -0,0 +1,55 @@
...
+	// Use the 'atomic' package to write the file atomically.
+	// This writes the content to a temporary file, and then moves the
+	// temporary file to the final destination in an atomic operation.
+	res := atomic.WriteFile(w.path, reader)
+
+	if !res.IsTempFileDeleted {
+		// The temporary file could not be deleted by the 'atomic' package.
+		// Try to delete the temporary file again asyncronously.
+		go func() {
+			time.Sleep(time.Second * 5)
```
tzheleznyak, line 40: Move 5 to const

pkg/atomicwriter/atomic_writer.go
```diff
@@ -0,0 +1,108 @@
...
+// takes the file path during construction, so the code which calls
+// `Write()` doesn't need to be concerned with the destination, just like
+// any other writer.
+type AtomicWriter struct {
+	path        string
+	permissions os.FileMode
+	tempFile    *os.File
+}
+
+func NewAtomicWriter(path string, permissions os.FileMode) (*AtomicWriter, error) {
```
doodlesbykumbi, line 28: For the sake of limiting what is made public to that which is necessary I think we should make the struct definition private `AtomicWriter => atomicWriter` and `NewAtomicWriter` can return an interface`(io.WriteCloser, error)`.

pkg/atomicwriter/atomic_writer.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (w *AtomicWriter) Write(content []byte) (n int, err error) {
+	// Write to the temporary file
+	return w.tempFile.Write(content)
+}
+
+func (w *AtomicWriter) Close() error {
+	defer func() {
+		go w.Cleanup()
+	}()
```
doodlesbykumbi, line 55: 1. Why does this need retry logic ?
2. Deferring a call to a go routine feels like trouble

szh: 1. See [this thread](https://github.com/cyberark/secrets-provider-for-k8s/pull/428#discussion_r794975481)
2. Why? Is there a preferred way to do this? My assumption is that a regular defer() function will block execution, which is a problem if we're doing timed retries.

pkg/atomicwriter/atomic_writer.go
szh: Changed in 3e1def1ede9eacd3d5d24d268036dd2770aaddaa

diverdane (commented):
@szh I've added some comments re. one way to mock the `os.Remove` function for testing to your query on the `about-golang` Teams channel (along with the `os.Chmod` and `os.Rename` functions).

pkg/atomicwriter/atomic_writer.go
doodlesbykumbi: 1. I'm not sure about the appropriateness of retry here. I think this is an instance of an application assuming the reliability responsibilities of something like the OS/file system. I think we ought to be able to safely assume reliability here. I understand that deletion might fail, but I think a single attempt + errors should be enough. An alternative might be to try and delete and if it doesn't work Truncate the file to 0, which is highly likely to work if done after successful writes. 
2. It's sort of a code smell, at least in my eyes. Defer creating Go routines that are not tracked by anything... An alternative, if we really decide to keep an eye on those tmp file and make sure they are cleaned up might be to a have some sort of temp file manager that keeps track of a collection (maybe a queue) of temp files and in the background in a single goroutine shared between all the files can carry out the regular task of cleaning up the temp files.

szh: @andytinkham What do you think?

pkg/atomicwriter/atomic_writer.go
```diff
@@ -0,0 +1,109 @@
...
+// takes the file path during construction, so the code which calls
+// `Write()` doesn't need to be concerned with the destination, just like
+// any other writer.
+func NewAtomicWriter(path string, permissions os.FileMode) (io.WriteCloser, error) {
+	dir, file := filepath.Split(path)
+	if dir == "" {
+		dir = "."
+	}
+
+	f, err := ioutil.TempFile(dir, file)
```
szh, line 35: @andytinkham I'm thinking it would be better to put this in a temp directory so it's not in the shared volume. Do you agree?

szh:
> @szh I've added some comments re. one way to mock the `os.Remove` function for testing to your query on the `about-golang` Teams channel (along with the `os.Chmod` and `os.Rename` functions).

Thanks! I've added some unit tests the ideas you posted.

pkg/atomicwriter/atomic_writer.go
andytinkham: It feels like truncate to 0 solves the problem of possibly leaking secrets which is my main concern. If that's easier, I'm ok with it.  For #2, I don't think I have any specific thoughts - seems OK.

szh: OK. So instead of the whole retry logic, we just
1. Attempt to delete the file
2. If that fails, attempt to truncate it (and maybe log it)
3. If that fails, log the error and exit
Is that right?

pkg/atomicwriter/atomic_writer_test.go
```diff
@@ -0,0 +1,243 @@
...
+				// Check that the file exists
+				assert.FileExists(t, path)
+				// Check the contents of the file
+				contents, err := ioutil.ReadFile(path)
+				assert.NoError(t, err)
+				assert.Equal(t, "test content", string(contents))
+				// Check the file permissions
+				mode, err := os.Stat(path)
+				assert.NoError(t, err)
+				assert.Equal(t, os.FileMode(0644), mode.Mode())
```
diverdane, line 42: Coverage looks really good!

For completeness... would it be possible to check that the temp file is successfully deleted in the happy path? (Or maybe I'm missing where that's tested already?) I'm not sure if it's easier to add here (maybe return the `tempFile` in your `writeFile()` function, so it can be checked by the caller?), or some other happy path case.

diverdane (commented):
Looks terrific!
One minor comment/question, otherwise good-to-go!!!

szh:
Requesting security review from @andytinkham

diverdane (approved):
LGTM!!!

andytinkham (approved):
LGTM!

pkg/atomicwriter/atomic_writer.go
szh: Update: putting this in a temp folder results in a `invalid cross-device link` error when trying to call os.rename when the destination directory is on a different volume.

pkg/atomicwriter/atomic_writer_test.go
```diff
@@ -0,0 +1,255 @@
...
+func mockErrorLog() *bytes.Buffer {
+	buf := &bytes.Buffer{}
+	log.ErrorLogger.SetOutput(buf)
+	return buf
+}
+
+// Unmocks the logger output, sending it back to os.Stderr
+func unmockErrorLog() {
+	log.ErrorLogger.SetOutput(os.Stderr)
+}
```
doodlesbykumbi, line 255: In this case it doesn't seem that we have to worry about running tests in parallel but if we did I suspect the mocking and un-mocking might breakdown. To make testing the logs simpler I would have dependency injected the error logger function, much like you did the os funcs, since the atomic writer only uses `log.Error`.

pkg/secrets/provide_conjur_secrets_test.go
```diff
@@ -143,7 +144,7 @@ func TestSecretProvider(t *testing.T) {
 			description:   "sidecar container, happy path",
 			mode:          "sidecar",
 			interval:      time.Duration(10) * time.Millisecond,
-			testTime:      time.Duration(65) * time.Millisecond,
+			testTime:      time.Duration(65+atomicWriteDelayMsecs) * time.Millisecond,
```
doodlesbykumbi, line 147: What's all this about ? (Just curious)

doodlesbykumbi (commented):
Left a couple comments. Otherwise PR looking good

pkg/secrets/provide_conjur_secrets_test.go
szh: In the case where we're doing atomic writes, it adds a very small delay which causes the retry testing to get thrown off (though only on Jenkins, not when running locally). This fixes that.

PR #446, opened 2022-03-05
Don't overwrite secret file if atomic temp file not written

diverdane:
### Desired Outcome

When secret rotation is enabled in Push-to-File mode, and the Secrets Provider (periodically) checks for changes in secret files and finds no change in content, it correctly refrains from writing rendered file content to the temp file, as expected.

However, even though no content has been written to the temp file, there is a deferred call to the atomic file writer's Close() method that unconditionally renames the temp file to the destination secret file, effectively deleting the secret file content (i.e. overwriting the file with an empty file).

Here is a log showing this behavior:
```
INFO:  2022/03/04 23:28:34.506362 conjur_client.go:21: CSPFK002I Creating DAP/Conjur client
ERROR: 2022/03/04 23:28:34.597053 atomic_writer.go:108: CSPFK059E Could not delete temporary file '/conjur/secrets/dummy.yaml2445463986'. Truncated file.%!(EXTRA string=/conjur/secrets/dummy.yaml)
INFO:  2022/03/04 23:28:34.596982 push_to_writer.go:134: CSPFK018I No change in secret file, no secret files written
INFO:  2022/03/04 23:28:34.597159 push_to_writer.go:134: CSPFK018I No change in secret file, no secret files written
ERROR: 2022/03/04 23:28:34.597505 atomic_writer.go:108: CSPFK059E Could not delete temporary file '/conjur/secrets/application.yaml1889516433'. Truncated file.%!(EXTRA string=/conjur/secrets/application.yaml)
INFO:  2022/03/04 23:28:34.597514 provide_conjur_secrets.go:94: CSPFK015I DAP/Conjur Secrets pushed to shared volume successfully
```

And inside the container, the secret files have no content:
```
/conjur/secrets $ ls -l
total 0
-rw-r--r--    1 nobody   nobody           0 Mar  4 23:30 application.yaml
-rw-r--r--    1 nobody   nobody           0 Mar  4 23:30 dummy.yaml
/conjur/secrets $ 
```
The desired outcome is:
- The destination secret file shouldn't be overwritten (contents cleared) when secrets haven't changed.
- We shouldn't attempt to delete the temp file if it's been renamed to the target secret file name.

### Implemented Changes

- The secret file is no longer overwritten by the temp file if the temp file hasn't been written.
- Deleting of the temp file is no longer attempted if it's been renamed to the target secret file.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-17885](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-17885)

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/pkg/atomicwriter/atomic_writer_test.go
+++ b/pkg/atomicwriter/atomic_writer_test.go
@@ -6,21 +6,112 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"syscall"
 	"testing"
 
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
 	"github.com/stretchr/testify/assert"
 )
 
-type assertFunc func(path string, tempFilePath string, t *testing.T, err error)
-type errorAssertFunc func(buf *bytes.Buffer, wc io.WriteCloser, t *testing.T, err error)
+// osFuncCounts struct is used to track the number of calls that were
+// made to each OS function during an individual atomic writer test case.
+type osFuncCounts struct {
+	chmodCount    int
+	renameCount   int
+	removeCount   int
+	syncCount     int
+	tempFileCount int
+	truncateCount int
+	writeCount    int
+}
 
-func TestWriteFile(t *testing.T) {
+// injectErrors define errors which should be returned (i.e. simulated) for
+// OS functions for a given atomic writer test case.
+type injectErrors struct {
+	chmodErr    error
+	renameErr   error
+	removeErr   error
+	syncErr     error
+	tempFileErr error
+	truncateErr error
+	writeErr    error
+}
+
+// newTestOSFuncs creates an osFuncs table of OS functions, each of which
+// wraps the equivalent standard OS function, and adds the following
+// functionality for testing:
+//
+//    - Track the number of times that the function was called
+//    - Optionally return an error for testing purposes
+//
+func newTestOSFuncs(injectErrs injectErrors) (osFuncs, *osFuncCounts) {
+	counts := &osFuncCounts{}
+	funcs := osFuncs{
+		chmod: func(filename string, mode os.FileMode) error {
+			counts.chmodCount++
+			if injectErrs.chmodErr != nil {
+				return injectErrs.chmodErr
+			}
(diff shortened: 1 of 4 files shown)


szh:
Good catch!

pkg/atomicwriter/atomic_writer_test.go
```diff
@@ -150,47 +283,41 @@ func TestLogsErrors(t *testing.T) {
...
+			path:       "test_file.txt",
+			injectErrs: injectErrors{errOnChmod: true},
+			assert: func(t *testing.T, buf *bytes.Buffer, wc io.WriteCloser, err error) {
 				assert.NoError(t, err)
 				assert.Contains(t, buf.String(), "Could not set permissions on temporary file")
 
 				// Check that the file was still renamed
 				writer, ok := wc.(*atomicWriter)
 				assert.True(t, ok)
-				assert.NoFileExists(t, writer.tempFile.Name())
```
szh, line 190: I wonder if we should actually add this line to more of the failure tests. Thoughts?

szh (commented):
Looks fantastic. Just left one comment.

pkg/atomicwriter/atomic_writer_test.go
diverdane: Yeah, we can add checks that the temp file does not exist, but we'll have to capture/record the temp file name just after we call writer.Write(), since the `tempFile` pointer in the `atomicWriter` struct is now getting nil'd out whenever the temp file has in fact been deleted.

I'll add the capture of the temp file name and the checks that the temp file has been deleted where it makes sense.

I also noticed that we're not checking for `IsNotExist(err)` when the call to `os.remove()` fails, so I'll add that in.

Also, when `os.remove()` fails, and then we do an `os.truncate()`, if the truncate ever returns `os.ErrNotExist` (granted, this should theoretically never happen, but for the purposes of safe coding...), we should just set `writer.tempFile = nil` and return without any error messages.

szh (approved):
Looks great. 🥇

PR #447, opened 2022-03-09
Delete secrets files when retrieval fails

szh:
### Desired Outcome

Delete secret file if Conjur returns a 403 or 404 status code.

### Implemented Changes

When secrets rotation is enabled, if a batch fetch call to Conjur fails during a refresh due to a deleted secret or revoked access, delete the secrets files from the shared volume to avoid leaving unauthorized secrets on the file system. Currently there's no support for specifying which secret(s) failed, so we need to delete all secrets included in the fetch, which includes all secret groups. A future effort will improve this.
This feature can be disabled using an annotation, `conjur.org/remove-deleted-secrets-enabled: "false"`.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-17938](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-17938)

### Definition of Done
merged.*

- [x] Secret files are deleted if a secret doesn't exist or isn't authorized
- [x] An annotation exists to disable this functionality
- [x] UT exists
- [x] Integration tests exist

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] Docs (e.g. `README`s) were updated in this PR

#### Behavior

#### Security

- [x] Security architect has reviewed the changes in this PR,

### DIFF
--- a/pkg/secrets/config/config.go
+++ b/pkg/secrets/config/config.go
@@ -284,7 +284,6 @@ func NewConfig(settings map[string]string) *Config {
 	// ignore errors here, if the interval string is null, zero is returned
 	refreshInterval, _ := time.ParseDuration(refreshIntervalStr)
 
-	// TODO: Should we only enable this when rotation is enabled?
 	sanitizeEnableStr := settings[RemoveDeletedSecretsKey]
 	sanitizeEnable := parseBoolFromStringOrDefault(sanitizeEnableStr, DefaultSanitizeEnabled)
 
--- a/pkg/secrets/config/config.go
+++ b/pkg/secrets/config/config.go
@@ -14,27 +14,46 @@ import (
 // Constants for Secrets Provider operation modes,
 // and Defaults for some SP settings
 const (
-	K8s                     = "k8s_secrets"
-	File                    = "file"
-	ConjurMapKey            = "conjur-map"
-	DefaultRetryCountLimit  = 5
-	DefaultRetryIntervalSec = 1
-	MinRetryValue           = 0
-	MinRefreshInterval      = time.Second
-	DefaultRefreshIntervalStr  = "5m"
-
+	K8s                       = "k8s_secrets"
+	File                      = "file"
+	ConjurMapKey              = "conjur-map"
+	DefaultRetryCountLimit    = 5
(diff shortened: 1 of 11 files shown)


deploy/test/test_in_docker.sh
```diff
@@ -5,6 +5,9 @@ set -xeuo pipefail
 
 # Clean up when script completes and fails
 finish() {
+  dump_conjur_namespace_upon_error
```
diverdane, line 8: Thank you for putting this in!!!!

pkg/secrets/config/config.go
```diff
@@ -254,21 +276,26 @@ func NewConfig(settings map[string]string) *Config {
...
-	refreshEnableStr   := settings[SecretsRefreshEnabledKey]
+	refreshEnableStr := settings[SecretsRefreshEnabledKey]
 
 	if refreshIntervalStr == "" && refreshEnableStr == "true" {
 		refreshIntervalStr = DefaultRefreshIntervalStr
 	}
 	// ignore errors here, if the interval string is null, zero is returned
 	refreshInterval, _ := time.ParseDuration(refreshIntervalStr)
 
+	// TODO: Should we only enable this when rotation is enabled?
```
diverdane, line 287: I'm wondering if we need this TODO. Seems that the checks for 403/404 errors (and subsequent sanitizing) would only occur if the refresh periodic is in effect, and that's only going to happen when rotation is enabled?

diverdane (commented):
Changes look good! Thanks for all of the gratuitous, style-related cleanup, and for adding in the Namespace resource dumps on E2E test failure.

pkg/secrets/config/config.go
szh: Hah, I actually removed this but forgot to commit it :)

deploy/test/test_in_docker.sh
szh: I ended up removing it because I was getting "kubectl: command not found" errors... Not sure what that's about but I don't want it to hold this up.

diverdane: Ah, yeah, the 'kubectl' commands would need to be run via a test client container.

PR #448, opened 2022-03-15
Implement detection of changes in K8s Secret content with UT

rpothier:
### Desired Outcome

The Kubernetes secrets should only be updated when there is new content.

### Implemented Changes

Updated to detect changes the k8s secrets values and Conjur secrets.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-17475](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-17475)

### Definition of Done
- [] K8s secrets are updated only when there are changes in the k8s secrets content or Conjur secrets.
#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/secrets-provider-for-k8s#447](https://github.com/cyberark/secrets-provider-for-k8s/pull/447)
 - Secrets Provider allows for its status to be monitored through the creation of a couple of empty sentinel files: `CONJUR_SECRETS_PROVIDED` and `CONJUR_SECRETS_UPDATED`. The first file is created when SP has completed its first round of providing secrets via secret files / Kubernetes Secrets. It creates/recreates the second file whenever it has updated secret files / Kubernetes Secrets. If desirable, application containers can mount these files via a shared volume.
   [cyberark/secrets-provider-for-k8s#450](https://github.com/cyberark/secrets-provider-for-k8s/pull/450)
+- Adds support for secrets rotation with Kubernetes Secrets.
+  [cyberark/secrets-provider-for-k8s#448](https://github.com/cyberark/secrets-provider-for-k8s/pull/448)
 
 ## [1.4.0] - 2022-02-15
 
--- a/ROTATION.md
+++ b/ROTATION.md
@@ -71,7 +71,15 @@ There are two new annotations introduced and one annotation is updated.
 Prerequisites:
 
 Requires secrets-provider-for-k8s v1.4.0 or later.
+
+<details><summary>For Push to File mode</summary>
+
 Follow the procedure to set up Secrets Provider for [Push to File](PUSH_TO_FILE.md#set-up-secrets-provider-for-push-to-file)
+</details>
+<details><summary>For Kubernetes Secrets mode</summary>
+
+Follow the procedure to set up [Kubernetes Secrets](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/k8s-ocp/cjr-k8s-secrets-provider-ic.htm?tocpath=Integrations%7COpenShift%252FKubernetes%7CSet%20up%20applications%7CSecrets%20Provider%20for%20Kubernetes%7CInit%20container%7C_____1#SetupSecretsProviderasaninitcontainer)
+</details>
 
 Modify the Kubernetes manifest
 1. Change the Secrets provider container to be a sidecar. If it was configured
@@ -135,7 +143,5 @@ This feature is a **Community** level project that is still under development.
 There could be changes to the documentation so please check back often for updates and additions.
 Future enhancements to this feature will include:
 
-- Support for secrets rotation of Kubernetes secrets
 - atomic writes for multiple Conjur secret values
 - reporting of multiple errored secret variables for bulk Conjur secret retrieval and selective deletion of secret values from files
-- atomic write of secret files
--- a/pkg/log/messages/info_messages.go
+++ b/pkg/log/messages/info_messages.go
@@ -31,3 +31,4 @@ const CSPFK016I string = "CSPFK016I There are no secrets to be retrieved from Co
 const CSPFK017I string = "CSPFK017I Creating default file name for secret group '%s'"
 const CSPFK018I string = "CSPFK018I No change in secret file, no secret files written"
 const CSPFK019I string = "CSPFK019I Error fetching secrets, deleting secrets file"
+const CSPFK020I string = "CSPFK020I No change in Kubernetes secret, no secrets updated"
--- a/pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
+++ b/pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
@@ -47,6 +47,8 @@ func (m testMocks) setPermissions(denyConjurRetrieve, denyK8sRetrieve,
 	}
 	if denyK8sUpdate {
 		m.kubeClient.CanUpdate = false
+	} else {
+		m.kubeClient.CanUpdate = true
 	}
 }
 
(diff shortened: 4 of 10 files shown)


pkg/log/messages/info_messages.go
```diff
@@ -31,3 +31,4 @@ const CSPFK016I string = "CSPFK016I There are no secrets to be retrieved from Co
 const CSPFK017I string = "CSPFK017I Creating default file name for secret group '%s'"
 const CSPFK018I string = "CSPFK018I No change in secret file, no secret files written"
 const CSPFK019I string = "CSPFK019I Error fetching secrets, deleting secrets file"
+const CSPFK020I string = "CSPFK018I No change in Kubernetes secret, no secrets updated"
```
szh, line 34:
```suggestion
const CSPFK020I string = "CSPFK020I No change in Kubernetes secret, no secrets updated"
```

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -372,6 +375,7 @@ func TestProvide(t *testing.T) {
 		for secretName, secretData := range tc.k8sSecrets {
 			mocks.kubeClient.AddSecret(secretName, secretData)
 		}
+		prevSecretsChecksums = map[string]utils.Checksum{}
```
szh, line 378: Doesn't look like this is used anywhere

szh (changes requested):

pkg/utils/checksum.go
```diff
@@ -0,0 +1,27 @@
+package utils
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"io"
+)
+
+type Checksum []byte
+func FileChecksum(buf *bytes.Buffer) (Checksum, error) {
```
szh, line 10: I noticed this PR decreases the code coverage a bit. Should we add tests for this file?

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
rpothier: This is needed for TestProvide() to zero out prevSecretsChecksums so that it writes values in this test.

diverdane: Oooh.... I'm glad @szh asked this question!!!  I don't think we should be using a global variable for `prevSecretsChecksums` here, or for the Push-to-File provider either. The issue that we might face with using global variables is that if our UT test cases are ever run in parallel, then one test case could potentially stomp on another test case's values in the global `prevSecretsChecksums`.

I believe the `prevSecretsChecksums` should be moved to the `K8sProvider` struct. Each UT test case should be creating its own `K8sProvider` to test with, so each one should be created with its own `prevSecretsChecksums`. Thinking about this a bit more, this would also help in the future if we should ever need to run multiple `K8sProviders` in parallel.

We need to fix this for the Push-to-File provider as well. I don't know if you want to tackle that here or separately.

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -372,6 +375,7 @@ func TestProvide(t *testing.T) {
 		for secretName, secretData := range tc.k8sSecrets {
```
diverdane, line 375: While we're changing these unit tests, would you mind modifying the way that we're invoking test cases so that each is called via `t.Run(tc.desc ...)`. What the iterations would look like is:

```
	for _, tc := range testCases {
		t.Run(tc.desc, func(t *testing.T) {
			// Set up test case
			mocks := newTestMocks()
			. . .
```

The advantage to doing this is that if/when UT failures occur, the `tc.desc` is displayed prominently in the failure summary, and it makes it very obvious which test case(s) have failed.

Also, I believe that we would need to run the test cases via `t.Run(...)` if we should ever want to run the test cases in parallel (reference: https://eleni.blog/2019/05/11/parallel-test-execution-in-go/).

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -383,3 +387,91 @@ func TestProvide(t *testing.T) {
 		}
 	}
 }
+
+func Test_secrets_contentChanges(t *testing.T) {
```
diverdane, line 391: Ruh-Roh, the name of this test function uses `snake_case`.  (Function name should be `TestSecretContentChanges`).

I fear we've been doing too many context switches between Bash and Go.

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -383,3 +387,91 @@ func TestProvide(t *testing.T) {
...
+	provider := mocks.newProvider(requiredSecrets)
+	err := provider.Provide()
+	assert.False(t, mocks.logger.InfoWasLogged(messages.CSPFK020I))
+	assertSecretsUpdated(
+		expectedK8sSecrets{
+			"k8s-secret1": {"secret1": "secret-value1"},
+		},
+		expectedMissingValues{} ) (t, mocks, err, desc)
+
+	// Call Provide again, verify is doesn't try to update the secret
```
diverdane, line 422: Typo, should be "verify it doesn't ...".

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -383,3 +387,91 @@ func TestProvide(t *testing.T) {
...
+	// as there should be an error if it tried to write the secrets
+	denyK8sUpdate = true
```
diverdane, line 424: That's a clever way to detect that it's not trying to update! :)

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -383,3 +387,91 @@ func TestProvide(t *testing.T) {
...
+	mocks.setPermissions(denyConjurRetrieve, denyK8sRetrieve , denyK8sUpdate)
+	err = provider.Provide()
+	assert.NoError(t, err)
+	assert.True(t, mocks.logger.InfoWasLogged(messages.CSPFK020I))
+	// verify the same secret still exists
+	assertSecretsUpdated(
+		expectedK8sSecrets{
+			"k8s-secret1": {"secret1": "secret-value1"},
+		},
+		expectedMissingValues{} ) (t, mocks, err, desc)
```
diverdane, line 434: Something that might be worth considering here... this test involves a whole bunch of operations, and if when there's a test failure, it won't be immediately obvious which assert in this test is failing (and where in the sequence of operations it's failing), without having to match up line numbers in the source code. What you could do is to modify the `desc` string as you work your way through the sequence. That way, the failure summary for the assert that fails will provide us with details on where exactly it's failing.

szh (commented):
Looks great. I left a couple docs comments.

ROTATION.md
```diff
@@ -71,7 +71,13 @@ There are two new annotations introduced and one annotation is updated.
 Prerequisites:
 
 Requires secrets-provider-for-k8s v1.4.0 or later.
+
+<details><summary>For Push to File mode</summary>
```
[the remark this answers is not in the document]
rpothier: Good catch!

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/secrets-provider-for-k8s#447](https://github.com/cyberark/secrets-provider-for-k8s/pull/447)
 - Secrets Provider allows for its status to be monitored through the creation of a couple of empty sentinel files: `CONJUR_SECRETS_PROVIDED` and `CONJUR_SECRETS_UPDATED`. The first file is created when SP has completed its first round of providing secrets via secret files / Kubernetes Secrets. It creates/recreates the second file whenever it has updated secret files / Kubernetes Secrets. If desirable, application containers can mount these files via a shared volume.
   [cyberark/secrets-provider-for-k8s#450](https://github.com/cyberark/secrets-provider-for-k8s/pull/450)
+- Kubernetes Secrets are only updated when there are changes when secrets rotation is enabled.
+  [cyberark/secrets-provider-for-k8s#448](https://github.com/cyberark/secrets-provider-for-k8s/pull/448)
```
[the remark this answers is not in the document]
rpothier: How about "Adds support for secrets rotation with Kubernetes Secrets."

szh (approved):
LGTM!

PR #457, opened 2022-04-05
Fix rotation tests on OpenShift

szh:
### Implemented Changes

- Fixed integration test yaml that mistakenly had the postStart lifecycle hook in the app container instead of the secrets-provider container, and added livenessProbe to restart the app container when secrets change. This should fix testing reliability on OpenShift.

- Also removed the secret deletion portion of the rotation tests in k8s-secrets mode. This test was unreliable because the removal of the secret caused the app container to fail before starting. We would need another way to test this other than checking that the environment variable is cleared. Being that rotation is mainly used with Push-to-File, it may not be worth going through the effort to develop tests for this scenario.

### Connected Issue/Story

N/A

### Definition of Done

- [x] CI passes consistently (including OCP)

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/deploy/config/k8s/test-env-k8s-rotation.sh.yml
+++ b/deploy/config/k8s/test-env-k8s-rotation.sh.yml
@@ -36,6 +36,18 @@ spec:
         name: test-app
         command: ["sleep"]
         args: ["infinity"]
+        livenessProbe:
+          exec:
+            command:
+            - /mounted/status/conjur-secrets-unchanged.sh
+          failureThreshold: 1
+          initialDelaySeconds: 5
+          periodSeconds: 5
+          successThreshold: 1
+          timeoutSeconds: 1
+        volumeMounts:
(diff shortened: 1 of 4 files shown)


rpothier (commented):
LGTM! Should we add a comment or some way to remind us in the future that the secrets deletion is not being tested for k8s secrets mode?

szh:
> LGTM! Should we add a comment or some way to remind us in the future that the secrets deletion is not being tested for k8s secrets mode?

I added a comment [here](https://github.com/cyberark/secrets-provider-for-k8s/pull/457/files#diff-942903317b8ee55d62b3222e64de80c8b0489dd66ffcab4285efde6f3508ea3fR21)

rpothier (approved):

PR #462, opened 2022-04-29
Bump version to 1.4.1

jtuttle:


szh:
Do we still need to update `version.go` with the new release process?

jtuttle:
@szh Yes we do, but for some reason `version.go` was already set to 1.4.1 which is why it doesn't appear in this PR.

szh:
Doesn't look like it to me...
https://github.com/cyberark/secrets-provider-for-k8s/blob/main/pkg/secrets/version.go#L6

jtuttle:
Weird, that file isn't mentioned at all in the `CONTRIBUTING.md` release instructions. I'll add it in this branch.

jtuttle:
@szh Okay, all set.

PR #466, opened 2022-05-05
Remove docker tag echo statements

neil-k-zero:
### Desired Outcome

- Remove gratuitous echo statements

### Implemented Changes

- Remove echo statement that prevents a RedHat image from being tagged during promotion.

#### Changelog

  CHANGELOG update


szh (approved):
👍

PR #478, opened 2022-07-13
Added more replace statements for golang.org/x/net and crypto

andytinkham:
Signed-off-by: Andy Tinkham <andy.tinkham@cyberark.com>

### Desired Outcome

The last round of replaces brought in new commit references for golang.org/x/net and golang.org/x/crypto that were vulnerable versions. Prune those new versions from the tree. 

### Implemented Changes

Pruned the newly introduced versions from the dependency tree.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -88,6 +88,8 @@ replace golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550 => golang.org/x/c
 
 replace golang.org/x/crypto v0.0.0-20201002170205-7f63de1d35b0 => golang.org/x/crypto v0.0.0-20220525230936-793ad666bf5e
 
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -66,11 +66,11 @@ require (
 
 // Automated release process replaces
 // DO NOT EDIT: CHANGES TO THE 3 BELOW LINES WILL BREAK AUTOMATED RELEASES
-replace github.com/cyberark/conjur-api-go => github.com/cyberark/conjur-api-go latest
+replace github.com/cyberark/conjur-api-go => github.com/cyberark/conjur-api-go v0.10.1
```
rpothier, line 69: This won't break the automated releases?

andytinkham: Yep. it will. Thought the build would fail for that. Hmmm. I'll update. Thanks, @rpothier!

PR #500, opened 2023-02-13
Enable SP to retrieve and provide German umlaut chars

john-odonnell:
### Desired Outcome

Enable Secrets Provider to deliver secrets with special characters/binary values.

### Implemented Changes

In [conjur#1989](https://github.com/cyberark/conjur/pull/1989), the `/secrets` endpoint was updated to allow for the batch retrieval of secrets with binary values - adding the `Accept-Encoding` header set to `base64` means Conjur will base64 encode secret values before attempting to render response JSON.

We can use that functionality here to enable Secrets Provider to deliver binary/special character secrets by switching `client.RetrieveBatchSecrets()` to `client.RetrieveBatchSecretsSafe()`.

### Connected Issue/Story

CNJR-466

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/deploy/policy/load_policies.sh
+++ b/deploy/policy/load_policies.sh
@@ -35,6 +35,7 @@ done
 conjur variable values add secrets/test_secret "some-secret"
 conjur variable values add "secrets/var with spaces" "some-secret"
 conjur variable values add "secrets/var+with+pluses" "some-secret"
+conjur variable values add "secrets/umlaut" "some-secret"
 conjur variable values add secrets/url "postgresql://test-app-backend.app-test.svc.cluster.local:5432"
 conjur variable values add secrets/username "some-user"
 conjur variable values add secrets/password "7H1SiSmYp@5Sw0rd"
--- a/pkg/secrets/clients/conjur/conjur_client.go
+++ b/pkg/secrets/clients/conjur/conjur_client.go
@@ -8,13 +8,13 @@ import (
 )
 
 /*
-	Client for communication with Conjur. In this project it is used only for
-    batch secrets retrieval so we expose only this method of the client.
+Client for communication with Conjur. In this project it is used only for
+batch secrets retrieval so we expose only this method of the client.
 
-	The name ConjurClient also improves readability as Client can be ambiguous.
+The name ConjurClient also improves readability as Client can be ambiguous.
 */
 type ConjurClient interface {
-	RetrieveBatchSecrets([]string) (map[string][]byte, error)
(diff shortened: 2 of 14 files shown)


deploy/policy/load_policies.sh
```diff
@@ -35,6 +35,7 @@ done
 conjur variable values add secrets/test_secret "some-secret"
 conjur variable values add "secrets/var with spaces" "some-secret"
 conjur variable values add "secrets/var+with+pluses" "some-secret"
+conjur variable values add "secrets/umlaut" "some-secret"
```
jtuttle, line 38: Do you not need an umlaut in this secret value? Or is this just a placeholder that gets replaced later on?

pkg/secrets/clients/conjur/conjur_client.go
```diff
@@ -8,13 +8,13 @@ import (
...
-    batch secrets retrieval so we expose only this method of the client.
+Client for communication with Conjur. In this project it is used only for
+batch secrets retrieval so we expose only this method of the client.
 
-	The name ConjurClient also improves readability as Client can be ambiguous.
+The name ConjurClient also improves readability as Client can be ambiguous.
 */
 type ConjurClient interface {
-	RetrieveBatchSecrets([]string) (map[string][]byte, error)
+	RetrieveBatchSecretsSafe([]string) (map[string][]byte, error)
```
jtuttle, line 17: Are we sure it's totally safe now or are there other corner cases that might sneak in? :)

deploy/policy/load_policies.sh
john-odonnell: It's a [placeholder](https://github.com/cyberark/secrets-provider-for-k8s/blob/main/deploy/policy/load_policies.sh#L32-L34) that's replaced when a test case runs [[test case](https://github.com/cyberark/secrets-provider-for-k8s/pull/500/files#diff-d88d31cb2f68e87775b2b9c5ee06811ba27e2392fdce07440f3a59b7211ff675R8)][[CLI variable load](https://github.com/cyberark/secrets-provider-for-k8s/blob/main/deploy/utils.sh#L504)].

pkg/secrets/clients/conjur/conjur_client.go
jtuttle: Oh right, you're just using a different `conjur-api-go` method? Nah, that's fine then.

gl-johnson (approved):
LGTM!

PR #502, opened 2023-02-24
Base64 Secrets Decoding POC

gl-johnson:
### Desired Outcome

*Please describe the desired outcome for this PR.  Said another way, what was
the original request that resulted in these code changes?  Feel free to copy
this information from the connected issue.*

### Implemented Changes

*Describe how the desired outcome above has been achieved with this PR. In
particular, consider:*

- _What's changed? Why were these changes made?_
- _How should the reviewer approach this PR, especially if manual tests are required?_
- _Are there relevant screenshots you can add to the PR description?_

### Connected Issue/Story

Resolves #[relevant GitHub issue(s), e.g. 76]

CyberArk internal issue ID: [insert issue ID]

### Definition of Done
*At least 1 todo must be completed in the sections below for the PR to be
merged.*

#### Changelog

- [ ] The CHANGELOG has been updated, or
- [ ] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

- [ ] This PR includes new unit and integration tests to go with the code
  changes, or
- [ ] The changes in this PR do not require tests

#### Documentation

- [ ] Docs (e.g. `README`s) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]
- [ ] This PR does not require updating any documentation

#### Behavior

- [ ] This PR changes product behavior and has been reviewed by a PO, or
- [ ] These changes are part of a larger initiative that will be reviewed later, or
- [ ] No behavior was changed with this PR

#### Security

- [ ] Security architect has reviewed the changes in this PR,
- [ ] These changes are part of a larger initiative with a separate security review, or
- [ ] There are no security aspects to these changes

PR #504, opened 2023-03-02
Base64 Secrets Decoding Design

gl-johnson:
Add design doc for adding configurable decoding of Conjur secrets stored in base64.

POC branch can be found [here](https://github.com/cyberark/secrets-provider-for-k8s/pull/502)

### DIFF
--- a/design/secrets_decoding_design.md
+++ b/design/secrets_decoding_design.md
@@ -1,0 +1,207 @@
+# Secrets Provider - Base64 Secrets Decoding Design
+
+## Table of Contents
+
+- [Secrets Provider - Base64 Secrets Decoding Design](#secrets-provider---base64-secrets-decoding-design)
+  * [Table of Contents](#table-of-contents)
+  * [Useful links](#useful-links)
+  * [Background](#background)
+  * [Solution](#solution)
+    + [Kubernetes Secrets Mode](#kubernetes-secrets-mode)
+    + [P2F Mode](#p2f-mode)
+  * [Design](#design)
+    + [Kubernetes Secrets Mode](#kubernetes-secrets-mode-1)
+    + [P2F Mode](#p2f-mode-1)
+    + [Backwards compatibility](#backwards-compatibility)
+  * [Test Plan](#test-plan)
+    + [Annotation Tests (K8s Secrets)](#annotation-tests--k8s-secrets-)
+    + [Annotation Tests (P2F)](#annotation-tests--p2f-)
+    + [Decoding Tests](#decoding-tests)
+    + [e2e Tests](#e2e-tests)
+    + [Audit](#audit)
+  * [Documentation](#documentation)
+  * [Open questions](#open-questions)
+  * [Tasks](#tasks)
+    + [Add a struct for storing content-types of individual Conjur secrets (K8s Secrets)](#add-a-struct-for-storing-content-types-of-individual-conjur-secrets--k8s-secrets-)
+    + [Add/update a struct for storing content-types of individual Conjur secrets (P2F)](#add-update-a-struct-for-storing-content-types-of-individual-conjur-secrets--p2f-)
+    + [Add decoding of secrets with 'base64' content-types](#add-decoding-of-secrets-with--base64--content-types)
+    + [Update e2e tests and dev environment scripts/manifests](#update-e2e-tests-and-dev-environment-scripts-manifests)
+    + [Create/update docs](#create-update-docs)
+    + [Docs walkthrough](#docs-walkthrough)
+
+## Useful links
+
+| Name                                     | Link                                                         |
+| -----------------------------------------| ------------------------------------------------------------ |
+| Secrets Provider Configuration Reference | https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/k8s-ocp/cjr-k8s-secrets-provider-ref.htm |
+| Feature Functional Spec (internal)       | https://{cyberark_confluence_host}/display/rndp/Secrets+Provider%3A+Secrets+decoding+for+k8s+Functional+Specification |
+
+## Background
+Secrets Provider provides Kubernetes-based applications with access to secrets that are stored and managed in Conjur. These secrets can be any binary/string data. In some cases it may be beneficial, or necessary, to store secret values in Conjur as base64 encoded strings depending on the nature and origin of the secret. For example, Vault/EPV requires that binary secrets are stored with base64 encoding which can result in these encoded secrets being synchronized to Conjur.
+
+We would like to increase flexibility of consuming applications by introducing a mechanism for Secrets Provider to decode individual secrets on the fly. This would allow Secrets Provider to work with applications where it may be impossible to decode the values, i.e. when running compiled 3rd party applications.
+
+Currently it is possible to decode secrets only when using Push-to-file mode (P2F), and when defining custom templates for the secrets file. See: [P2F - Additional Template Functions](https://github.com/cyberark/secrets-provider-for-k8s/blob/main/PUSH_TO_FILE.md#additional-template-functions). This should be extended to work with K8s secrets mode via a common notation to simplify usage of the feature, as well as decode secrets without needing a custom template definition in P2F mode.
+
+## Solution
+### Kubernetes Secrets Mode
+Kubernetes secrets mode relies on a secrets manifest to configure Conjur secrets, rather than pod annotations. In this case we should extend the schema for secrets manifests to include an optional content-type on conjur secrets. This could look like:
+```
+apiVersion: v1
+kind: Secret
+metadata:
+  name: test-k8s-secret
+type: Opaque
+stringData:
+  conjur-map: |-
+    secret: secrets/test_secret
+    user: 
+      path: policy/path/base64_user
+      content-type: base64
+```
+
+### P2F Mode
+Kubernetes Pod annotations are currently used to define which Conjur secrets are to be retrieved when using P2F (push-to-file) mode. [See annotations reference](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/k8s-ocp/cjr-k8s-secrets-provider-ref.htm?tocpath=Integrations%7COpenShift%252FKubernetes%7CApp%20owner%253A%20Set%20up%20workloads%20in%20Kubernetes%7CSet%20up%20workloads%20(JWT-based%20authn)%7CSecrets%20Provider%20for%20Kubernetes%7C_____3#AdditionalannotationsforPushtoFilemode). The schema for the `conjur.org/conjur-secrets.{secret-group}` annotations can be extended to include an optional content-type attribute. This could look like:
+```
+conjur.org/conjur-secrets.groupname: |
+  - user: policy/path/base64_user     # valid annotation with content-type
+    content-type: base64
+  - policy/path/base64_password:      # this is also valid when no alias is needed
+      content-type: base64
+```
+
+P2F already leverages Kubernetes' Downward API and volumeMounts to pass these annotations into the Secrets Provider container. This will make it possible to determine which secrets should be decoded before pushing them to the Kubernetes secret file. 
+
+## Design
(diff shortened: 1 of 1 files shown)


adamouamani:
Hi @gl-johnson  Thanks for putting this together.  I had a couple of comments/questions regarding test scenarios.  please see below..

1) how do we handle case with \n etc is part of value - this could cause issue with base64 encoding

2) uppercase or mixed case values for text and base64 - for example does Base64 work?

3) for E2E tests - if we do happy path I see 4 scenarios for text and base64 annotation with k8s secrets and p2f

4) Do we have unit tests with original values containing special chars - such as umlat

5) Do we trim whitespace - for example if you have a string " h e l l o " - are leading and trailing spaces trimmed before converting to base64

6) Base64 can include a slash '/' - does this cause any issue with P2F?

niteshtaneja (commented):
Can we make sure to test with -

- binary secrets > 65K (65K is the limit for Vault apparently and since customer has secrets >65K, they are going to store them directly in Conjur)
- rotated secrets (The secrets are expected to be rotated every 3 weeks by the customer)

design/secrets_decoding_design.md
```diff
@@ -0,0 +1,207 @@
...
+  - <alias_value>: <path_value>               # to be added
+    content-type: <content_type_value>
+  - <path_value>:                             # to be added
+      content-type: <content_type_value>
+```
+
+### Backwards compatibility
+This enhancement will not modify the default behavior of Secrets Provider. It will continue to assume that secrets should be delivered "as-is" unless it specifically finds a content-type annotation which indicates that the secret value should be decoded.
+
+## Test Plan
```
andytinkham, line 114: Is it a correct assumption that if someone mistakenly puts the same secret twice in the annotation and gives them different types, the one defined later in the file would be the type used? Do we need any tests around that?

andytinkham (approved):
There isn't a security concerns section in this, so I'll put a general review comment in for it. I forget if Secrets Provider does any memory clearing to make sure that secret values don't persist in memory after they're provided, but I think it does. We'll need to make sure that this cleanup mechanism (if it exists) cleans up both the encoded and the decoded secret values.  I've got a note to check on this in the final security review.

Other than that, approved by security.

design/secrets_decoding_design.md
gl-johnson: That would likely be the case if they are exactly the same alias/path and in the same manifest.

Although we should support the same Conjur secret with multiple content-types as long as it is in different secret groups (P2F) or under a different kubernetes secret. 

For example:
```
conjur.org/conjur-secrets.decoded-group: |
  - mysecret: /path/to/secret
    content-type: base64
conjur.org/conjur-secrets.encoded-group: |
  - mysecret: /path/to/secret
    content-type: text
```
Both definitions can exist at the same time but the value for mysecret in `decoded-group` will be decoded while the value for mysecret in `encoded-group` will be provided as its original base64 value.

gl-johnson:
> Hi @gl-johnson Thanks for putting this together. I had a couple of comments/questions regarding test scenarios. please see below..
> 
> 1. how do we handle case with \n etc is part of value - this could cause issue with base64 encoding

We will be using the Golang `base64` pkg to do the decoding so any valid base64 strings shouldn't pose an issue. If a newline exists in an encoded string, it is represented as \n in the output. It should also handle escape sequences fine.

> 2. uppercase or mixed case values for text and base64 - for example does Base64 work?

In the POC, no. Any string that doesn't match exactly "base64" would be treated as a 'text' secret and not attempt decoding. I believe all the existing annotations in Secrets Provider are also case-sensitive.

> 3. for E2E tests - I see 4 happy path scenarios for text and base64 annotation with k8s secrets and p2f

I agree that each mode needs its own e2e test. Although I don't believe we need to explicitly test the 'text' content type since it is the default cause which should be covered sufficiently covered by existing tests. We would basically be adding two extra e2e tests just to verify that the parsing works, which should be handled in unit tests.

It would probably be worthwhile to have two additional e2e tests for rotation of encoded secrets too per @niteshtaneja's comment

> 4. Do we have unit tests with original values containing special chars - such as umlaut

Secrets Provider has unit tests for retrieving Conjur secrets containing special characters. I don't know that we need to test the same for secrets decoding, since we would just be validating that the Golang base64 pkg works as expected.

> 5. Do we trim whitespace - for example if you have a string " h e l l o " - are leading and trailing spaces trimmed before converting to base64

Secrets Provider is simply going to provide the value output by `base64.StdEncoding.DecodeString(secretValueFromConjur)` with no additional processing on the value

> 6. Base64 can include a slash '/' - does this cause any issue?

Good question - I see that `/` and `+` are part of the base64 library. A few sample test strings I've tried seem to be supported. Again I think this should be tested sufficiently in the base64 library and probably doesn't need explicit testing in Secrets Provider

> 7. How could this fail -what are some failure/error scenarios and how do we report these?

Failure on configuration:
- Fail to parse annotations
   - Malformed secret (alias/path)
     - Log YAML parsing error
     - Secrets Provider fails to start
   - Malformed content-type
     - If YAML can still be parsed, create a secret with default content type (text)
     - Log (warning)

Failure on providing secrets:
- Fail to decode an encoded secret (invalid base64)
   - Log (warning)
   - Provide original value
   - Secrets provider continues to run

> 8. how will customers use them - is there any size limit to data being encoded?

According to the docs, Conjur's maximum variable size is 10MB so that would be the upper limit as far as I'm aware.


Will get this document and Jira tasks updated based on the feedback/questions. Thanks!

design/secrets_decoding_design.md
```diff
@@ -173,35 +179,63 @@ Unit tests for any helper functions added to assist with decoding secrets.
...
+
+Failure on configuration:
+  - Malformed secret (alias/path)
+    - Log YAML parsing error
+    - Secrets Provider fails to start
+  - Malformed content-type
+    - If YAML can still be parsed, create a secret with default content type (text)
+    - Log (warning)
+
+Failure on providing secrets:
```
andytinkham, line 200: Do we already log a request for a non-existent secret or a secret the API key used doesn't have access to? I'd assume so, but we definitely should if we don't already.

gl-johnson: Yes, both cases log the resulting 404 if I recall

jvanderhoof:
I'd like to strongly we bring back a solution from Conjur v4 to augment the plan proposed in this Design. In Conjur v4, variables could include information about how they were to be sent back to the requestor. 
```
- !variable
  id: secret-1
  annotations:
    content-encoding: text/base64
```
The above would send the secret back using `text/base64` encoding in the header. This allows the client to understand the data format (and what it needs to do to handle the response).

As the design currently stands, the person creating the K8s manifests needs intimate knowledge of the secret format in order to correctly define the decode mechanism. Moving context to the variable allows us to leverage the HTTP protocol for handling a wide variety of formats (ex. `text/json`, `application/octet-stream`).

gl-johnson:
Thanks for chiming in on that @jvanderhoof. That actually addresses the open question I left in the solution design upon seeing that Conjur supports a mime_type attribute, along with custom annotations. It would allow us to implement this behavior in the SDK so it doesn't require any additional configuration by the Secrets Provider user.

Is that behavior specific to the v4 API? I still see it in the [docs for Variables](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Operations/Policy/statement-ref-variable.htm?tocpath=Fundamentals%7CPolicy%7CPolicy%20statement%20reference%7C_____12#Attributes), but haven't done any verification.

Do you think there is still value in allowing the Secrets Provider manifest to dictate which variables are decoded, along with a separate addition to the upstream SDKs to handle decoding based on the Conjur API headers?

jvanderhoof:
@gl-johnson It looks like mime-type support is part of Conjur already, which simplifies things.  Good catch!

jvanderhoof:
> Do you think there is still value in allowing the Secrets Provider manifest to dictate which variables are decoded, along with a separate addition to the upstream SDKs to handle decoding based on the Conjur API headers?

Personally, I believe there's little value in defining decoding at the K8s manifest level. 

In the best case scenario, the SDK decodes the value based on the provided header. The developer does not need to be concerned about the form of the variable value. It appears in the relevant K8s secret already decoded.

In the worst case scenario, the secret appears in K8s base64 encoded. In this case, it feels more logical for the application to be aware of the secret format, and handle that format appropriately (just like it would with environment variable). Managing format at the manifest level adds an additional layer which needs to kept in sync with Vault, Conjur, and target application. It feels like we're adding an opportunity to make the process more error prone for our customers.

jvanderhoof:
>  Another consideration about using the Conjur variable annotations is whether it would be compatible with the batch retrieval endpoint currently used by Secrets Provider. I assume we would have to make separate requests for each variable to receive the correct Content-Type header for each.

Excellent catch! Batch retrieval is far more complex as a feature than most people give it credit. I'd strongly recommend we add a per-variable content type (the batch request would return `text/json`). This allows us to convert non-ascii content to a format appropriate for transfer via JSON (ex. binary data needs to be base64 encoded).  For example, something like:

```json
{
  // base64 encoded
  "variable-1": {
    "value": "Zm9vLWJhcg==\n",
    "encoding": "text/base64"
  },
  // no encoding, same result as 'text/plain'
  "variable-2": {
    "value": "foo-bar"
  },
  // none-text formats are always base64 encoded so they can be included in JSON
  "variable-3": {
    "value": "Zm9vLWJhcg==\n", // base64 encoded binary
    "encoding": "application/octet-stream"
  }
}
```
The above allows us to offload the decode to the target SDK.

gl-johnson:
> ```json
> {
>   // base64 encoded
>   "variable-1": {
>     "value": "Zm9vLWJhcg==\n",
>     "encoding": "text/base64"
>   },
>   // no encoding, same result as 'text/plain'
>   "variable-2": {
>     "value": "foo-bar"
>   },
>   // none-text formats are always base64 encoded so they can be included in JSON
>   "variable-3": {
>     "value": "Zm9vLWJhcg==\n", // base64 encoded binary
>     "encoding": "application/octet-stream"
>   }
> }
> ```
> 
> The above allows us to offload the decode to the target SDK.

I tend to agree that this would be the better approach long-term, but I'm a little hesitant to suggest a breaking change to the Conjur API to accommodate what seems to be a fairly niche enhancement request. I suppose we could add an optional request param to indicate that the requestor wants the full response w/ content-types, and leave the default response as-is (JSON list of key/val pairs). But I'd be curious on your thoughts on how best to manage this in the Conjur API.

In either case I think this warrants a deeper discussion since it'll affect more than just Secrets Provider and likely delay delivery a little bit. (cc @niteshtaneja)

niteshtaneja:
Will the above proposal of using the mime_type attribute have any compatibility issues with secrets synced via Synchronizer? Also, scope increase aside, this will also impact delivery mechanism? The original idea was to implement this in Secrets Provider and deliver it as part of monthly OSS release.

gl-johnson:
> Will the above proposal of using the mime_type attribute have any compatibility issues with secrets synced via Synchronizer? Also, scope increase aside, this will also impact delivery mechanism? The original idea was to implement this in Secrets Provider and deliver it as part of monthly OSS release.

My understanding of Synchronizer isn't very deep, but I don't foresee an issue. It would require a minor change to the Conjur policy for synchronized secrets to include the mime_type (if needed) so that info exists in the Conjur DB.

We would need releases for Conjur, conjur-api-go, and Secrets Provider to implement the change. And the user would have to be running both the newly released Conjur and Secrets Provider to use the feature.

design/secrets_decoding_design.md
```diff
@@ -0,0 +1,241 @@
...
+### P2F Mode
+For P2F, we can update the existing `conjur.org/conjur-secrets.{secret-group}` annotations with the following allowed formats:
+```
+conjur.org/conjur-secrets.groupname: |
+  - <alias_value>: <path_value>               # supported
+  - <path_value>                              # supported
+  - <alias_value>: <path_value>               # to be added
+    content-type: <content_type_value>
+  - <path_value>:                             # to be added
+      content-type: <content_type_value>
```
rpothier, line 112: For this case with the implicit key the user will have to add a colon to the line, and the indenting is different that the case with the alias is specified which could lead to user errors. So since they would have to either add a colon on the end, or add the alias_value at the beginning, I think it's clearer to just support the one case which is the ` - <alias_value>: <path_value>` and add the content-type.

gl-johnson: Agreed and updated in the doc

rpothier (approved):
LGTM!

PR #506, opened 2023-03-14
Support content-type annotation for K8s secrets

gl-johnson:
### Desired Outcome
Add support to K8s secrets mode for setting secret content-types via annotations in the secrets manifest. It should support the following:
- text (default)
- base64

The conjur-map annotation for secrets will now support the format:
```
conjur-map: |-
  <alias_value>: <var_id_value>               # already supported
  <alias_value>:                              # to be added
    id: <var_id_value>
    content-type: <content_type_value>  
```
Where `id` is required to have a non-empty value, and `content-type` is an optional key which defaults to `text` unless explicitly set to `base64`

### Implemented Changes
- Parse new YAML annotations into existing `updateDestination` struct in  K8sProvider 
- Add tests to check for the expected content-type of secrets via relevant log/error messages

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [insert issue ID]

#### Behavior

#### Security

### DIFF
--- a/pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
+++ b/pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
@@ -21,6 +21,9 @@ var testConjurSecrets = map[string]string{
 	"conjur/var/umlaut":       "ÄäÖöÜü",
 	"conjur/var/binary":       "\xf0\xff\x4a\xc3",
 	"conjur/var/empty-secret": "",
+	"conjur/var/encoded1":     "ZGVjb2RlZC12YWx1ZS0x", // == decoded-value-1
+	"conjur/var/encoded2":     "ZGVjb2RlZC12YWx1ZS0y", // == decoded-value-2
+	"conjur/var/encoded3":     "ZGVjb2RlZC12YWx1ZS0z", // == decoded-value-3
 }
 
 type testMocks struct {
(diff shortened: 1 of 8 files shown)


pkg/secrets/k8s_secrets_storage/provide_conjur_secrets_test.go
```diff
@@ -325,6 +342,121 @@ func TestProvide(t *testing.T) {
...
+						"test-still-encoded2": map[string]interface{}{
+							"id":           "conjur/var/encoded2",
+							"content-type": "text",
+						},
+					},
+				},
+			},
+			requiredSecrets: []string{"k8s-secret1", "k8s-secret2"},
+			asserts: []assertFunc{
+				// TODO - uncomment when decoding is implemented
```
szh, line 372: Is the actual decoding going to be in a separate PR

gl-johnson: I planned on tackling it in a followup task, yes. The decoding implementation should be a minor code change but that PR will also include error handling, additional tests, dev environment/manifest updates, etc.  so I thought it would be more manageable as a separate task.

This assertion was copied over from the POC since it'll be helpful for that PR.

PR #511, opened 2023-03-23
Support content-type annotation

rpothier:
### Desired Outcome

*Please describe the desired outcome for this PR.  Said another way, what was
the original request that resulted in these code changes?  Feel free to copy
this information from the connected issue.*

### Implemented Changes

*Describe how the desired outcome above has been achieved with this PR. In
particular, consider:*

- _What's changed? Why were these changes made?_
- _How should the reviewer approach this PR, especially if manual tests are required?_
- _Are there relevant screenshots you can add to the PR description?_

### Connected Issue/Story

CyberArk internal issue ID: [insert issue ID]

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -40,7 +41,13 @@ func FetchSecretsForGroups(
 				)
 				return nil, err
 			}
-
+			if spec.ContentType == "base64" {
+				sValue, err = base64.StdEncoding.DecodeString(string(sValue))
+				if err != nil {
+					err = fmt.Errorf("failed to decode secret. Error: %v", err)
+					return nil, err
```
gl-johnson, line 48: In the K8s secrets implementation we are logging the decoding error as a warning here, but then still providing the original secret value. I think this makes it a bit more resilient if there is a change in the secret value to plaintext where it won't immediately break here.

pkg/secrets/pushtofile/secret_spec.go
```diff
@@ -64,15 +66,21 @@ func (t *SecretSpec) unmarshalFromMap(node *yaml.Node) error {
...
+	t.ContentType = "text"
+	count := 0
 	for k, v := range mapValue {
-		t.Path = v
-		t.Alias = k
+		if k == "content-type" {
+			t.ContentType = v
+		} else {
+			count = count + 1
+			if count > 1 {
```
gl-johnson, line 77: I don't think any unit tests trigger this condition (too many mapping entries)

pkg/secrets/pushtofile/secret_spec.go
```diff
@@ -98,6 +117,25 @@ func validateSecretPaths(secretSpecs []SecretSpec, groupName string) []error {
...
+		}
+	}
+	return errors
+}
+
+func validateSecretContent(content, groupName string) error {
+	if content == "text" || content == "base64" {
+		return nil
+	} else {
+		return fmt.Errorf("secret group %s: the content-type of %s is invalid",
```
gl-johnson, line 134: Somewhat similar to the other comment - in the K8s secrets implementation, if a user provides an invalid content-type we log a warning and assume it should be treated as `text` so that it doesn't fail

Happy to discuss which outcome is more intuitive for the user, but either way we should aim to be consistent

pkg/secrets/pushtofile/secret_spec.go
rpothier: The test TestNewSecretSpecs/secret_spec_map_with_multiple_keys triggers this condition.

gl-johnson (approved):
LGTM

PR #514, opened 2023-04-03
Null out local variables containing secrets after use

gl-johnson:
### Desired Outcome

In security review, we discussed making sure that local variables containing secret values are nulled out to make sure they are no longer in memory after being used.

### Implemented Changes

Explicitly null out local variables which could contain secret values. After some research, this is more reliable and secure with byte slices than strings since Golang treats strings as immutable and may be left hanging around in memory, so switch to using slices where possible

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

- [x] Security architect has reviewed the changes in this PR,

### DIFF
--- a/deploy/test/test_cases/TEST_ID_1.7_providing_large_decoded_variable_successfully.sh
+++ b/deploy/test/test_cases/TEST_ID_1.7_providing_large_decoded_variable_successfully.sh
@@ -20,7 +20,7 @@ echo "Verifying pod test_env has environment variable '$environment_variable_nam
 test_pod="$(get_pod_name "$APP_NAMESPACE_NAME" 'app=test-env')"
 actual_value=$($cli_with_timeout "exec $test_pod -- printenv | grep VARIABLE_WITH_BASE64_SECRET | cut -d= -f2")
 
-if [[ "$actual_value" == "$secret_value" ]]; then
+if [[ "$actual_value" == "$secret_value"* ]]; then
     echo "$environment_variable_name is set correctly"
     # Reset the secret value to the original value for subsequent tests
     set_conjur_secret secrets/encoded "$(echo "secret-value" | tr -d '\n' | base64)" # == "c2VjcmV0LXZhbHVl"
--- a/pkg/secrets/k8s_secrets_storage/provide_conjur_secrets.go
+++ b/pkg/secrets/k8s_secrets_storage/provide_conjur_secrets.go
@@ -404,22 +404,25 @@ func (p K8sProvider) createSecretData(conjurSecrets map[string][]byte) map[strin
 
 			// Check if the secret value should be decoded in this K8s Secret
 			if dest.contentType == "base64" {
-				decodedSecretValue, err := base64.StdEncoding.DecodeString(string(secretValue))
+				decodedSecretValue := make([]byte, base64.StdEncoding.DecodedLen(len(secretValue)))
+				_, err := base64.StdEncoding.Decode(decodedSecretValue, secretValue)
+				decodedSecretValue = bytes.Trim(decodedSecretValue, "\x00")
 				if err != nil {
 					// Log the error as a warning but still provide the original secret value
 					p.log.warn(messages.CSPFK064E, secretName, dest.contentType, err.Error())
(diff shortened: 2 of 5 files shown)


pkg/secrets/pushtofile/retrieve_secrets.go
```diff
@@ -50,6 +50,7 @@ func FetchSecretsForGroups(
 					log.Warn(messages.CSPFK064E, spec.Alias, spec.ContentType, err.Error())
 				} else {
 					sValue = decodedSecretValue
+					decodedSecretValue = nil
```
andytinkham, line 53: Same as above - move the nil statement after the error checking if/else.

pkg/secrets/k8s_secrets_storage/provide_conjur_secrets.go
```diff
@@ -411,13 +411,15 @@ func (p K8sProvider) createSecretData(conjurSecrets map[string][]byte) map[strin
 					secretData[k8sSecretName][secretName] = secretValue
 				} else {
 					secretData[k8sSecretName][secretName] = decodedSecretValue
+					decodedSecretValue = nil
```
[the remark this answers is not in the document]
andytinkham: Addiitonally, are you sure that setting the values to nil removes the secret from the heap entirely? Or does it just change the memory location that "decodedSecretValue" points to, leaving the old value elsewhere on the heap? Things I'm seeing in Stack Overflow are old, but seem to indicate that strings can't be set to nil, and nil is a pointer.  I don't see any places across our code base where we currently use "zero out" or "null out" to clear strings.

Reading https://stackoverflow.com/questions/39968084/is-it-possible-to-zero-a-golang-strings-memory-safely, it sounds like we may not be able to ensure that we get all the copies of a string that might be in memory.  I suspect that secretValue is a [] byte for this reason. 

I don't know enough go to say what the best approach is for decodedSecretValue, since that's coming back as a string.  This may require discussion from a broader group.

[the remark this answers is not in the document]
gl-johnson: It's easy enough to use byte arrays instead of strings here - I hadn't looked into how strings are handled in Go, but it seems like that is the safer option so long as we handle them correctly.

Assuming that `decodedSecretValue` is now a []byte, is it sufficient to set it to nil in order to remove any references to the existing value, or would we need to iterate over the bytes and set them to a dummy val before removing the reference?

deploy/test/test_cases/TEST_ID_1.7_providing_large_decoded_variable_successfully.sh
```diff
@@ -20,7 +20,7 @@ echo "Verifying pod test_env has environment variable '$environment_variable_nam
 test_pod="$(get_pod_name "$APP_NAMESPACE_NAME" 'app=test-env')"
 actual_value=$($cli_with_timeout "exec $test_pod -- printenv | grep VARIABLE_WITH_BASE64_SECRET | cut -d= -f2")
 
-if [[ "$actual_value" == "$secret_value" ]]; then
+if [[ "$actual_value" == "$secret_value"* ]]; then
```
andytinkham, line 23: So actual_value could be secret_value plus more? What would the "plus more" part be here?

andytinkham (approved):
One question for my own education, no changes needed from my perspective. Thanks, @gl-johnson!

deploy/test/test_cases/TEST_ID_1.7_providing_large_decoded_variable_successfully.sh
gl-johnson: The "plus more" is accounting for base64 padding characters.

Base64 encoded data sometimes contains between 1 or 2 padding characters (=) at the end in order to reach a length that is a multiple of 4. 

In switching to byte slices, the decoded output was (incorrectly) leaving these padding characters in the resulting secret as null bytes. I added code to trim any null bytes from the output, which means this comparison now needs to allow for the fact there may have been a couple bytes trimmed off the end of the input data.

PR #522, opened 2023-05-08
E2E Test Refactor POC

gl-johnson:
### Desired Outcome

Enable running E2E tests locally and refactor them from bash-driven to Golang using [e2e-framework](https://github.com/kubernetes-sigs/e2e-framework) for Kubernetes.

See the solution design here: https://github.com/cyberark/secrets-provider-for-k8s/pull/523


szh:
I believe this has now been implemented, so we shouldn't need to keep the PoC open. Feel free to re-open if I'm wrong on that.

PR #523, opened 2023-05-10
E2E Testing Refactor Design Doc

gl-johnson:
### Desired Outcome

Provide a solution design for converting bash-driven E2E tests in Kubernetes to Golang using [e2e-framework](https://github.com/kubernetes-sigs/e2e-framework). The current focus is to allow running E2E tests locally against a dev/test environment, rather than relying on platform-specific E2E tests in CI which take anywhere from 30 mins to 2 hours to run and provide developer feedback.


szh:
I love this. Really excited to get started on this effort.

szh (approved):

PR #532, opened 2023-06-26
Increase test duration times

rpothier:
### Desired Outcome

Unit tests should pass every time.

The test TestRunSecretsProvider can be flaky as the timing may be off by 
one as below. 

[2023-06-22T19:53:17.382Z]     provide_conjur_secrets_test.go:318: 
[2023-06-22T19:53:17.382Z]         	Error Trace:	/secrets-provider-for-k8s/pkg/secrets/provide_conjur_secrets_test.go:318
[2023-06-22T19:53:17.382Z]         	Error:      	Not equal: 
[2023-06-22T19:53:17.382Z]         	            	expected: 7
[2023-06-22T19:53:17.382Z]         	            	actual  : 6
[2023-06-22T19:53:17.382Z]         	Test:   TestRunSecretsProvider/sidecar_container,_happy_path,_no_targets_updated

### Implemented Changes

The duration time of the tests are currently 10 msec, which was used to keep the tests as fast as possible.
However these short duration times are not what will be used in practice, and also any small processing
delays can add up and cause the test to run quicker or slower than expected.
Increasing the test times are more realistic.
These changes add less than 10 seconds to the total unit test run time.

### Connected Issue/Story

CyberArk internal issue ID: CNJR-1833

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


pkg/secrets/provide_conjur_secrets_test.go
```diff
@@ -315,7 +315,11 @@ func TestRunSecretsProvider(t *testing.T) {
 
 			// Check results
 			if err == nil {
-				assert.Equal(t, tc.expectedCount, tc.provider.count())
+				diff := tc.expectedCount - tc.provider.count()
```
jtuttle, line 318: I don't think we should be allowing this fuzzy matching for most test cases, especially the ones that are trying to distinguish between 0 and 1. It may be better to just not check the count on the cases where it's inconsistent.

rpothier: This could just be applied for the cases with higher counts. Would that work?

jtuttle: Possibly but that feels like a hacky solution to me. If we don't have a reliable count I don't think we should be testing it if there's a reasonable way to exclude the test when an "expected count" isn't provided.

jtuttle: Feel free to get a second opinion though and have someone else do the actual review here. I will not be offended. :)

rpothier: The problem is we are testing the Go timer and the inconsistencies are not run to run but in different platforms. So Jenkins runs behave a little different that running on a laptop. So when we are expecting to see seven timer ticks, occasionally we see six or eight depending on the platform.
So we're not really testing the accuracy of the Go timer but just making sure something doesn't get stuck, giving a very low or very high count.

rpothier: another option is to increase the duration times so any small variations are relatively insignificant.

jtuttle (approved):

PR #542, opened 2023-08-16
E2E Test Migration for Tests 1.1 - 1.7

juniortaeza:
### Desired Outcome

This pull request contains newly implemented tests that address Test IDs 1.1 - 1.7.

### Implemented Changes

- adds Tests 1.1 - 1.7 following the new E2E testing framework
- updates config files to contain environment variables that suffice the new E2E tests

#### Changelog

- [X] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [X] This PR does not require updating any documentation

#### Behavior

- [X] These changes are part of a larger initiative that will be reviewed later, or

#### Security

- [X] These changes are part of a larger initiative with a separate security review, or

### DIFF
--- a/e2e/k8s_test.go
+++ b/e2e/k8s_test.go
@@ -7,11 +7,83 @@ import (
 	"bytes"
 	"context"
 	"testing"
+	"strings"
+	"crypto/rand"
+	"encoding/base64"
 
 	"github.com/stretchr/testify/assert"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 	"sigs.k8s.io/e2e-framework/pkg/features"
 )
+
+func TestSSHKeysProvidedK8s(t *testing.T) {
+	f := features.New("ssh keys provided").
+		// Replaces TEST_ID_1.1_providing_ssh_keys_successfully
+		Assess("ssh key set correctly in pod", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			var stdout, stderr bytes.Buffer
+			command := []string{"printenv", "|", "grep", "SSH_KEY"}
+
+			RunCommandInSecretsProviderPod(cfg.Client(), command, &stdout, &stderr)
+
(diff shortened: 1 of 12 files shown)


e2e/k8s_test.go
```diff
@@ -29,3 +100,48 @@ func TestSecretsProvidedK8s(t *testing.T) {
...
+
+func TestLargeDecodedVariablesSecretProvidedK8s(t *testing.T) {
+	f := features.New("large decoded variables secret provided").
+		// Replaces TEST_ID_1.7_providing_large_decoded_variable_successfully
+		Assess("large decoded variable secret set correctly in pod", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			// set encoded value in conjur
+			value := "secret-value"
+			encoded := base64.StdEncoding.EncodeToString([]byte(val))
+
+			err := SetConjurSecret(cfg.Client(), "VARIABLE_WITH_BASE64_SECRET", encoded)
```
gl-johnson, line 129: Since this is running in a mode without secrets rotation, updating a secret value will not be reflected in the container unless you redeploy the secrets provider pod. It just so happens that the `VARIABLE_WITH_BASE64_SECRET` is already equal to `secret-value` from the original deployment. 

It is also intended to test decoding a value larger than 65k characters.

The flow for this test should be:

1. Generate a random string > 65k characters
2. Base64 encode the string
3. Set Conjur secret `secrets/encoded` to your encoded string
4. Re-deploy Secrets Provider pod (`ReloadWithTemplate(cfg.Client(), K8sTemplate)` should do it, I think)
5. Check the environment variable for expected value

e2e/k8s_test.go
```diff
@@ -29,3 +100,59 @@ func TestSecretsProvidedK8s(t *testing.T) {
...
+
+			encodedStr := base64.StdEncoding.EncodeToString(str)
+
+			// set encoded value in conjur and reload template
+			err = SetConjurSecret(cfg.Client(), "secrets/encoded", encodedStr)
+			if err != nil {
+				fmt.Errorf("error setting conjur secret: %s", err)
+			}
+
+			err = ReloadWithTemplate(cfg.Client(), K8sTemplate)
```
gl-johnson, line 141: It seems like this method isn't redeploying the pod as expected so the test fails, but it should be fixed in my CI PR #540 

Will go ahead and approve since everything else looks good

gl-johnson (approved):

e2e/k8s_test.go
juniortaeza: Awesome. Thank you!

PR #543, opened 2023-08-21
E2E Test Migration for Tests 2 - 16

juniortaeza:


e2e/k8s_test.go
```diff
@@ -156,3 +160,46 @@ func TestLargeDecodedVariablesSecretProvidedK8s(t *testing.T) {
...
+
+			os.Setenv("CONJUR_AUTHN_LOGIN", loginURI)
+
+			// reload template with new login configuration
+			err := ReloadWithTemplate(cfg.Client(), K8sTemplate)
+			if err != nil {
+				fmt.Errorf("error reloading secrets provider: %s", err)
+			}
+
+			// get pod name and fetch logs from pod (we cannot run kubectl commands inside a container)
```
juniortaeza, line 185: this section appears to work properly but I am currently unable to test it with proper conjur authn configuration.

the logic of retrieving the pod name, checking the pod's logs, then checking the logs for expected error code is shown below

e2e/k8s_test.go
```diff
@@ -156,3 +160,46 @@ func TestLargeDecodedVariablesSecretProvidedK8s(t *testing.T) {
...
 	testenv.Test(t, f.Feature())
 }
+
+*/
+
+func TestConjurSecretsNoPermissionK8s(t *testing.T) {
+	f := features.New("no permission to view conjur secrets").
+		// Replaces TEST_ID_12_no_conjur_secrets_permission
+		Assess("no permission to view conjur secrets in pod", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			// set environment variable in local environment
```
juniortaeza, line 170: this section of setting/getting env variables then reloading the template is not finished.

the environment variables being pulled seem to be pulled directly from my system, and not `bootstrap.env`, thus I am unable to configure a proper CONJUR_AUTHN_LOGIN id that expectedly causes the conjur secret retrieval to fail.

e2e/k8s_test.go
```diff
@@ -132,3 +152,238 @@ func TestLargeDecodedVariableSecretProvidedK8s(t *testing.T) {
 
 	testenv.Test(t, f.Feature())
 }
+
+func TestMultiplePodsChangingPwdInbetweenSecretProvidedK8s(t *testing.T) {
```
juniortaeza, line 156: I've replicated the following flow for this particular test based off of `TEST_ID_2_multiple_pods_changing_pwd_inbetween`:

1. first, check original `TEST_SECRET` value and confirm it's `supersecret`
2. determine which platform we're running on and set Conjur secret (`secret2`) for pod2
3. check `TEST_SECRET` value and confirm it's `secret2`
4. set Conjur secret (`secret3`) for pod3
5. check `TEST_SECRET` value and confirm it's `secret3`

I have not (and currently unsure how to) implemented the `cli_with_timeout` logic as well as waiting for `x` amount of pods running, `scale deployment`, etc.

Further feedback is appreciated!

e2e/k8s_test.go
```diff
@@ -132,3 +152,238 @@ func TestLargeDecodedVariableSecretProvidedK8s(t *testing.T) {
...
+		}).
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			platform := os.Getenv("PLATFORM")
+
+			if platform == "kubernetes" {
+				err := SetConjurSecret(cfg.Client(), "secrets/test_secret", "secret2")
+				if err != nil {
+					fmt.Errorf("error setting conjur secret: %s", err)
+				}
+			} else if platform == "openshift" {
```
gl-johnson, line 177: I would first check if either the kubernetes Go API or the e2e-framework library includes any functionality to do what you need here. It looks like we need two operations:
- delete pods
- scale deployments

e2e/k8s_test.go
```diff
@@ -132,3 +152,238 @@ func TestLargeDecodedVariableSecretProvidedK8s(t *testing.T) {
...
+			var stdout, stderr bytes.Buffer
+			command := []string{"printenv", "|", "grep", "TEST_SECRET"}
+
+			RunCommandInSecretsProviderPod(cfg.Client(), command, &stdout, &stderr)
+
+			assert.Contains(t, stdout.String(), "supersecret")
+
+			return ctx
+		}).
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
```
gl-johnson, line 169: Might want to double check that this works as expected. I would assume that `setup` runs before all your assertions. I'm not sure what the behavior is when you have multiple `setup` functions nested between assertions like this

e2e/k8s_test.go
gl-johnson: I would recommend looking into the e2e-framework and Kubernetes Go APIs a little bit. I would be surprised if there is not a way to scale deployments/delete pods through the k8s client

If not, then it may be necessary to add a helper script which executes the kubectl/oc commands you need

Issue #549, opened 2025-02-13
Secrets rotation improvements


romanfurst:
Hello, 

in our enterprise company, we have faced the urgency of password rotation. One of the main requirements was to be able to rotate credentials without service restarts (containers can read and apply the changed credentials on-the-fly). Also, the push-to-file solution did not fit as it requires a sidecar inside each Kubernetes pod. In the case where we are running hundreds of Kubernetes clusters, each with dozens of services, it creates significant resource overhead (besides the Istio service-mesh sidecar). Not to mention deployment configuration for every pod would be redundant.

We decided to make some secret-provider implementation modifications to fulfill the requirements for smoother password rotation.
In the following sections, I will describe how the whole solution works and which modifications in the codebase were necessary. Also, some draft pull requests were made. If there is interest in any of it, we can proceed.

![Image](https://github.com/user-attachments/assets/a0a4aacc-cf04-4408-b82d-a536b64166d2)

### The Workflow

![Image](https://github.com/user-attachments/assets/eb4bea9e-01ef-4036-aaf8-bbad87cd5ced)

note: For password rotations, we encourage dual accounts.  

0. Conjur provider is deployed as a Deployment in conjur-namespace. A Kubernetes webhook is registered to react to every new or modified labeled Secret in the cluster.
1. A new labeled secret is deployed. The mutation webhook is called.
2. The Conjur provider webhook server processes the new event and triggers secret provision for the given Secret object.
3. Retrieved Conjur secrets values (and group templates) are injected into the Kubernetes Secret.
4. The Kubernetes Secret's key is mounted (projected) as a file into container filesystem.
5. The mounted file is read during application service bootstrapping.
6. The Kubernetes Secret is modified/redeployed.
7. (Same as step 2)
8. (Same as step 3)
9. New Secret key values are automatically projected into the container filesystem.
10. Application service runtime reloads modified filesystem file on-the-fly.
11. Conjur-provider timer runs out. All labeled K8s Secrets are retrieved, all Conjur secrets are retrieved.
12. New values and template groups are injected into appropriate Kubernetes Secrets.
13. New Secrets mounts are projected into container filesystem.
14. Every application service reloads its modified file.

Advantages:

- Only one instance of the secrets-provider needs to be deployed in the cluster
- Easy addition or removal of targeted Kubernetes Secrets via labeling
- Since the provider periodically checks and retrieves Conjur secrets, it is well-suited for password rotation scenarios
[...]

PR #553, opened 2025-02-13
Run in standalone mode

romanfurst:
### Desired Outcome

The Conjur secret provider component can run as a standalone Kubernetes (K8s) deployment and periodically trigger the provisioning of Conjur secrets. Meanwhile, it reacts to every new or modified K8s secret and triggers provisioning only for that specific secret.

It would be good to introduce this feature along with  https://github.com/cyberark/secrets-provider-for-k8s/pull/550

### Implemented Changes

A new 'standalone' run mode is provided. The component periodically runs provisioning and updates target secrets. Additionally, a new HTTP server is exposed to handle registered Kubernetes (K8s) mutation webhooks. While provisioning runs automatically within the defined time period, the webhook ensures that every new or modified K8s secret is provisioned immediately.

### Connected Issue/Story

DRAFT for:

https://github.com/cyberark/secrets-provider-for-k8s/issues/549


helm/secrets-provider/templates/secrets-provider.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.favorite.drink "application" }}
```
szh, line 1: This will always be false. .Values.favorite.drink will always be "coffee"
🤣☕️

Romanecek: Oh, that's quite awkward. Probably a Ctrl+C Ctrl+V type of error. 
Fixed. (force push made)

romanfurst: Oh, that's quite awkward. Probably a Ctrl+C Ctrl+V type of error.
Fixed. (force push made)

